PR #805, opened 2014-12-18
Migrate MarketMetadata, fix HitBtc

rafalkrupinski:
Changed getMetadata() to let it return Map<CurrencyPair, ? extends MarketMetadata> so user can get all the currencies with single call.
Moved the the method to AccountService, because it uses some form of getAccountInfo most of the times anyway.

In case of HitBtc, TradeService uses the data to prepare orders and interpret trade history, so the constructor calls getMetadata() and passes the result to the TradeService. I'm not really fond of calling remote method from constructor, but I didn't find other way to do it. I'm open to suggestions.

Also, removed default currency from getTradeHistory as it's optional parameter, and fixed placeMarketOrder timeInForce parameter. I can push these fixes in separate patch, if you don't like the rest.


rafalkrupinski:
@timmolter is the above agreeable, and what about hitbtc?

timmolter:
@rafalkrupinski Sorry for the delay. With the holidays and many birthdays in my family I haven't have a chance to sit down and think this completely through. I'll hopefully get to it soon. Thanks for your patience.

rafalkrupinski:
@timmolter, what do you mean by "get rid of MarketMetaData"?
Map<CurrencyPair, MarketMetadata> is made with one remote call.
If we remove it, we either make a remote call per method you've listed, or we cache the data internally in the service without any user control over it. I don't think either option is better than the current design, where the user is making a single call, and the service, if it needs the data internally, like hitbtc, may keep the result of the last call.

3) getExchangeSymbols() will be deprecated by getMarketMetadata().keys()

timmolter:
@rafalkrupinski I mean move it into  `PollingTradeService`. Merge all the functionality there instead. Perhaps put the code that does the file reading and loading in `BasePollingService`. The "metadata" relevant to  all polling services (like currency pairs) could actually be part of `BasePollingService`. 

What do you think about that?

timmolter:
The user could call something like `loadMetadata` whenever they want to (at start up and when they think it should be refreshed. If they call `getPriceStep()` before it was loaded they get an exception telling them they need to call `loadMetadata` first.

rafalkrupinski:
@timmolter if we move the methods from MarketMetadata to PollingTradeService, then we have a copy of their implementation in every TradeService or, if we move it to BasePollingService, then every service of every exchange has a copy of metadata. I think the design proposed in this patch is slightly better :)

Alternatively please see https://github.com/rafalkrupinski/XChange/tree/meta3
It implements suggestions from before January: 
- I've removed trading fee from MarketMetadata
- getMarketMetadata() is moved to PollingTradeService.
- The metadata is still quietly kept if the TradeService requires it (HitBtc).

Problem with initialization remains. In my version the user is required to call getMarketMetadata before placeOrder or getHistory but the method throws exception if it's not implemented. Poor design for a common interface :/

This looks much better in your solution: if no metadata is needed then loadMetadata() could be a noop.
This is actually an initialisation method, so why not call it init()? We might find more uses for it in the future (e.g. some exchanges call getExchangeSymbols() for internal needs, it could be moved there).

I'd like to use loadMetadata idea (but call it init) and keep Map<CurrencyPair, MarketMetadata> exposed (as in meta3 branch). What do you think?

PS.
I'll introduce init() method in meta3 in a minute.

rafalkrupinski:
Oh, I could wrap the map in a single object, ExchangeMetadata, it would make the API look cleaner.

rafalkrupinski:
@timmolter OK, I'll take it for "I love the idea" :)

@felipelalli If you still want to contribute an implementation, you can start now :)

felipelalli:
Nice job guys! @timmolter I can't right now, I am sorry.

PR #816, opened 2015-01-12
ItBit orderbook missing date

jheusser:


rafalkrupinski:
Wouldn't it be better to leave it null instead of artificially setting to local time?
I mean, if the server returns some value, than OK, it's a time of the last operation, or in worst case, current server time, but with new Date() the field is worthless.

timmolter:
Yes, I agree. It was decided at the beginning to only allow a date that comes from the server and not to call `new Date()`. Maybe we should call the field `serverTime` to make that clearer.

rafalkrupinski:
BTW, it still wouldn't be clear if it's the timestamp of the call or the last modification of order book. We don't need to rename the field, javadoc should suffice.

jheusser:
Ok makes sense. Quite a few imps use polling date so we should change it
there as well.

PR #820, opened 2015-01-14
changed partner to api_key

davassi:
'partner' key is no more longer supported by OkCoin usd

### DIFF
--- a/xchange-okcoin/src/main/java/com/xeiam/xchange/okcoin/OkCoin.java
+++ b/xchange-okcoin/src/main/java/com/xeiam/xchange/okcoin/OkCoin.java
@@ -64,7 +64,7 @@ public interface OkCoin {
 
 	@POST
 	@Path("future_userinfo_4fix.do")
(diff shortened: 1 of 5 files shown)


jheusser:
@davassi Is this tested for okcoin.cn as well?

zholmes1:
It should be fine. In zTrader I am using the implementation that can be seen in my pull request and I haven't had any complaints from my users on okcoin.cn. In my implementation I just changed everything to String but still sent it as "partner" and everything seems fine.

zholmes1:
Just checked the documentation for both .cn and .com and they are both now looking for "api_key."

davassi:
@jheusser Actually not, I did not test against .cn. But as @zholmes1 confirm, and I even asked directly to the okcoin online support, the partner key is deprecated both side. The documentation of okcoin has some typos as well...

jheusser:
ok cool, thanks guys.

MaximSultakov:
Is it possible to have it fixed in 2.1 which is stored in Maven Repo?

rafalkrupinski:
It's impossible to change artifacts released to maven central. You could backport the fix and try asking @timmolter to release a 2.1.1 . You'd need to make branch rooted in xchange-2.1.0 tag for that.

Issue #787, opened 2014-12-02
Bitstamp: Ripple deposits and withdrawals


mmazi:
I just wanted to note that I'm working on this.

Bitstamp's own API implementation still has some bugs and I'm waiting for their team to fix them.

mmazi:
Bitstamp's bug still exists: if two-factor auth is enabled, ripple withdrawal via API will always fail with this message:
`{"error": {"auth_code": ["Invalid authentication code."]}}`
A workaround is to disable 2FA.

PR #824, opened 2015-01-17
Bitstamp: Ripple stuff: deposit address, withdraw to ripple.

mmazi:
Note that due to a bug in Bitstamp's implementation, ripple withdrawal only works if Google Authenticator authentication is disabled for the Bitstamp account.


mmazi:
This resolves #787.

PR #826, opened 2015-01-20
Add trading fee to AccountInfo @ BTC-e and Hitbtc

rafalkrupinski:
BTC-e API currently reports fee of .002 (.2%) for all listings, so I put one of the values in the AccountInfo.

Hitbtc fees vary between listings, so AccountInfo.getTradingFee() return value is configurable through xchange.properties.
All the fees are available through HitbtcTradingServiceHelper

ANX: make AccountInfo.getTradingFee() return fraction instead of percent.


rafalkrupinski:
Note that calling BTCExchange.init() is not strictly required, but it allows AccountInfo,getTradingFee() to return a non-null

PR #830, opened 2015-01-20
[pom] Remove duplicate managed dependency

rafalkrupinski:
the first definition is in line 198

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,7 +146,12 @@
 			<dependency>
 				<groupId>org.slf4j</groupId>
 				<artifactId>slf4j-api</artifactId>
-				<version>1.7.8</version>
+				<version>1.7.10</version>
+			</dependency>
+			<dependency>
+				<groupId>org.slf4j</groupId>
+				<artifactId>slf4j-simple</artifactId>
+				<version>1.7.10</version>
 			</dependency>
 
 			<!-- SLF4J implementation for use in testing -->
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -146,7 +146,12 @@
 			<dependency>
 				<groupId>org.slf4j</groupId>
 				<artifactId>slf4j-api</artifactId>
-				<version>1.7.8</version>
+				<version>1.7.10</version>
+			</dependency>
+			<dependency>
+				<groupId>org.slf4j</groupId>
+				<artifactId>slf4j-simple</artifactId>
+				<version>1.7.10</version>
```
timmolter: slf4j-simple should not be here, but rather in your own client app dependent on xchange. Or am I missing something??

rafalkrupinski: The line you commented is only dependency management to declare the version, all actual uses I added are in the test scope

timmolter: How about putting it in the dependencies section with scope test? Then it will be available to all modules automatically.

rafalkrupinski: I find setting scope in root depemdencyManagement generally a bad idea. If you use such dependency in a submodule in a different scope, then you have to declare both dependency and dependencyMangement in that submodule.

It's shouldn't be a problem in case of xchange though. Nearly all the modules have homogenic role - they don't talk to each other, and examples uses logback. Let me know if you still want me to change it.

rafalkrupinski: On a second thought, slf4j-simple is an exception here - it's hardly ever used outside of testing, but I'd still leave it as is, simply, for clarity. You decide :)

timmolter:
See my latest commit. Thanks for bringing this up!

Issue #684, opened 2014-10-03
Remove "XYZ returned an error: " ?


zholmes1:
I'm wondering if it would be a good idea to remove instances of the prefix that is sometimes applied to ExchangeException (for example "BTC-E returned an error: " + e.getMessage()). It would allow more flexibility in displaying error messages to end users. Regardless, we should choose one or the other and make it consistent across all exchange implementations.

timmolter:
Done. Methods now throw IOException or some subclass of `ExchangeException`, which you can catch the specifc ones and handle things properly.

PR #832, opened 2015-01-21
Introduce exception types for nonce error, throttling

rafalkrupinski:
and insufficient funds.

I'd like to push issue #684 forward.

I can handle programatically nonce error and throttling. @mmazi's example with insufficient funds is a good one too.

Until all the exchanges are modified, some of the exchanges will throw exceptions that don't extend ExchangeException - we break the generic interface.
I guess we will have to live with `catch(ExchangeException | IllegalArgumentException e)` for the time being.

`FrequencyLimitExceededException extends IllegalStateException` because It represents a transient state of the exchange - internal counter that at some point exceeds some threshold.
`FundsExceededException extends IllegalArgumentException` - it could bee seen as illegal argument (requested order exceeds the funds) or illegal state (funds are insufficient to place the order), so it's either this or `InsufficientFundsException extends IllegalStateException`.

### DIFF
--- a/xchange-core/src/main/java/com/xeiam/xchange/FundsExceededException.java
+++ b/xchange-core/src/main/java/com/xeiam/xchange/FundsExceededException.java
@@ -1,0 +1,20 @@
+package com.xeiam.xchange;
+
+public class FundsExceededException extends IllegalArgumentException {
+  private static final long serialVersionUID = 2406896439418334853L;
+
+  public FundsExceededException() {
+  }
+
+  public FundsExceededException(String s) {
+    super(s);
+  }
+
+  public FundsExceededException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public FundsExceededException(Throwable cause) {
+    super(cause);
+  }
+}
--- a/xchange-core/src/main/java/com/xeiam/xchange/FundsExceededException.java
+++ b/xchange-core/src/main/java/com/xeiam/xchange/FundsExceededException.java
@@ -1,10 +1,7 @@
 package com.xeiam.xchange;
 
-public class FundsExceededException extends IllegalArgumentException {
+public class FundsExceededException extends ExchangeException {
   private static final long serialVersionUID = 2406896439418334853L;
-
-  public FundsExceededException() {
-  }
 
   public FundsExceededException(String s) {
     super(s);
@@ -14,7 +11,4 @@ public class FundsExceededException extends IllegalArgumentException {
     super(message, cause);
   }
 
-  public FundsExceededException(Throwable cause) {
-    super(cause);
-  }
 }
--- a/xchange-anx/src/main/java/com/xeiam/xchange/anx/v2/service/polling/ANXAccountServiceRaw.java
+++ b/xchange-anx/src/main/java/com/xeiam/xchange/anx/v2/service/polling/ANXAccountServiceRaw.java
@@ -3,10 +3,10 @@ package com.xeiam.xchange.anx.v2.service.polling;
 import java.io.IOException;
 import java.math.BigDecimal;
 
+import si.mazi.rescu.HttpStatusIOException;
 import si.mazi.rescu.RestProxyFactory;
 import si.mazi.rescu.SynchronizedValueFactory;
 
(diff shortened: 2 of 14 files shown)


rafalkrupinski:
@timmolter, if the exceptions are acceptable, I can provide some implementations before this is merged.

timmolter:
Thanks for digging this issue up again. 

What's the advantage of extending `IllegalStateException` instead of `ExchangeException`? I like `ExchangeException` because the constructor forces extenders to provide a nice message. 

In any case, we should stick to just one parent exception type, probably `IllegalStateException` if not `ExchangeException`.

rafalkrupinski:
> What's the advantage of extending IllegalStateException instead of ExchangeException?

Leveraging the standard and having a good design: e.g. Funds Exceeded is-a Illegal Argument, Nonce error is-a Illegal State.

> I like ExchangeException because the constructor forces extenders to provide a nice message.

I don't think it's doing a good job - `"Error calling $method(): " + e.getError()` doesn't look very nice ;)
Anyway, I can hide the default c-tors in the exception types I'm introducing here.

> In any case, we should stick to just one parent exception type, probably IllegalStateException if not ExchangeException.

We already have to catch IOException, ExchangeException and `${exchange}Exception extends RuntimeException`.
I don't know about others, but my approach is to catch IOException and RuntimeException anyway.

timmolter:
Just pasting this here: 

`public class IllegalArgumentException extends RuntimeException`
Thrown to indicate that a method has been passed an illegal or inappropriate argument.

 `public class IllegalStateException extends RuntimeException`
Signals that a method has been invoked at an illegal or inappropriate time. In other words, the Java environment or Java application is not in an appropriate state for the requested operation.

rafalkrupinski:
@timmolter are you saying that these exceptions are inappropriate for remote methods?

timmolter:
Here's what I propose:

`ExchangeException extends RuntimeException`
`FrequencyLimitExceededException extends ExchangeException`
`InsufficientFundsException extends ExchangeException`
etc...

These are all unique exceptions related to exchanges and they do follow a Java standard of exception class hierarchy. Why introduce another layer such as `IllegalArgumentException`? Just get straight to the point: e.g. it's a `InsufficientFundsException` which is a `ExchangeException` which is a `RuntimeException`. If you want a human readable message to print out you have it, but you don't need too.

And like @mmazi said:

Just my two cents: @timmolter , @zholmes1 , I tend to see @rafalkrupinski's suggestion to use Exception class hierarchy instead of a type field as a better idea. @zholmes1's code then becomes:

    catch (InsufficientFundsException e) {
        limitOrder.setTradableAmount(amount.multiply(new BigDecimal("0.99")));
        tradeService.placeLimitOrder(limitOrder);
    } 
    catch (TooFrequentException e) {
        Thread.sleep(2000);
        tradeService.placeLimitOrder(limitOrder);
    }
    catch (...)

This gives several advantages, eg. it conforms to the general exception paradigm in Java and allows for different exception classes to have different properties (eg. InsufficientFundsException could have requiredFuns, actualFunds fields etc.), and for different exception types to be handled on different levels of the method call stack easily.

timmolter:
@rafalkrupinski Yes, I think an `ExchangeException ` is more relevant, that's all. I don't think we should feel forced to use Java standard Exceptions when we have our own more appropriate one, that's all.

rafalkrupinski:
No problem, I'll change the hierarchy, it doesn't influence the error checking logic on the xchange side, nor for the users.

Please don't close the PR :)

rafalkrupinski:
@timmolter I've pushed the changes in the exception type hierarchy you requested,
and implementations for the 4 exchanges that I have account at.

xchange-core/src/main/java/com/xeiam/xchange/FundsExceededException.java
```diff
@@ -0,0 +1,14 @@
+package com.xeiam.xchange;
+
+public class FundsExceededException extends ExchangeException {
+  private static final long serialVersionUID = 2406896439418334853L;
```
timmolter: is this long really needed?

timmolter:
looks good.

PR #834, opened 2015-01-23
Pom

rafalkrupinski:
- remove duplicate managed dep commons-lang3
- add missing slf implementation for some tests - slf4j-simple, managed dependency in the root pom
- add pluginManagement for the javadoc plugin to fix missing version warnings


rafalkrupinski:
Oh, OK

PR #840, opened 2015-01-26
[pom] Bring back source plugin to the default build

rafalkrupinski:
I't useful and unlike javadoc, doesn't influence build time


timmolter:
thanks. I agree.

PR #847, opened 2015-01-30
Adding extra (optionally provided) fields to Ticker.

sebster:
Hi,

Many exchanges provide more fields in the ticker than just the basic information contained in the Ticker class. Perhaps it would be an idea to add the common extra fields to the Ticker class and fill them in when they are available. That way a user of the API can still use the generic Ticker class, and see what information is available, instead of heaving to resort to using specific <Exchange>Ticker classes and not being able to use the generic classes at all.

As an example, the Volume Weighted Average Price (VWAP) is included in many tickers (though not always). I have added it to the Ticker class below and updated the Bitstamp and BTC-e adapters to adapt the ticker including the VWAP info. If an exchange does not provide VWAP, the ticker field will be empty (null).

This change does not break any existing code using the API.

If you consider this a useful change, I am willing to go through the other exchanges and update the adapters there where needed.

Best regards,
Sebastiaan


timmolter:
I'd be OK with this. Over the years, I've seen vwap quite a lot, so I think coverage of this parameter would be decent across exchanges.

@sebster can you see whay the PR cannot be merged? Do you need to update you develop branch with mine?

sebster:
Hi,

I messed up with the pull request (I forgot the git fetch before the rebase), sorry about that. I have fixed it, now the pull request merges.

Best regards,
Sebastiaan

PR #848, opened 2015-01-31
Cryptsy public API bug fix for v3.0.0-SNAPSHOT, Kraken general enhancements, Trade Builder additions

ww3456:
re: CryptsyBasePollingService now that it contains both the public and private API does it really make sense for it to stay templated in its current form? 

The Cryptsy fix was to stop the URL being generated as http://pubapi.cryptsy.com/api/api.php?method=orderdatav2
but instead http://pubapi.cryptsy.com/api.php?method=orderdatav2

### DIFF
--- a/xchange-core/src/main/java/com/xeiam/xchange/dto/marketdata/Trade.java
+++ b/xchange-core/src/main/java/com/xeiam/xchange/dto/marketdata/Trade.java
@@ -132,6 +132,11 @@ public class Trade {
 
     }
 
+    public static Builder from(Trade trade) {
+      return new Builder().type(trade.getType()).tradableAmount(trade.getTradableAmount()).currencyPair(trade.getCurrencyPair()).price(trade.getPrice()).timestamp(trade.getTimestamp())
+          .id(trade.getId());
+    }
+
     public Builder type(OrderType type) {
 
       this.type = type;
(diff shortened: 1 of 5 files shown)


timmolter:
Thank you.

timmolter:
FYI, I just pushed a commit that removed all the generics from Cyptsy to be consistent with the rest of XChange.

ww3456:
That Cryptsy class is looking cleaner. However, when you subsequently removed generics from PoloniexBasePollingService the public API broke. e.g. it should have been sending 
https://poloniex.com/public?command=returnTicker but was sending 
https://poloniex.com/tradingApi?command=returnTicker
I've reverted to the previous version of this Poloniex class and it works fine. I'm not sure of the exact intent behind these recent BasePollingService changes so will leave it to others to decide how best to fix this. Thanks!

timmolter:
@ww3456 Good catch. Thanks. I see what the problem is now. I will check all exchange modules and fix appropriately.

timmolter:
@ww3456 I pushed a fix.

ww3456:
@timmolter great, this is working now. For the record I now have the latest develop 3.0.0-SNAPSHOT running OK on BTC-E, Bitfinex, Cryptsy, Kraken and Poloniex for both market data and trading.

timmolter:
Excellent. Thanks for telling me which exchanges are working for you too!

PR #851, opened 2015-02-02
Added a more robust integer nonce factory. Use it for BTC-e.

mmazi:
I'm hoping to end the BTC-e nonce troubles once and for all. This is a v3 / SynchronizedValueFactory re-implementation of the nonce factory I implemented and described in #756. (Please see that issue or/and the test included in this PR if you're interested why this PR is necessary. Tl;dr: it supports several exchange instances operating at the same time.)

I've reviewed the many previous BTCE nonce issues and pull requests and I believe this should resolve all of them. (It doesn't resolve my GAE issue 100%, but it's good enough.)

This PR includes a test that fails with the `AtomicLongIncrementalTime2013NonceFactory` that is currently used for BTC-e, and passes with the new `TimestampIncrementingNonceFactory`.

A few notes:

* I think the new `TimestampIncrementingNonceFactory` is the best choice for any exchange that uses an integer nonce. (For long nonces, I think `CurrentTimeNonceFactory` is the best choice.)
* Correct me if I'm wrong but I don't think AtomicInteger (or AtomicLong) is necessary since the nonce factories are used in a synchronized manner anyway.
* This nonce factory is compatible with the one that was previously used for BTC-e so people won't have to drop and re-create the API keys they are already using. (The exception to this could be if someone _constantly_ (as opposed to occasionally) makes more than 4 requests per second, which I think is impossible anyway.)

### DIFF
--- a/xchange-btce/src/main/java/com/xeiam/xchange/btce/v3/BTCEExchange.java
+++ b/xchange-btce/src/main/java/com/xeiam/xchange/btce/v3/BTCEExchange.java
@@ -8,11 +8,11 @@ import com.xeiam.xchange.ExchangeSpecification;
 import com.xeiam.xchange.btce.v3.service.polling.BTCEAccountService;
 import com.xeiam.xchange.btce.v3.service.polling.BTCEMarketDataService;
 import com.xeiam.xchange.btce.v3.service.polling.BTCETradeService;
-import com.xeiam.xchange.utils.nonce.AtomicLongIncrementalTime2013NonceFactory;
+import com.xeiam.xchange.utils.nonce.TimestampIncrementingNonceFactory;
 
 public class BTCEExchange extends BaseExchange implements Exchange {
 
-  private SynchronizedValueFactory<Long> nonceFactory = new AtomicLongIncrementalTime2013NonceFactory();
+  private SynchronizedValueFactory<Long> nonceFactory = new TimestampIncrementingNonceFactory();
 
   @Override
   public void applySpecification(ExchangeSpecification exchangeSpecification) {
--- a/xchange-btce/src/test/java/com/xeiam/xchange/btce/v3/BTCEExchangeTest.java
+++ b/xchange-btce/src/test/java/com/xeiam/xchange/btce/v3/BTCEExchangeTest.java
@@ -1,0 +1,46 @@
+package com.xeiam.xchange.btce.v3;
+
+import org.junit.Assert;
+import org.junit.Test;
+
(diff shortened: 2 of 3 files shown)


timmolter:
Yeah, some of the nonce factories are far superior than others and the AtomicLong probably isn't necessary like you say.

rafalkrupinski:
I'm not sure about this patch. It's not a problem with BTC-e, it's a problem with user environment. We should either use this ValueGenerator impl everywhere if it's so good, or provide a way for users to pass a custom one.

timmolter:
@rafalkrupinski A few comments.

1. During my refactor, I made about 8 new NonceFactory classes. I just pulled out hardcoded nonce functions from the modules and made a Nonce Factory, or used another factory if it already existed. 
1. Yeah, some of the nonce factories are far superior than others and it *would* be nice if all the inferior ones were replaced. I cannot do that though as I'd probably break something.
1. Users **cannot** pass whatever custom one they please, like you said. Should we make it part of the `ExchangeSpecification`?

mmazi:
I think providing a way to provide custom might be a good idea, since it seems in different environments or usage scenarios different nonce factories will be ideal. But it should be optional, so a default nonce factory will be used if a user doesn't set up a custom one.

As @timmolter said, replacing existing nonce factories with a different one would render some existing users' API keys useless (because in some cases the new nonce factory might start producing lower nonces than the old one).

mmazi:
@rafalkrupinski , what are your concrete concerns about this patch? Do you see anything that is worse now after the patch than it was before? This patch just replaces one nonce factory with another that is better in some scenarios (like the one in the included test case) and not worse in any known scenarios. It seems to me the issues you've raised (which are valid issues) are not really directly connected with this particular patch.

rafalkrupinski:
@mmazi 
> what are your concrete concerns about this patch?

You said said it:

> in different environments or usage scenarios different nonce factories will be ideal.

My concern is that we're putting user-specific fix in a common place. Someone might need something different.

@timmolter Yes, I think exchange specification is the place. Keep in mind, users must be aware what generic argument (Long vs Int) is required by the the exchange.

mmazi:
@rafalkrupinski 
> My concern is that we're putting user-specific fix in a common place.

In my view, the new nonce factory is no more user-specific than the previous one; you could call the previous one a user-specific fix too (eg. compared to a simple counter nonce).

> Someone might need something different.

This was just as true before this patch. (In fact, this was the _reason_ for this patch.) Moreover, this patch strictly reduces the set of people that might need something different.

I think a new issue should be opened for what you're saying. I agree it's a valid issue.

PR #854, opened 2015-02-04
Convert HH:mm:ss to timestamp within the last 24 hours.

sebster:
Huobi timestamps are of the form "HH:mm:ss", and are converted by the BitVcAdapters class to timestamps with the same hour, minute and date on January 1st, 1970. This date is incorrect, but there is even another more serious problem when sorting the dates by this timestamp. If there are two consecutive trades with times 23:59:59 and 00:00:01 then the second trade is after first trade, but when sorting by the timestamp, it will be placed before the first (it will be placed in the beginning of the list instead of the end), since 00:00:01 January 1st, 1970 is before 23:59:59 January 1st, 1970.

The following patch converts the the HH:mm:ss time to a timestamp that is within a 24 hour window ending in the current time. The timestamp is allowed to be up to 5 minutes in the future to compensate for any clock skew. This means that effectively, if your clock is not off by more than 5 minutes (which it shouldn't be with internet time synchronization), all trades within the last 23 hours and 55 minutes will get the correct timestamp.

Note that this patch is based on the time zone of Huobi being China Standard Time (Asia/Shanghai). If this adapter needs to be used for exchanges outside this time zone, the time zone should probably be a parameter.

Best regards,
Sebastiaan

### DIFF
--- a/xchange-bitvc/src/main/java/com/xeiam/xchange/bitvc/BitVcAdapters.java
+++ b/xchange-bitvc/src/main/java/com/xeiam/xchange/bitvc/BitVcAdapters.java
@@ -8,14 +8,13 @@ import static com.xeiam.xchange.dto.Order.OrderType.BID;
 import static com.xeiam.xchange.dto.marketdata.Trades.TradeSortType.SortByTimestamp;
 
 import java.math.BigDecimal;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
-import java.util.Locale;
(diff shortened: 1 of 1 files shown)


rafalkrupinski:
Is it possible to use server time instead if hardcoded 5 min margin?

sebster:
I don't really see how, since the response simply does not contain the server time:

http://api.huobi.com/staticmarket/detail_btc_json.js

The 5 minute margin is not really a big deal I think. For it to go wrong, the last 15 trades would have to go back 23 hours and 55 minutes which is really unlikely. And note that it breaks anyway if the last 15 trades go back for more than 24 hours, because the fundamental assumption that is made is that all the trades in the list happened in the last 24 hours.

timmolter:
Geez, why can't exchanges just send epoch timestamps in JSON?

PR #861, opened 2015-02-17
OkCoin websocket initial commit

jheusser:
Initial commit for websocket market data streaming. It's based on the reference implementation by OkCoin and already contains its trade functionality, but has not been wired up into the XChange interface yet. It's mostly reusing the existing market data DTOs.

`OkCoinSocketIODemo` shows the standard demo for streaming.

### DIFF
--- a/xchange-okcoin/pom.xml
+++ b/xchange-okcoin/pom.xml
@@ -33,7 +33,6 @@
 		<dependency>
 			<groupId>io.netty</groupId>
 			<artifactId>netty-all</artifactId>
-			<version>4.1.0.Beta3</version>
 		</dependency>
 
 	</dependencies>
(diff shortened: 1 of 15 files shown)


jheusser:
Probably have to commit some code to make currency switching easier by default

xchange-okcoin/pom.xml
```diff
@@ -29,6 +30,12 @@
 			<version>${project.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>io.netty</groupId>
+			<artifactId>netty-all</artifactId>
+			<version>4.1.0.Beta3</version>
+		</dependency>
+
```
timmolter: Can you put this under `<dependencyManagement>` in parent.pom?

jheusser: thanks, will do

timmolter:
Good work!!

PR #884, opened 2015-03-10
Loyalbit implementation

mmazi:
Currently supported: market depth (orderbook), get balance, place order, cancel order, get open orders, get trade history.


timmolter:
great!

Issue #837, opened 2015-01-23
Proposal to change Map<CurrencyPair, ? extends TradeMetaData> getTradeMetaDataMap()..


timmolter:
to `Map<CurrencyPair, TradeMetaData> getTradeMetaDataMap()` in `PollingTradeService`, where `TradeMetaData` is a regular ol' DTO containing the union of all the meta data params found across all exchanges. 

We would then add an adapter and adapt call in the *Raw tradeservice class, just like for everything else.

@rafalkrupinski This is mostly a question directed towards you. I'd like to change it to be consistent with the rest of XChange's API. I guess I need to know how you're using this method. Are guess you are casting to a specific MetaData Impl like:

    HitbtcTradeMetaData eur = (HitbtcTradeMetaData) metaMap.get(CurrencyPair.BTC_EUR);

It's important to me to that this library follows the principle of least surprise and this "outlier" breaks the pattern.

Do you have a compelling reason for it to be the way you created it?

rafalkrupinski:
I agree with your arguments, @timmolter , but how many fields would you like to add?

I'd suggest implementing more exchanges and better see what are we talking about, before we go for more common solutions.

timmolter:
@rafalkrupinski That's the million dollar question as usual!

Your approach is interesting. You still need to cast though, so your logic needs to be aware of what exchange it is operating on to do the cast and call the specific methods it needs to get those unique exchange meta data values. It's not completely generic as desired. Correct me if I'm wrong. That's why I asked how you were using it. 

Your approach doesn't give any advantage over casting to a *Raw service to get to the exchange-specific info (if we moved the metadata stuff there). BTW, I'm not criticizing, but rather just making an observation. Please let me know if I'm missing something. 

I'll be thinking through all this over the next couple of days.

rafalkrupinski:
Primary purpose of the metadata was min amount and scales, and the whole idea came out of validating user orders. Fees were put there only by the way, so I might have not think it through :)

timmolter:
@rafalkrupinski As you will see, I started a refactor. I apologize in advance for messing with your code.

rafalkrupinski:
@timmolter If you fill like apologizing, then don't mess with it ;)

Seriously though, if you can make it better, it's also better for me :)

timmolter:
@rafalkrupinski  It needs to be messed with. ;) In the end it will be better, but there might be some growing pains, and for that I apologize. 

[...]

PR #886, opened 2015-03-12
Proposed implementation of #837 raw+generic meta data

rafalkrupinski:
This patch extends MetaData class to hold a map of CurrencyPair instead of a list, so for backwards compatibility the old class has been renamed to SimpleMetaData.

Meta data is available in two flovours: raw and generic:

- raw metadata contains the data from the json file
- generic MetaData is a common type, witch objects are built from the json file and online information.

The user can call exchange specific methods to get raw local and remote metadata to acquire at least all the of the information available from MetaData objects.

One thing that still requires change is polling restrictions. I think about two options

- single field minimumPollDelay (millisecond)
- minimumPollDelayPublic and minimumPollDelayPrivate

I'll do the single field, which is enough for me, and wait for feedback, if any.


timmolter:
Is it possible to not have a new `xchange-generators` module? I'd like to keep all exchange-specifc code in the existing exchange-specific module, i.e. `ANXGenerator` in `xchange-anx`.

timmolter:
Is it necessary to have both:

    com.xeiam.xchange.dto.MarketMetaData;
    com.xeiam.xchange.dto.MetaData;

or could we just have `com.xeiam.xchange.dto.MetaData` alone?

rafalkrupinski:
> Is it possible to not have a new xchange-generators module? I'd like to keep all exchange-specific code in the existing exchange-specific module

Similar to `xchange-examples`, `generators` isn't a production code, just the functions that generate the json files. It's virtually single-use and I was going to remove it from the xchange source tree, but in the future some files might require so much changes that it would be infeasible to do it by hand. Id rather remove it altogether rather than move the code to modules.

> Is it necessary to have both com.xeiam.xchange.dto.MarketMetaData [and] com.xeiam.xchange.dto.MetaData [?]

MetaData (which maybe could be renamed to ExchangeMetaData) is exchange-specific, e.g. polling rate, MarketMetaData is specific to currency pair at a particular exchange, like price scale. There is one to many relation between them, so I don't see how they could be merged.

I wonder why you're not asking about the CurrencyMetaData class, which right now has just one int field and could be simply replaced by an Integer. I think there is more data for currency, like deposit/withdrawal fees, but I left it for later :)

timmolter:
Alright. I cloned your fork and checked out the meta branch. Just a few remarks as I go through it.

1. We should probably move all the *meta DTOs to a `meta` sub-package.
1. What demo app, if any, can I run to see the ANX meta data in action?
1. regarding the `generators` module, how about putting those classes in the respective modules test package under a `bootstrap` package. I've done that in many projects.
1. Where can I see where "objects are built from the json file and online information" and then used by say a *TradeService ?

I'm really liking the direction you're taking this.

rafalkrupinski:
> We should probably move all the *meta DTOs to a meta sub-package.

Will do.

> What demo app, if any, can I run to see the ANX meta data in action?

There is ANXMetaDataDemo, but I guess it doesn't really show how to use it. TBD

> regarding the generators module, how about putting those classes in the respective modules test package under a bootstrap package. I've done that in many projects.

Well, the generators aren't really tests, but I see that you want to avoid another module that depends on everything else, and I don't see better alternatives. Will do.

> Where can I see where "objects are built from the json file and online information" and then used by say a *TradeService ?

I'm not really sure if any of the trade services actually requires *online* information. HitBtc needs lot sizes, but IIRC it's not remotely available. In this case, having separate raw/local and generic metadata saves us a remote call before the first use of the trade service. I guess I'll have to look into other exchanges. Thanks for pointing that out.

> I'm really liking the direction you're taking this.

Thanks, nice to hear it.

timmolter:
@rafalkrupinski there is a compile error. Caught by Travis CI.

rafalkrupinski:
@timmolter don't worry about that :)
I know, I just haven't got a time

timmolter:
Ah, OK. Take your time....

timmolter:
is this ready for me to look at now?

rafalkrupinski:
@timmolter 
>> Where can I see where "objects are built from the json file and online information" and then used by say a *TradeService ?

> I'm not really sure if any of the trade services actually requires online information. HitBtc needs lot sizes, but IIRC it's not remotely available.

Update: I'm just working on Hitbtc, the lot sizes are available remotely and required by `HitbtcTradeService.getTradeHistory()`.
I could use either the common ExchangeMetaData or HitbtcSymbols to adapt the result, but I'd rather not call the remote `getSymbols()` each time. We can get it either in `applySpecification()` or in dedicated `init()` method, if the former would break the rule of the least surprise. I don't think adding `init()` to the common interface would affect many users.

After implementing Hitbtc, I'll start testing it with my application. It should be OK now, so I think you can check it out.

rafalkrupinski:
@timmolter If you still want to move adapter methods to *Service hierarchy, then Exchange classes would loose access to them, and in some cases `adapt*MetaData` uses other `adapt*` methods.

One of the solutions would be to reintroduce MetaDataService, but the last time we didn't find a consensus on how to store and pass MetaData objects between the services.

Anyway, I think it would be good to do it before merging this branch.

timmolter:
1. How about having an `initRemoteMetaData` method that must be implemented by all exchanges and which is called during the exchange initialization? In many cases it will not need to be implemented. This is where these remote calls could be added and used to build up the meta data, which can then act as a cache of metadata so it doesn't have to be constantly called.
1. Why both in ExchangeMetaData?

  @JsonProperty("currency_pairs")
  private Map<CurrencyPair, MarketMetaData> currencyPairs;

  @JsonProperty("currency")
  private Map<String, CurrencyMetaData> currency;

rafalkrupinski:
> How about having an initRemoteMetaData method that must be implemented by all exchanges and which is called during the exchange initialization?

Good, this is what I'm doing right now.

> Why both in ExchangeMetaData?

    private Map<CurrencyPair, MarketMetaData> currencyPairs;
    
    private Map<String, CurrencyMetaData> currency;

`CurrencyMetaData` is meta data of currencies (wallets) not tied to specific currency pair. Now it's just the wallet scale (which might be different than that of the markets), but this would also be the place for deposit and withdrawal fees in the future.

rafalkrupinski:
@timmolter a simple demo app you asked about should appear soon in the latest commit.
Right now i'm bringing back order verification. Previously it were methods in the DTO, but I'm putting it in PollingTradeService, because now we implement raw/generic pattern instead of inheritance.

rafalkrupinski:
@timmolter I was thinking about representing rate limit data. how about this form:

    {
      "publicRateLimits":[
        {
           "timeSpan":10, // number of time units
           "timeUnit":"seconds", // one of java.util.concurrent.TimeUnit values, case insensitive
           "calls":10 // number of calls allowed
        }
      ],
      "privateRateLimits":{...},
      "rateLimitsShared":false
    }

the above example says it's allowed to make 10 calls per 10 second. publicRateLimits is a list, so there may be many such limits to supports burst calls allowed by some exchanges.
There would also be `privateRateLimits` with the same structure and a flag saying that public and private calls share rate limiting.

It would be possible to represent most (if not all) rate limiting policies, and quite easy to compute sleep for constant polling.

timmolter:
@rafalkrupinski That looks really great.

rafalkrupinski:
@timmolter It's ready for review & merge.

In most exchanges I've just migrated the metadata json file without any integration with the remote data, so we could just get on with this merge, and actually implement getting remote data later.
I added priceScale in most cases, sometimes these values were guessed from the current order book.

Some exchanges use remote metadata already,  these will need Exchange.remoteInit() called before some or all other remote calls. Of course it's safe to call it on any Exchange object.

PollingTradeService adds verifyOrder() for both Market and LimitOrder, implemented in BaseExchangeService.

MetaData json may contain these information:
- currencyPair - info related to market / currency pair, price scale, fees and so on.
- currency - related to currency / commodity / wallet, like value scale.
- call rate limits, as described above

I've removed getExchangeSymbols() from most places, there is the default implementation in BaseExchangeService for compatibility, but I'd like to remove it entirely in 4.0

timmolter:
Thanks for doing this. You probably noticed I'm super busy lately. I apologize for that and apprciate your hard work on this issue.

rafalkrupinski:
No, I think it's super fast review&merge for a 10kloc change :)

mmazi:
@rafalkrupinski thanks for doing this, it will be very useful.

Do you have any quick pointers how on migrating the old json metadata files into the new format? I see this hasn't been done yet for some exchanges (eg. BitVC, BTC-e, Hitbtc) which causes failures in a project of mine that relies on the metadata.

mmazi:
@rafalkrupinski Nevermind my question, I took a better look and it seems I can do what I need easily.

rafalkrupinski:
Call exchange.remoteInit(), it should help with  Btc-e and Hitbtc.

Whats wrong with BitVC? I tested the code only with provided examples, only some exchanges required working api keys.

mmazi:
I'm confused about BitVC in XChange in general. It seems to have merged with Huobi or something as it apparently doesn't exist as a separate XChange module any more.

One problem is that several exchange modules have been removed in v3.0.1-SNAPSHOT, but the version hasn't been bumped and all these still reside in Maven repositories (remote and local), so they are still visible to anyone using this version.

(None of this seems to be related to this PR though.)

rafalkrupinski:
Can't help you much with bitvc/huobi :) there is huobi.json and it looks OK to me.

Does other exchanges' metadata work for you?

mmazi:
I think I'll be fine, thanks.

timmolter:
We can bump up the version actually to 3.1.0.

PR #887, opened 2015-03-12
Parse return value of create limit order

Achterhoeker:


Achterhoeker:
Parse return value of create limit order on clevercoin

PR #892, opened 2015-03-18
Add comments to hitbtc market data service

bryan4887:
Provides some useful commends for the `Object[] args` parameter.


rafalkrupinski:
Why not a javadoc?

timmolter:
Yep, JavaDoc would be much better. Of course, the regular comments are much better than non at all!

PR #895, opened 2015-03-20
[exchange-okcoin] differentiate the cn,com,futures exchange name

lazyp:


jheusser:
:+1: 

futures is pretty untested/unusted atm. If anyone wants to revisit the futures API that would be great.

PR #896, opened 2015-03-20
Huobi: remove sorting of ask side of orderbook

jheusser:
Huobi changed the sort order of the ask side of their orderbook at `http://market.huobi.com/staticmarket/depth_btc_json.js`

Sorting is not necessary anymore.


jheusser:
This is quite a tricky issue to detect. Probably some tests requesting actual market data (with no auth needed) can easily detect such issues though.

PR #905, opened 2015-03-25
Added Taurus Exchange implementation

mmazi:
This is an almost full implementation (market orders and lookup order are missing).


timmolter:
wow.

PR #906, opened 2015-03-26
Huobi Streaming Support

glabmoris:
This is still experimental but the demo file provided in examples works fine.

This has been built on top of a client that may need major refactoring in the near future, but I've chosen additional features over pretty code.

Also included are 3 examples for the huobi module, a few minor fixes for other projects (such as OkCoin)

Enjoy!


timmolter:
great!

jheusser:
great work, will give this a test run :+1:

glabmoris:
Thanks, but I only gave @sutra 's client the final push it needed to reach the tree by working out the final kinks. He did most of the legwork :)

bryan4887:
Hi @glabmoris,
Thanks again for this contribution.  I've been working to get this working, but something seems fishy with the orderbooks returned by the Huobi `StreamingExchangeService`.  Try running the following code.  What you'll find is that the BTC/CNY orderbook is only returned once in the beginning, but never again.  LTC/CNY books, trades, and tickers continue to be returned, however only trades and tickers are returned for BTC/CNY.

```
public static void main(String[] args) throws InterruptedException {
		SLF4JBridgeHandler.removeHandlersForRootLogger();
	    SLF4JBridgeHandler.install();

		ExchangeSpecification exSpec = new ExchangeSpecification(HuobiExchange.class);
		
		Exchange exchange = ExchangeFactory.INSTANCE.createExchange(exSpec);
		ExchangeStreamingConfiguration streamCfg = new HuobiExchangeStreamingConfiguration();
		StreamingExchangeService service = exchange.getStreamingExchangeService(streamCfg);
		service.connect();
		for (int i = 0; i < 1000; i++) {
			ExchangeEvent evt = service.getNextEvent();
			ExchangeEventType type = evt.getEventType();
			System.out.print(type);
			CurrencyPair cp = null;
			Trade trade;
			OrderBook ob;
			if (type.equals(ExchangeEventType.TRADE)) {
				trade = (Trade) evt.getPayload();
				cp = trade.getCurrencyPair();
				System.out.println(":" + cp);
			} else if (type.equals(ExchangeEventType.DEPTH)) {
				ob = (OrderBook) evt.getPayload();
				cp = ob.getAsks().get(0).getCurrencyPair();
				if (cp.equals(new CurrencyPair("BTC", "CNY"))) {
					System.err.println(":" + cp);
				} else {
					System.out.println(":" + cp);
				}
			} else {
				System.out.println("");
			}
		}
		service.disconnect();

	}
```

I looked into it, but I can't figure out where the problem is coming from. Any ideas?

glabmoris:
Hi @bryan4887 , sorry for the delay, work has been keeping me insanely busy. I've been looking for such an answer myself for my marketmaking bot but to no avail. The Api docs being in mandarin makes it hard to dig into the services' functionality. Last time I checked there were "channels" one could subscribe to when opening à WS connexion. Might be worth investigating, or maybe @sutra may chime in, since this is mostly his code.

glabmoris:
PS orderbooks from huobi are fishy as **** and have been known to make some very weird moves, such as being swapped overnight (bids became asks and vice versa). Use at your own risk

bryan4887:
Hahaha, ok.  Well we're not in this because it's a safe place to play ;)  Thanks for the thoughts.  It looks like in the meantime, I can use the REST API.  @sutra , if you know of something obvious concerning these issues, feel free to suggest places to look, and perhaps I can make time to fix it up.

Issue #914, opened 2015-03-31
BTC-e TickerFetchIntegration test failing on SSL handshake.


PeterNSteinmetz:
On develop branch at e905c9dd3e1710aa65a94ccceca757b3e27007fb the attempt to run TickerFetchIntegration test for xchange-btce is failing with the complaint about the certificate:

javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

This appears to be related to another issue https://github.com/timmolter/XChange/issues/52 . 

This also happens when trying to fetch account data. Switching to compiling XChange-btce with java 6 on my machine fixes this problem.

PeterNSteinmetz:
Looks like this is only with Java 7, at least on Mac OS 10.10.2. Switching to Java 8 40 solves the problem as well. BTC-e is now using a certificate issued by GoDaddy and so the JVM in 8 must have GoDaddy's root CA certificate installed. 

    keytool -list -keystore /Library/Java/JavaVirtualMachines/jdk1.8.0_40.jdk/Contents/Home/jre/lib/security/cacerts | grep godaddy

gives 

    godaddyclass2ca, Jan 20, 2005, trustedCertEntry, 
    godaddyrootg2ca, Jul 18, 2014, trustedCertEntry, 

so it looks like the latter CA cert is the one required for BTC-e's current certificate on their website.

timmolter:
Yup, that's the issue. You can read more about it here: https://github.com/timmolter/XChange/issues/52

PR #915, opened 2015-03-31
Integration test of retrieving account information.

PeterNSteinmetz:
This commit provides an integration test of pulling account information for BTCe. It is presently failing, like the TickerFetchIntegration with a complaint about the certificate and illustrates issue #914.


PeterNSteinmetz:
To run this requires setting the apiKey and secretKey as well as setting to a Java 6 compilation on my machine.

If one sets the apiKey and secretKey to the values used in the demo application, the nonce value is incorrect and cannot be set properly, even based on the exception thrown which names a number, since the nonceFactory field is private and it is likely best policy not to add methods, such as changing the nonceFactory, just for testing.

PR #920, opened 2015-04-04
Transaction history retrieval

PeterNSteinmetz:
This branch refactors the TradeHistoryParams object, fixes a few uses of that, and adds the ability to retrieve the transaction history from BTCE. Also includes an integration test of retrieval using the exchangeConfiguration.json file when on the class path.


rafalkrupinski:
I don't understand what you were trying to achieve in BTCETradeService. In what cases is the new method called?

PeterNSteinmetz:
The intention was to provide access to the transaction history for the BTCe exchange. These functions are not used within the library, but by other programs. 

I placed this function on the TradeService by analogy to how this was done with the Bitstamp exchange. While it might be possible to extract some type of TransactionService, it seems like this will be more difficult given the high variability of what is included in transaction history on the different exchanges.

rafalkrupinski:
Is it a raw version of getTradeHistory(TradeHistoryParams)?

PeterNSteinmetz:
No, it retrieves the transaction history, not the trade history. The transaction history contains all trades but also all deposits, withdrawals and warrants.

PR #924, opened 2015-04-10
independent reserve marketdata

kzbikowski:


xchange-independentreserve/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+		<name>Xeiam, LLC</name>
+		<url>http://xeiam.com</url>
+	</organization>
+
+	<!-- Parent provides default configuration for dependencies -->
+	<dependencies>
+        <dependency>
+            <groupId>joda-time</groupId>
+            <artifactId>joda-time</artifactId>
+            <version>2.7</version>
```
timmolter: Can you avoid jodatime? XChange strives to keep the dependencies to a minimum as has so far avoided jodatime.

rafalkrupinski: The only usage I found is converting timestamp in seconds to millis, I am right? If so, there is DateUtils class in the project for that.

kzbikowski: Oh, ok. Haven't known about that. I'll change that.

PR #929, opened 2015-04-14
Clean warnings in xchange-core.

sutra:


timmolter:
This is IDE specific so I'm not too excited about adding it. The generics, I would add in a separate PR.

PR #932, opened 2015-04-16
Ir trades

kzbikowski:


.gitignore
```diff
@@ -22,6 +22,7 @@ META-INF/
 
 # Misc.
 .DS_Store
+xchange-examples/src/main/java/com/xeiam/xchange/examples/independentreserve/IndependentReserveDemoUtils.java
```
timmolter: can you remove this?

xchange-core/src/main/java/com/xeiam/xchange/dto/Order.java
```diff
@@ -16,7 +16,7 @@
      * Buying order (you're making an offer)
      */
     BID,
-    /**
+      type, /**
```
timmolter: type?

.gitignore
kzbikowski: of course, I incorrectly assumed that gitignore is also ignored

xchange-core/src/main/java/com/xeiam/xchange/dto/Order.java
kzbikowski: IDE added, removing

PR #936, opened 2015-04-29
OkCoin Futures: trading functionality

jheusser:
This adds/fixes OkCoin futures trading (which was previously implemented but unfinished) - spot trading is functionality unchanged (fixed some deprecated stuff in the API). 

- supports only *cross margin mode* right now. Some work needs to be done to support fixed/cross with configurable switch
- account info returns total account equity, raw method can access realised/unrealised pnl
- default leverage is 10x, specifying `Futures_Leverage` different leverage can be specified

Not sure why this can't be automatically merged, will have to see after submitting PR.


jheusser:
Will fix merge conflicts

jheusser:
Ok that was an extremely frustrating merge conflict. Eventually split `OkCoinOrder` and `OkCoinFuturesOrder` as the fields are too different and diverging, new adapters for futures, etc.

PR #942, opened 2015-05-04
Bitso: Fix an incorrectly named json property.

mmazi:
This caused a NPE when attempting to access the user's trade history.


mmazi:
I just noticed there may be other errors that need fixing in Bitso.

Please wait a bit before merging so I have a chance to fix those too in the same branch. I'll post when I'm done.

timmolter:
OK, I'll wait. Thanks!

mmazi:
Ok, done.

PR #945, opened 2015-05-06
OkCoin: Fix futures typo

jheusser:
Typo in extracting margin field when querying for cross margin funds.


jheusser:
Will have to add another commit here

PR #946, opened 2015-05-06
Added bitcoin.de implementation for market data 

bryan4887:


timmolter:
excellent addition! 

Does the ticker and orderbook require an API key or is it completely public?

bryan4887:
It requires an API key unfortunately, and they're not exactly quick about issuing them either.

timmolter:
@bryan4887 do you also have to have an account there or can anyone request an API key?

bryan4887:
@timmolter good question.  I registered with a login and everything, but I didn't input any financial info.

timmolter:
@bryan4887 Cool, thanks.

PR #948, opened 2015-05-08
Added ticker implementations for Bitso, Cointrader, and Loyalbit

mmazi:


rafalkrupinski:
@mmazi  @jheusser, Tim is on vacation. May I make a suggestion that one of you guys with commit access do the merges?

jheusser:
@rafalkrupinski thanks, will do the merges.

PR #968, opened 2015-05-30
Fix/test locale

rafalkrupinski:
Fixed a test broken if run with non-english locales.
Also setting travis to test with locales:
C
pl_PL
zh_CN
so any errors depending on locale should show.


timmolter:
excellent!

Issue #181, opened 2013-10-14
Ripple API support


MarkusTeufelberger:
While this might be a bit more difficult to manage (Ripple allows several issuers per currency, so there might be the need to set up multiple "exchanges" e.g. "USD.Bitstamp <--> BTC.Bitstamp" and "USD.Bitstamp <--> BTC.Weexchange") it might be good to have the prices on Ripple for both display and active trading available in XChange.

Documentation of API commands is available at https://ripple.com/wiki/RPC_API and the interface to ripple would be either a local server ("rippled") or any public one (a safe one to use would be s1.ripple.com). Commands can be sent via POST or (preferrably) websockets.

I don't do much Java but at least I could help with any questions that arise. :)

ww3456:
I've now added order entry, order cancel and balance query [here](https://github.com/ww3456/XChange/commit/efe8bc5c0f8d19659fee55c4a273229ee7274b08) . However, I've only briefly tested it so don't feel it is quite ready for a pull request yet. However, the order cancel is causing a couple of rescu warnings in the log file:
```
WARN  [si.mazi.rescu.RestMethodMetadata] DELETE request declared as consuming method body as application/json. While body is allowed, it should be ignored by the server. Is this intended?
WARN  [si.mazi.rescu.RestInvocation] DELETE request will contain a body. While this is allowed, the body should be ignored by the server. Is this intended?
```
Since the Ripple REST API requires a json body in the order cancel DELETE request (https://github.com/ripple/ripple-rest#cancel-order), can @mmazi or anyone suggest the best way of suppressing, or avoiding these warnings?

mmazi:
The reason for the warnings is that ripple-rest doesn't conform to the HTTP protocol. Now that you know this, you can safely ignore the warnings :) The warnings are intended for the (rescu) users to start bugging the server (ripple-rest in this case) developers to make a HTTP-protocol-conforming server.

An alternative to ignoring the warnings is to configure logging differently (eg. set threshold for `si.mazi.rescu.RestInvocation` to ERROR)  -- note that I **don't recommend doing this** since it will turn off other logging that may be important.

Also note that the RestMethodMetadata warning will be logged once (per client instance), while the RestInvocation warning will be logged with every order cancel request.

Bottom line: I think the current behavior with the logged warnings makes sense, considering that people should not write servers that don't conform to protocols, and considering that some clients will simply fail (or, I believe, issue a POST instead of a DELETE, etc.) in such cases.

[...]

PR #974, opened 2015-06-06
[ripple] Basic/Initial Ripple implementation with order book support

ww3456:
Currently only supports polling for the order book but I intend to add order entry, cancel, open order query, own trade history, hopefully in the next few days. Ripple support was previously mentioned in issue #181. I tried to follow the New Implementation Best Practices wiki, let me know if there is anything that has been missed, or that can just be improved in general. Thanks.


timmolter:
beautiful

PR #982, opened 2015-06-09
Ripple order entry, order cancel, account orders, trades and balance queries added

ww3456:
Ripple order entry, order cancel, own order query and balance query added. Also added an additional data map to the core Order class, for passing counterparty information on order entry, own order query and order book query - hope this is OK. 

Main missing feature (for me) is a lack of trade history. There does not currently appear to be a way of obtaining this using their [REST API](https://github.com/ripple/ripple-rest/). 

Related to #181


timmolter:
I'm hesitant on adding that Map to `Order`. I'd either want to go "all in" and add that to all the DTOs or stick to the current philosophy of "down casting" to the exchange-specific service and pulling out the info needed.

ww3456:
For standard stuff such as order entry, polling for own orders, I think it would be good to be able to use the generic interfaces and objects as much as possible. The other way I thought it could be done was on order entry go down a route similar to TradeHistoryParams, with a 2nd argument for placeLimitOrder (and placeMarketOrder) that extends OrderEntryParams. On the way back from the exchange for queries, the returned objects could be a RippleLimitOrder that extends LimitOrder. 

What is your concern about adding the additional data map to Order? There is already a similar map on Trade.

ww3456:
I've now added trade history support. Assume merge decision is pending outcome of discussion #990 about the additional data maps. 

The CI build failed due to SocketTimeoutExceptions in the Coinmate project, perhaps these should be integration tests rather than standard unit tests.

timmolter:
Yes, any test that operates over the network should be an integration test. See step #9 here: https://github.com/timmolter/XChange/wiki/New-Implementation-Best-Practices

timmolter:
@ww3456 Can you look at the travis build and see why it's failing?

ww3456:
The build failed due to Coinmate integration tests failing on the network, I've renamed them as part of #1008. This Ripple PR isn't ready for merge yet, I still need to get rid of the additional data maps in Order and Trade. Will try and do that this week.

ww3456:
* Removed additional data map from core Order and Trade classes (#990).
* Order.Builder classes have been reworked to add inheritance. 
* On the Ripple side everything is now working for me.

Please let me know if there are any issues. Thanks.

xchange-core/src/main/java/com/xeiam/xchange/dto/Order.java
```diff
@@ -168,4 +168,66 @@ public boolean equals(Object obj) {
     }
     return true;
   }
+
+  public static class Builder {
```
timmolter: thanks for adding a builder!

PR #989, opened 2015-06-14
Loading CoinbaseEx exchange symbols from the exchange itself

radumiron:
Load the exchange symbols from the exchange directly instead of loading the exchange symbols from a file.

### DIFF
--- a/xchange-coinbaseex/src/test/java/com/xeiam/xchange/coinbaseex/CoinbaseExExchangeTest.java
+++ b/xchange-coinbaseex/src/test/java/com/xeiam/xchange/coinbaseex/CoinbaseExExchangeTest.java
@@ -1,7 +1,13 @@
 package com.xeiam.xchange.coinbaseex;
 
 import com.xeiam.xchange.ExchangeFactory;
+import com.xeiam.xchange.currency.CurrencyPair;
 import org.junit.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+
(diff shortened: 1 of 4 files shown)


xchange-coinbaseex/src/test/java/com/xeiam/xchange/coinbaseex/CoinbaseExExchangeTest.java
```diff
@@ -1,12 +1,29 @@
 package com.xeiam.xchange.coinbaseex;
```
timmolter: I don't think this should be a unit test because if Coinbase changes the currency pairs, the test will fail. At the least, it should be an integration test. Preferably, it should be move to `xchange-examples` as a demo class.

timmolter:
No. Sorry, I forgot to mention the main issue. While that's better, the test will fail, if the connection to Coinbase is down. For tests like this in XChange, we designate them with `Integration` at the end of the class name and they can be run with: `mvn clean verify -DskipIntegrationTests=false`.

For unit tests, we stick to locally runnable tests such as deserializing JSON into the DTOs.

I'll update the docs here: https://github.com/timmolter/XChange/wiki/New-Implementation-Best-Practices

radumiron:
OK, thanks for letting me know. I did the changes, hopefully they are ready to be merged. 
Also, I'll pay more attention to the Best Practices doc in the future pull requests!

PR #998, opened 2015-06-18
Making BitsoMarketDataService smarter

bryan4887:
I added a private `retry()` method that takes as an argument, among other things, an interface with a single method that it will call a specified number of times before giving up.  I think this is a nice behavior to expect from a `PollingMarketDataService`.  Feel free to add your suggestions about the best place for this logic.  
It could be implemented at the client interface, but then that client has to be aware of exactly the error message that will be returned for a particular exchange.  Since that it going to be specific to each exchange, I thought it makes sense to keep it local to that exchange, as I have done here.

Thoughts?


mmazi:
This could be a good idea. I think it needs some work though. Here's my first thoughts after a quick review (roughly in the order of how important I find them):

0. The `Object... args` parameters are extremely error-prone and I would try to avoid them if at all possible. They can simply be omitted if using in-line classes anyway (the original method parameters must then be declared `final`). If the parameters are omitted, `java.util.concurrent.Callable` can simply be used instead of `PollingFunction`.
0. In the last iteration, the loop will handle the exception, log the messages, but won't actually retry.
0. Please create a new pull request from a branch that is freshly rebased on the develop branch.
0. If the caught exception does not match, it should be re-thrown, not logged.
0. This functionality could be used in general in XChange, not just for Bitso, but I think it needs some more thought first.
0. The exception could be checked for the correct type by some kind of `interface Predicate<E extends Exception> { boolean matches(E e); }` instead of just the error message.
0. The wait time could back off exponentially.

timmolter:
@mmazi Thanks for your comments. 

@bryan4887 The idea is good, but we cannot put that logic in there as we need to keep those remote API calls "dumb" in the name of keeping the XChange API clean and simple to understand and use. See the [principle of least astonishment (POLA)](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), which is relevant here. I agree with @mmazi that the idea is good, it needs more thought and it could be used in general in XChange and not specific to just a few implementations.

mmazi:
@timmolter We've had a chat with @bryan4887 about this and he's preparing a proposal solution that will be decoupled from Bitso, independent from exchange implementation (and perhaps in fact from XChange), easy to use, and the user will be free to choose whether and where to use it.

timmolter:
@mmazi Excellent. Great to hear.

timmolter:
what's the status of this?  Can it be closed?

bryan4887:
Yeah, go ahead and close it for now.  I'll resubmit when I have the refactoring complete.

Issue #986, opened 2015-06-12
1000 mBTC bounty – implement meXBT Exchange


williamcoates:
Documentation for the meXBT exchange API is here: https://mexbt.com/api/http. Best way to get playing around is to test in the sandbox, more information on the previous link. 

You can also check our clients for Ruby, Node and Python:

Ruby: https://github.com/meXBT/mexbt-ruby
Node.js: https://github.com/meXBT/mexbt-node
Python: https://github.com/meXBT/mexbt-python

To qualify for the bounty the code should also contain sufficient unit test coverage.

Authentication to the exchange API is basically identical to Bitstamp but with different parameter names, so should be able to re-use some code there. 

Feel free to drop me any questions!

sutra:
So this is the list of API to be implemented?

Public API

api.ticker()
api.orders()
api.trades()
api.currency_pairs()

Private API

api.create_order(1234.56, pair='btcmxn', side='buy', order_type='market')
api.cancel_order(123)
api.cancel_all_orders()
api.modify_order(123, 'move_to_top')
api.modify_order(123, 'execute_now')
api.account_info()
api.account_balance()
api.account_orders()
api.account_trades(start_index=-1, count=10)
api.account_addresses()
api.withdraw(1.23456789, 'foo')

williamcoates:
This is quite an exhaustive overview of the exchange api:

http://docs.mexbtprivateapi.apiary.io

But I was assuming it's only important to implement what is needed to integrate with the XChange interface... is there much value in implementing custom exchange API that is not part of the XChange core?

rafalkrupinski:
@williamcoates you're the sponsor, it's a question for you. What do you want to pay for?

sutra:
Do not find any special, seems just all common methods.
Taking this issue, ideally 1 or 2 weeks could finish the work.

williamcoates:
At least the minimum XChange core API, thoroughly tested.

sutra:
```
Content-Type: application/json
{
  "isAccepted": "true",
  "serverOrderId": "18",
  "dateTimeUtc": "952132543213543"
}  
```
Hi @williamcoates , what's the notation of the dateTimeUtc? Haven't found any description about it.

sutra:
OK, I finally found it:
```
Sadly, the timestamp is a .NET tick :( To convert to a regular JS date you can use this function:

convertDotNetTick = function(tick) {
    new Date((tick - 621355968000000000) / 10000);
}
```

[...]

PR #999, opened 2015-06-21
[MeXBT] Implement RESTful API & WebSocket API of MeXBT.

sutra:
refs #986


xchange-mexbt/src/main/java/com/xeiam/xchange/mexbt/dto/account/MeXBTWithdrawRequest.java
```diff
@@ -0,0 +1,29 @@
...
+  private final BigDecimal amount;
+  private final String sendToAddress;
+
+  public MeXBTWithdrawRequest(String apiKey, SynchronizedValueFactory<Long> nonceFactory, MeXBTDigest meXBTDigest, String ins, BigDecimal amount, String sendToAddress) {
+    super(apiKey, nonceFactory, meXBTDigest, ins);
+    this.amount = amount;
+    this.sendToAddress = sendToAddress;
+  }
+
+  public BigDecimal getAmount() {
```
williamcoates: It sucks but it turns out in this call the amount has to be a string, or the API gives an error. This is due to the underlying AlphaPoint API we are using unfortunately.

bryan4887:
Great work @sutra.  I tried out the websocket version.  Is there any reason why it's only returning TICKER and ORDER_ADDED events?  What happened to DEPTHs?

williamcoates:
@sutra I was unable to get authenticated calls to work, I always get the generic "invalid session" error which is thrown when invalid credentials are used, for example with the following code (ignoring imports and not showing credentials):

```Java
ExchangeSpecification exSpec = new MeXBTExchange().getDefaultExchangeSpecification();
exSpec.setExchangeSpecificParametersItem(MeXBTExchange.PRIVATE_API_URI_KEY, "https://private-api-sandbox.mexbt.com");
exSpec.setUserName("xxx@test.com");
exSpec.setApiKey("xxx");
exSpec.setSecretKey("xxx");
Exchange mexbt = ExchangeFactory.INSTANCE.createExchange(exSpec);
PollingAccountService accountService = mexbt.getPollingAccountService();
MeXBTAccountServiceRaw rawAccountService = (MeXBTAccountServiceRaw) accountService;
MeXBTUserInfoResponse userInfoResponse = rawAccountService.getMe();
System.out.println(userInfoResponse);
```
I tested this with sandbox credentials and also production (commenting out the setting of the private uri url). Does this work for you?

sutra:
Works for me:

```
23:00:29.266 [default] [main] DEBUG com.xeiam.xchange.ExchangeFactory - Creating exchange from specification 
23:00:29.593 [default] [main] DEBUG com.xeiam.xchange.BaseExchange - MetaData [currencyPairs=[BTC/MXN, BTC/USD], maxPrivatePollRatePerSecond=0, maxPrivatePollRatePer10Second=0, maxPrivatePollRatePerMinute=0, maxPrivatePollRatePerHour=0, maxPublicPollRatePerSecond=0, maxPublicPollRatePer10Second=0, maxPublicPollRatePerMinute=0, maxPublicPollRatePerHour=0, fiatAmountMultiplier=0, cryptoAmountMultiplier=0] 
23:00:29.610 [default] [main] DEBUG si.mazi.rescu.Config - Loaded properties from rescu.properties. 
23:00:29.613 [default] [main] DEBUG si.mazi.rescu.Config - Configuration from rescu.properties: 
23:00:29.613 [default] [main] DEBUG si.mazi.rescu.Config - httpConnTimeout = 300000 
23:00:29.613 [default] [main] DEBUG si.mazi.rescu.Config - httpReadTimeout = 300000 
23:00:29.613 [default] [main] DEBUG si.mazi.rescu.Config - proxyHost = localhost 
23:00:29.613 [default] [main] DEBUG si.mazi.rescu.Config - proxyPort = 8118 
23:00:29.613 [default] [main] DEBUG si.mazi.rescu.Config - ignoreHttpErrorCodes = false 
23:00:29.624 [default] [main] INFO  si.mazi.rescu.HttpTemplate - Using proxy HTTP @ localhost/127.0.0.1:8118 
23:00:29.635 [default] [main] INFO  si.mazi.rescu.HttpTemplate - Using proxy HTTP @ localhost/127.0.0.1:8118 
23:00:29.665 [default] [main] INFO  si.mazi.rescu.HttpTemplate - Using proxy HTTP @ localhost/127.0.0.1:8118 
23:00:29.904 [default] [main] DEBUG si.mazi.rescu.HttpTemplate - Executing POST request at https://private-api-sandbox.mexbt.com/v1/me 
23:00:29.904 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Request body = {"apiKey":"df0fbe2e52eb0e3a5eef4779898f5bba","apiNonce":1435676429,"apiSig":"923359D33DFD37E3C6745A60CADCEC2A46966384FB89443523B7722EDB2F3EE5"} 
23:00:29.904 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Request headers = {Accept=application/json, Content-Type=application/json} 
23:00:29.968 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Header request property: key='Accept-Charset', value='UTF-8' 
23:00:29.968 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Header request property: key='User-Agent', value='ResCU JDK/6 AppleWebKit/535.7 Chrome/16.0.912.36 Safari/535.7' 
23:00:29.968 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Header request property: key='Accept', value='application/json' 
23:00:29.968 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Header request property: key='Content-Type', value='application/json' 
23:00:35.252 [default] [main] DEBUG si.mazi.rescu.HttpTemplate - Request http status = 200 
23:00:35.255 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Http call returned 200; response body:
{"userInfoKVP":[{"key":"verificationLevels","value":"['0']"},{"key":"language","value":"en"},{"key":"name","value":"sutra"},{"key":"hasSeenInstrumentsTooltip","value":"true"},{"key":"affiliateId","value":"TBFIM"}],"isAccepted":true} 
com.xeiam.xchange.mexbt.dto.account.MeXBTUserInfoResponse@6ed3ccb2
```

williamcoates:
The problem is with the nonce factory you are using. For new accounts this won't matter, but for someone using another client the nonce will be too low. I will make a line comment for the fix.

sutra:
@bryan4887 according to the doc [here](https://mexbt.com/en/api/websocket/), there are only `Live Ticker` and `Live Trades and Order book`(`When action = 0 it's an order, when action = 1 it's a trade.`)
So only 3 events I have mapped:
1. ExchangeEventType.TICKER
2. ExchangeEventType.ORDER_ADDED(action = 0)
3. ExchangeEventType.TRADE(action = 1)

xchange-mexbt/src/main/java/com/xeiam/xchange/mexbt/MeXBTExchange.java
```diff
@@ -0,0 +1,67 @@
...
+import com.xeiam.xchange.Exchange;
+import com.xeiam.xchange.ExchangeSpecification;
+import com.xeiam.xchange.mexbt.service.polling.MeXBTAccountService;
+import com.xeiam.xchange.mexbt.service.polling.MeXBTMarketDataService;
+import com.xeiam.xchange.mexbt.service.polling.MeXBTTradeService;
+import com.xeiam.xchange.mexbt.service.streaming.MeXBTExchangeStreamingConfiguration;
+import com.xeiam.xchange.mexbt.service.streaming.MeXBTStreamingService;
+import com.xeiam.xchange.service.streaming.ExchangeStreamingConfiguration;
+import com.xeiam.xchange.service.streaming.StreamingExchangeService;
+import com.xeiam.xchange.utils.nonce.CurrentTime1000NonceFactory;
```
williamcoates: Should use CurrentTimeNonceFactory, because ruby client calculates the nonce using current epoch time in ms.

xchange-mexbt/src/main/java/com/xeiam/xchange/mexbt/MeXBTExchange.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class MeXBTExchange extends BaseExchange implements Exchange {
+
+  public static final String PRIVATE_API_URI_KEY = "private-api";
+  public static final String WSS_TICKER_URI_KEY = "wss.ticker";
+  public static final String WSS_TRADES_AND_ORDERS_URI_KEY = "wss.trades-and-orders";
+
+  private final SynchronizedValueFactory<Long> nonceFactory = new CurrentTime1000NonceFactory();
```
williamcoates: ditto

sutra:
@williamcoates one api key should not be shared with other clients, this is same as the other Exchanges.

williamcoates:
This is not a question of sharing api keys its a question of using the same credentials with different libraries. Just because I am using the Ruby library shouldn't mean the XChange interface breaks. I meant "client" == "client library"

xchange-mexbt/src/main/java/com/xeiam/xchange/mexbt/MeXBTExchange.java
sutra: Change to use epoch time in ms is fine.

But be better to change the document as well: `apiNonce: An integer that must be increasing for each request (for example use the current unix timestamp)`

williamcoates:
But in general looks fantastic can you please fix those small line comments and then I think we are good.

williamcoates:
Sorry, I meant also the withdraw issue, that amount has to be a string. Nicest would be to enforce passing to the method as a BigDecimal and convert it to string afterwards

xchange-mexbt/src/main/java/com/xeiam/xchange/mexbt/dto/account/MeXBTWithdrawRequest.java
sutra: Currently, the request body is like:
00:28:24.677 [default] [main] TRACE si.mazi.rescu.HttpTemplate - Request body = ```{"amount":0.000000001} ```

Do you mean it should be like:
```{"amount":"0.000000001"} ```

bryan4887:
Thanks for the explanation, @sutra.  @williamcoates, One friendly suggestion:  ORDER_ADDED is a pretty inconvenient data structure for most traders (and probably many users of XChange) because the useful data (orderbooks) needs to be reconstructed.  I could keep hitting your REST API to get the complete books, but that is expensive for both clients and the server.  I would recommend publishing full orderbooks anytime an order is added or canceled.

williamcoates:
Thanks for the suggestion, makes sense

PR #1006, opened 2015-06-26
Added Retries utility class.

bryan4887:
Added Retries class from this discussion: https://github.com/timmolter/XChange/pull/998
It's basically all @mmazi's handiwork.  I just put the pieces together.


bryan4887:
```
[ERROR] /home/travis/build/timmolter/XChange/xchange-core/src/main/java/com/xeiam/xchange/utils/Retries.java:[4,26] package java.util.function does not exist
[ERROR] /home/travis/build/timmolter/XChange/xchange-core/src/main/java/com/xeiam/xchange/utils/Retries.java:[40,25] cannot find symbol
  symbol:   class Predicate
  location: class com.xeiam.xchange.utils.Retries
```

Will refactor this to use a plain interface that works the same as Predicate<Exception>.

PR #1022, opened 2015-07-14
[core] made Order abstract and improved LimitOrder.Builder

ww3456:
As discussed with @rafalkrupinski [here](https://github.com/timmolter/XChange/commit/3f29273f80a64a2447934b25a3eaf45829c6e7f8#commitcomment-12137423) the LimitOrder.Builder from(...) method needs improvement. Also as agreed by @timmolter the parent Order class should be abstract, orders should be a MarketOrder or LimitOrder.


rafalkrupinski:
Thanks!

PR #1000, opened 2015-07-17
okcoin streaming not reconnecting

douggie:
Hi,

it would seem that com.xeiam.xchange.okcoin.service.streaming.WebSocketBase should automatically handle reconnects of the websockets.

How is it suggested to  handel certain exceptions such as hostnotfound in the event the connectivity is lost, or if the channel is not alive, nor do those exceptions seem to be bubbled up as call backs to the calling thread, so the main application cannot handel them by resubmitting another thread until connectivity is restored.

Quick simulation is to add a static route to okcoin.com once the streaming service has processed a set of events.
 route -n add 58.96.162.1/24 192.168.0.20
 route -n delete 58.96.162.0


timmolter:
I think Netty is the way to go in the future as well. A pretty major refactor would be necessary.

@douggie Can you  look into why the TravisCI build is failing?

PR #1026, opened 2015-07-19
[ripple] trade fixes and enhancements

ww3456:
Trade transfer fees now being considered, added some test cases for these.


ww3456:
Thanks. If anyone uses it and has any feedback please let me know. I've been running with Ripple enabled for the last couple of days and it seems fine so far.

PR #1050, opened 2015-08-23
[itbit] properly cast List of open orders to Array

doublesharp:
Previous code was causing a ClassCastException.


rafalkrupinski:
What's wrong with `orders.toArray(new ItBitOrder[orders.size()])` ?

doublesharp:
I don't see anything wrong with that approach. Honestly I'm not familiar
enough with the internals of ArrayList or Arrays to know which is more
efficient at copying the List to an array, but I would think they perform
similarly - please correct me if I am wrong. My understanding is that both
methods will iterate the List and copy to a typed array with the specified
size.

PR #1058, opened 2015-09-12
Fix ItBit withdrawal issue for high precision amounts

chrisrico:
ItBit doesn't support more than 4 decimals of precision in crypto amounts. Is rounding up the right answer? It seems so to me.

### DIFF
--- a/xchange-itbit/src/main/java/com/xeiam/xchange/itbit/v1/ItBitAdapters.java
+++ b/xchange-itbit/src/main/java/com/xeiam/xchange/itbit/v1/ItBitAdapters.java
@@ -1,6 +1,8 @@
 package com.xeiam.xchange.itbit.v1;
 
 import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.text.DecimalFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


timmolter:
I think the proper approach would be to use BigDecimal so that the exact amount is recorded.

chrisrico:
What I mean is that ItBit only supports 4 decimals of precision in their
API. If you send 5 decimals, they return an error.

chrisrico:
Just to verify this for you, in case you are skeptical (I was), I sent them this request which failed...

    {"side":"buy","type":"limit","currency":"XBT","amount":"0.01000000","price":"100","instrument":"XBTUSD"}

When I contacted them to find out why it was failing, this is what they said.

I cross tested your post body with our python wrapper, it return the same error. 
logs:
    {"instrument": "XBTUSD", "price": "100", "currency": "XBT", "amount": "0.01000000", "type": "limit", "side": "buy"}
    {u'validationErrors': [{u'field': u'amount', u'code': 51001, u'description': u'Value contains more than 4 decimals of precision.'}], u'code': 10005, u'description': u'Your request could not be processed because one or more parameters were invalid.', u'requestId': u'46d263f7-b8f6-473f-9528-818ccacbb215'}

You will need to adjust your precision to 4 decimal places on the amount field in the post body. 
you might want to tweak the core code on the library you are using or contact the core developer on that library to make adjustment.

timmolter:
Thanks! Makes sense.

Issue #1064, opened 2015-09-21
300 mBTC Bounty - Remove PollingTradeService deprecated `getTradeHistory`


timmolter:
... and update all implementations to use the `getTradeHistory(TradeHistoryParams params)` if necessary.

rafalkrupinski:
Let me do it

rafalkrupinski:
@felipelalli You were the original author of the MercadoBitcoin code, would you please verify if getTradeHistory works properly? Thanks!

felipelalli:
Sure. I'm in vacation, next week I can see it. Is it OK?

- from mobile

timmolter:
@rafalkrupinski Thank you! What BTC address should I send the bounty to?

rafalkrupinski:
@felipelalli Sure thing, thanks again!

timmolter:
Paid. Thanks!

felipelalli:
@rafalkrupinski I am back from vacation, but to be honest I can't have a look on this soon. But eventually I'll use the API in the future, so if I find anything wrong I send a patch to fix. Thank you and I am sorry.

PR #1066, opened 2015-09-22
Remove PollingTradeService.getTradeHistory(Object...)

rafalkrupinski:
implement #1064

Trade history not implemented for:
- LakeBTC
- CoinbaseEx
- TheRock
- Vircurex

Mercado isn't implemented, but should be trivial - WiP

exchanges accept long page length, but our api allows only int:
- Bitso
- Bitstamp


timmolter:
Great! Also thanks for summarizing.

Should we bump the int to a long?

rafalkrupinski:
@timmolter 

> Should we bump the int to a long?

Do you expect it to ever exceed 2^31?
I'd rather look into going down to int.

timmolter:
Oh, I thought it might be a constraint dictated by the exchange. Can the long be changed to an int then for Bitso and Bitstamp?

rafalkrupinski:
On another thought, it's a minor clean-up. I've looked at some api docs and some exchanges declare maximum value of 1000, others only say it's a numerical value. I'd pass on this for the time being. Otherwise it's ready to merge. What do you say, @timmolter?

PR #1069, opened 2015-09-23
BTCMarkets implementation (most of the API).

mmazi:
Also added some Jackson (De)Serializers, and some other minor utilities, to xchange-core.


timmolter:
Nice!

PR #1084, opened 2015-11-04
Adds missing value vwap for Ticker from Kraken

swiesend:
The vwap value will be the 24h weighted average provided by the kraken API:
https://www.kraken.com/help/api#get-ticker-info


swiesend:
I have tested the change with the provided data in:
src/test/resources/marketdata/example-ticker-data.json

PR #1105, opened 2015-11-23
Poloniex: switch orderNumber to Long

nopy:
Poloniex orderNumber is a Long not an Integer. Since saturday ( 11/21/2015 ),
ETH Market order number overflow integer value.

### DIFF
--- a/xchange-poloniex/src/main/java/com/xeiam/xchange/poloniex/dto/trade/PoloniexTradeResponse.java
+++ b/xchange-poloniex/src/main/java/com/xeiam/xchange/poloniex/dto/trade/PoloniexTradeResponse.java
@@ -11,11 +11,11 @@ import si.mazi.rescu.ExceptionalReturnContentException;
 
 public class PoloniexTradeResponse {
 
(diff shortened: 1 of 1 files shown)


andre77:
running into the error with release 3.1.0
SNAPSHOT is actually not a very good option...
when is the next release expected including this bug fix ?

timmolter:
@andre77 please try 4.0.0-SNAPSHOT.

andre77:
Hi Tim, thanks for the reply
like i said, snapshot is not the best option
no plans to release the 4.x.x in the near future?

PS: is it possible to deploy the source artefacts next to the binary jars as well on the snapshot repo?
https://oss.sonatype.org/content/repositories/snapshots/com/xeiam/xchange/xchange-poloniex/4.0.0-SNAPSHOT/

timmolter:
I'd like to release 4.0.0 soon, but I'm currently short on time. Maybe I'll just cut a release soon without meeting all my goals for 4.0.0. You could create your own releases on your own fork if you wanted as well.

PR #1111, opened 2015-11-26
All Integration Tests Pass

baffo32:


baffo32:
I added code to allow btctrade to work with a certificate intended for 33option.com, which I needed to pass its integration test locally.  But if this isn't needed elsewhere then 33option.com is mitm'ing me and the code should be removed.

PR #1113, opened 2015-11-28
[core] [refactor] Make Currency into a class

baffo32:
- Code clarity by having type names represent content
- A way to represent multiple currency codes for the same currency
- A place to store currency-specific values, such as scale or symbol
- A way to enumerate all known currencies, which is used to handle edge cases in CurrencyPairDeserializer
- An alternative to java.utils.Currency which supports non-ISO 4217 currencies

I may not have made all the best specific implementation choices here, but I think using a class for this in general will help code in the future.


timmolter:
I like it. Might change `String symbol` to `String unicode`. Will the `Integer scale,` ever be used? I wouldn't think so.

baffo32:
I thought it would be a good place to store defaults when exchange apis do not provide a scale.  Assuming they would be using whatever the currency's native scale is.

PR #1116, opened 2015-12-02
Adding withdrawal support for bitfinex

Muffon:


timmolter:
Please check the failures on Travis CI. Looks like compile errors.

xchange-bitfinex/src/main/java/com/xeiam/xchange/bitfinex/v1/BitfinexUtils.java
```diff
@@ -16,7 +17,25 @@ private BitfinexUtils() {
 
   public static String toPairString(CurrencyPair currencyPair) {
 
-    return currencyPair.base.getCurrencyCode().toLowerCase() + currencyPair.counter.getCurrencyCode().toLowerCase();
+    return currencyPair.base.toString().toLowerCase() + currencyPair.counter.toString().toLowerCase();
```
baffo32: Judging by the `convertToBitfinexWithdrawalType` function below, I think bitfinex may actually want `.getCommonlyUsedCurrency().toString()` in places like this, so that "XBT" will be converted to "BTC".

That could also be done inside the `convertToBitfinexWithdrawalType` function with something like `Currency.getInstance(currency.toUpperCase()).equals(Currency.BTC)` which will cause XBT to compare equal to BTC

Issue #1101, opened 2015-11-19
Full Support for Services Providing Only a Subset of Features


baffo32:
This means
- No extra code is required by implementations for features they do not support
- Applications can query exchanges for what features they provide

I am interested in coding a solution to this issue.  I think it will require at least a little refactoring.

This could provide
- The supported featureset to expand without increasing maintenance and implementation work immediately
- Ease of supporting exchanges with poor featuresets
- The creation of generic applications which can query exchanges for feature support in order to intelligently consolidate the data they provide

From #578:

> Now that so many exchanges are supported, we could re-work the polling services so that when it is
> extended, not every method needs be implemented. A generic "method not implemented" exception
> could be thrown.
> 
> For instance, say someone is adding a new exchange implementation to XChange. They're doing the
> market data service implementation. The exchange doesn't have an API method for bulk tickers. The dev
> doesn't have to think about getBulkTickers(). They don't even mention it in their
> NewExchangePollingMarketDataService, and if an end user attempts to call the method, they just see
> that the method is not available for NewExchange

This would also help allow #1096 #1098 .

Ideas for how this might be implemented:
1. Base classes for exchange services (rather than interfaces)
   - takes e.g. EnumSet of supported features in constructor
   - provides default implementations which throw NotAvailableFromExchangeException
   - this new base class would break inheritance tree for exchanges where generic services inherit from raw services.  composition could be used instead, providing a getter.
2. Providing registerServiceFeature() on e.g. Exchange to register features as callbacks or add them to a list
3. other?

baffo32:
An additional thought is that many exchanges provide multiple datastreams, each which provides its own featureset.  A more thorough solution to this would allow the application to enumerate the datastreams, the featuresets each provides, and decide which ones to access.

timmolter:
I'm not convinced, but willing to hear you out more.

> No extra code is required by implementations for features they do not support

Yeah, but there's a lot of extra code needed to define what featureset each impl. provides.

> The creation of generic applications which can query exchanges for feature support in order to intelligently consolidate the data they provide. Applications can query exchanges for what features they provide

That's the main feature you're going for right?

> takes e.g. EnumSet of supported features in constructor

So the client has to pass an EnumSet? Seems weird.

> this new base class would break inheritance tree for exchanges where generic services inherit from raw services. composition could be used instead, providing a getter.

Which services inherit from the raw services? I didn't know there was anything like that.

baffo32:
[...]

PR #1117, opened 2015-12-03
Refactor AccountInfo

### LINKED ISSUE #1039: OkCoin.cn returning several wallets for same currency.
I opened Loan/Lend wallet at okcoin.cn (https://www.okcoin.cn/lend/lends.do)

I think this made some difference on getting wallet list.
I currently get 6 wallets.

xchange-okcoin

First wallet is CNY, and it has "available" string in description.
I get some more wallets with same CNY and "loan" string in description at index 5.

Current API is just returning one wallet by currency string and I can't get normal CNY wallet that has "available" description.

I can call getWallets() and retrieve it manually but why I'm posting issue is that I found description field is deprected.
description field is important to distinguish wallets in okcoin.cn

I'm not familiar with XChange yet and I'm not sure what is better solution. Just want to say description field must not be deprecated in this situation.

### LINKED ISSUE #1059: Question to Balance / Available / Frozen
I write a program to generate statistics for me about a bot I use. 
I check the Exchanges BTer, Poloniex, Cryptsy, Bleutrade and Bittrex.
Now, the first three exchange that I verified have not the correct values in this fields.
Question:
frozen == value that used in trades. Not available for work.
available == value available for work.
balance = frozen + avaliable.
Correct?
Bleutrade/Cryptsy/Poloniex have only available value in balance and available fields.

Sample Pononiex:
Web:
[WDC] Worldcoin 566.62130431 
On orders: 375.56498163 
BTC value: 0.02205658

XChange:
Wallet [currency=WDC, balance=566.62130431, available=566.62130431, frozen=0, description=]

Is this correct? I think no.

### LINKED ISSUE #1065: 300 mBTC Bounty - Clean up `AccountInfo`
figure out best Wallets solution. See:

* <https://github.com/timmolter/XChange/issues/1039>
* <https://github.com/timmolter/XChange/issues/1059>

Propose a good solution, address TODOs, remove deprecated code, refactor all implementations and example code as needed.

baffo32:
- Renamed `Wallet` to `Balance` to leave room for a `Wallet` class having the same meaning as that used in multi-wallet exchange apis
- Split wallet data out from `AccountInfo` into new `Wallet` class with a string id and name
- `AccountInfo` now has multiple `Wallet`s if needed
- Implemented additional fields in `Balance` to provide for variety of balances offered by exchanges
- Removed deprecated methods
- All integration tests pass

- fixes #1039 
- fixes #1059 although it was already some addressed

I did not implement a solution to the bitfinex situation (offers a different wallet for trading, lending, etc).  I think this is a separate issue.  Current XChange only supports trading anyway, so it makes sense to only present that wallet.

I did not implement an interface to query services for only specific wallets.  I think this is better done by the split interface solution discussed in #1101 .

I propose that this fixes #1065 .


timmolter:
Excellent! 

> I did not implement a solution to the bitfinex situation (offers a different wallet for trading, lending, etc). I think this is a separate issue. Current XChange only supports trading anyway, so it makes sense to only present that wallet.

I'd say just leave it as is because Bitfinex is the only exchange with this feature. People can use the raw API if they need it.

> I did not implement an interface to query services for only specific wallets. I think this is better done by the split interface solution discussed in #1101 .

I guess that's the next issue to tackle then, right?

baffo32:
> > I did not implement an interface to query services for only specific wallets. I think this is better done by the split interface solution discussed in #1101 .
>
> I guess that's the next issue to tackle then, right?

Yup.  I'm working on that at https://github.com/baffo32/XChange/tree/refactor .  It'll be a big job though.

mmazi:
@baffo32 Great job on the balances!

Perhaps another method on `Balance` would be useful: getAvailableForWithdrawal(). I'm guessing it could simply be `{ return available.subtract(borrowed); }`, but I'm not sure whether borrowed funds can be withdrawn. Or perhaps this is exchange-specific?

baffo32:
@mmazi Thanks.  I'm glad to implement that method for you.  I think it makes sense -- even if it's exchange-specific it's the safe default.

mmazi:
Merged. Thanks!

PR #1119, opened 2015-12-03
Bittrex: implements withdraw and deposit/withdraw history retrieval

nopy:


playscorer:
How can we call the getFundingHistory(TradeHistoryParams) ?

At the moment createFundingHistoryParams() throws NotAvailableFromExchangeException

PR #1123, opened 2015-12-07
Crypto Facilities Derivatives Platform Implementation

jclaruelle:
Implementation of Xchange API client for Crypto Facilities Derivatives
Market


xloem:
Code is based off of old tree; doesn't have `Wallet` change.

PR #1128, opened 2015-12-10
Bitstamp balance withdrawing

mmazi:
Two commits, two changes:
- Bistamp balance withdrawing amount
- Log a warning instead of an exception when balances don't add up. I don't think the situation when balances don't add up is critical (individual balances are still useful). Also, I suspect many exchange implementations currently have issues in the implementation here (like Bitstamp) so I suggest we fail more gracefully to allow us some time to fix these.

The code was tested with Bitstamp.

@baffo32, please review. Thanks!


baffo32:
Looks good.  Be ideal to update example-accountinfo-data.json so that the test will trigger this issue, and the warning somehow cause tests to fail.

mmazi:
Thanks for reviewing. I'll leave the test for a later time though.

PR #1147, opened 2015-12-22
All other currencies uppercase. Solves bug in okcoin when calling get…

kmels:
…Balance(Currency.BTC) in 4.0.0-SNAPSHOT


kmels:
Sorry, I amended the wrong commit. Please ignore this

PR #1219, opened 2016-03-31
Corrected meta data reported fee on bleutrade.

Achterhoeker:
Tested on a real trade. A trade has 0.25 % fee. (old assumption that fee was calculated for base and counter coin was incorrect)


timmolter:
Could you please take a look at the broken unit tests?

Achterhoeker:
@timmolter Done. Sorry for the failures. The reported meta data was also hardcoded tested.

PR #1225, opened 2016-04-07
Add back new order multi and cancel order multi

evdubs:
I am not sure why this was removed. It was perhaps a little quirky to get working initially because the list of orders for new order multi was < Order > and not <? extends Order>, forcing a cast, but I have corrected that. There are other diffs for formatting that can be reverted if desired. I would like to know why the code was commented out if anyone knows.


timmolter:
I don't know why the code was commented out. I'll keep a closer eye on it.

PR #1229, opened 2016-04-09
Retrieve Chart Data From Poloniex

madwicket:


timmolter:
Can you check on the failing checks please?

Issue #1063, opened 2015-09-21
100 mBTC Bounty - Wrap all exchange-specific Exceptions


timmolter:
... for all implementations. 

For an example of how it should be implemented: `BitfinexTradeServiceRaw.java`

rafalkrupinski:
Does it require having accounts on all the exchanges?

timmolter:
No. Just check all implementations and ones like `BitsoTradeService`, don't allow it to be thrown further, but wrap it in `ExchangeException`. Also, would there be a way to not allow people to be lazy and throw those exceptions like that?

rafalkrupinski:
I should be doable with reflections

rafalkrupinski:
I have a solution in mind. Should Rescu exceptions also be wrapped or just exchange specific exceptions declared in exchanges' REST interfaces?

timmolter:
Declared Rescu exceptions should be wrapped - if it's defined in the Rescu interface. Otherwise, I believe Rescu throws a generic IOException.

mmazi:
I don't think requiring every exchange implementation to include tons of try/catch boilerplate is a very good idea, for several reasons, including (but not limited to):
- There will be huge amounts of boilerplate (perhaps more than actual logic) which makes code unreadable and difficult to maintain.
- People just won't do it, or will do it wrong.

If this is indeed desirable ( @timmolter , could you state the intended purpose of this request?), I think we should find a general solution in XChange and leave the implementations clean.

I created a draft for discussion about general exception handling in XChange in #1233.

timmolter:
Closing in order to discuss this first more in depth. Thanks @mmazi !

PR #1230, opened 2016-04-09
ItBit: Update market data API endpoints.

kmels:
The endpoints for trades and order book have been updated. GetTrades() wraps the response in an object with a count now.

There are sometimes that the API returns 404 of course. Any pointer on how to bind these 404 responses to java classes or exceptions with rescu?..

### DIFF
--- a/xchange-itbit/src/main/java/com/xeiam/xchange/itbit/v1/ItBit.java
+++ b/xchange-itbit/src/main/java/com/xeiam/xchange/itbit/v1/ItBit.java
@@ -10,18 +10,18 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
 
 import com.xeiam.xchange.itbit.v1.dto.marketdata.ItBitDepth;
-import com.xeiam.xchange.itbit.v1.dto.marketdata.ItBitTrade;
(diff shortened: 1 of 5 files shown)


timmolter:
@mmazi Do you have a quick answer to his question about 404 responses?

mmazi:
I think this should be resolved in the context of general exception handling within XChange. (I don't think it's a rescu issue.)

I've wanted to propose some general solution for exception handing within XChange for a while, but never got down to do it. I'll try to write something down soon since is don't fully agree with the way this is proposed eg. in #1063. More on this to follow.

kmels:
Don't know much about rescu's proxy, but what if it returned the json data wrapped in an exception when it gets 404 responses (or any 4xx)?

mmazi:
Oh, perhaps I misunderstood the original question.

> Don't know much about rescu's proxy, but what if it returned the json data wrapped in an exception when it gets 404 responses (or any 4xx)?

That's exactly what rescu can do, and it's very easy to implement: just create a Jackson-mapped exception class (preferably by extending `HttpStatusExceptionSupport`) and declare it on the interface method(s). There are many examples of this in XChange, and here's the docs:
https://github.com/mmazi/rescu/wiki/Exception-handling

Don't hesitate to ask if you need further help implementing it.

PR #1234, opened 2016-04-12
Bugfix Core and additions to CoinbaseEx

bryan4887:


pom.xml
```diff
@@ -446,6 +446,33 @@
 					<autoVersionSubmodules>true</autoVersionSubmodules>
 				</configuration>
 			</plugin>
+
```
timmolter: @bryan4887 We removed this from here a while ago because building source and javadocs during a build takes such a long time, given the vast amount of modules in this project. Do you absolutely need this?

bryan4887: I see.  Is there another easy way for the javadocs and source to be included in projects that are importing XChange via maven and not directly including the checkout as a dependency in the project?

timmolter: I'll try to work on adding a profile that will act as a flag for building that stuff or not. More later...

bryan4887: Good thinking.  Thanks, Tim!

PR #1247, opened 2016-04-28
Bitstamp EUR support

mmazi:


mmazi:
Still doing some end-to-end testing, will merge soon.

PR #1254, opened 2016-05-20
Update PoloniexUtils.java

LifeIsGoodMF:
mismatch base and counter currency PoloniexUtils in toPairString and toCurrencyPair

### DIFF
--- a/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
+++ b/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
@@ -15,7 +15,7 @@ public class PoloniexUtils {
 
   public static String toPairString(CurrencyPair currencyPair) {
 
-    String pairString = currencyPair.counter.getCurrencyCode().toUpperCase() + "_" + currencyPair.base.getCurrencyCode().toUpperCase();
+    String pairString = currencyPair.base.getCurrencyCode().toUpperCase() + "_" + currencyPair.counter.getCurrencyCode().toUpperCase();
     return pairString;
   }
 
--- a/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
+++ b/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
@@ -22,7 +22,7 @@ public class PoloniexUtils {
   public static CurrencyPair toCurrencyPair(String pair) {
 
     String[] currencies = pair.split("_");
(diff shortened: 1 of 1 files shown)


timmolter:
Are you sure about this? I'm surprised no one every caught this yet?

LifeIsGoodMF:
Yes.
Simple test:
you send BTC_USD currency pair:

ExchangeSpecification spec = new
ExchangeSpecification(PoloniexExchange.class);
Exchange exchange = ExchangeFactory.INSTANCE.createExchange(spec);
PoloniexMarketDataService dataService =
(PoloniexMarketDataService)exchange.getPollingMarketDataService();
PoloniexDepth poloniexDepth =
dataService.getPoloniexDepth(CurrencyPair.BTC_USD);

but in request currency pair looks like USD_BTC:
https://poloniex.com/public?command=returnOrderBook&currencyPair=USD_BTC

2016-05-21 21:45 GMT+03:00 Tim Molter <notifications@github.com>:

> Are you sure about this? I'm surprised no one every caught this yet?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/timmolter/XChange/pull/1254#issuecomment-220794029>
>



-- 
Best regards
Viktor Zakalyuzhnyy

mmazi:
This is incorrect. This PR broke the Poloniex implementation. It is Poloniex who requires pairs to be composed as COUNTER_BASE, i.e. inversely to the custom, as is evident from [their API docs](https://poloniex.com/support/api/).

I will revert this commit shortly.

BTW, the USD_BTC pair in the above example doesn't even exist on Poloniex; they do have [USDT](https://poloniex.com/public?command=returnOrderBook&currencyPair=USDT_BTC) though.

LifeIsGoodMF:
You are right. Sorry, such behavior confused me.

2016-05-23 23:13 GMT+03:00 Matija Mazi <notifications@github.com>:

> This is incorrect. This PR broke the Poloniex implementation. It is
> Poloniex who requires pairs to be composed as COUNTER_BASE, i.e. inversely
> to the custom, as is evident from their API docs
> <https://poloniex.com/support/api/>.
>
> I will revert this commit shortly.
>
> BTW, the USD_BTC pair in the above example doesn't even exist on Poloniex;
> they do have USDT
> <https://poloniex.com/public?command=returnOrderBook&currencyPair=USDT_BTC>
> though.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/timmolter/XChange/pull/1254#issuecomment-221082374>
>



-- 
Best regards
Viktor Zakalyuzhnyy

PR #1259, opened 2016-05-28
Poloniex: retrieve all balances (including margin + lending balances)

anwfr:
added parameter "account='all'" for retrieving all balances (including margin + lending balances).

As described on https://poloniex.com/support/api/:
> (returnCompleteBalances) set the "account" POST parameter to "all" to include your margin and lending accounts.

### DIFF
--- a/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAuthenticated.java
+++ b/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAuthenticated.java
@@ -1,26 +1,19 @@
 package org.knowm.xchange.poloniex;
 
-import java.io.IOException;
-import java.math.BigDecimal;
-import java.util.HashMap;
-
-import javax.annotation.Nullable;
-import javax.ws.rs.Consumes;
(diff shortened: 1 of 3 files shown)


mmazi:
@anwfr : Just wondering, what is the purpose of the generic method `PoloniexAccountService::getAccountInfo()` now returning exchange account balances summed together with lending and margin balances? The generic interface is only concerned with exchange so this information not only seems unnecessary, but in fact incorrect, since now there is no more any way to get just the exchange balance via the generic API.

While adding the option of the "all" parameter to the `PoloniexAuthenticated` method is of course welcome, what is the reason to not let it be specified (but hard-coded to "all") in `PoloniexAccountServiceRaw::getWallets()`, and, moreover, using it *always* in the generic `getAccountInfo()` method?

anwfr:
hi,
Sorry for inconvenience, feel free to revert this behavior.
I see no reason to have this behavior in the generic method, I probably just didn't know that we could access it directly from a specific method such as PoloniexAccountServiceRaw::getWallets()
 Thanks

mmazi:
@anwfr Thanks for the reply.

Issue #1269, opened 2016-06-10
No date in ItBit trades (change in API: date->timestamp)


mmithril:
ItBit most probably changed its API – call https://api.itbit.com/v1/markets/XBTUSD/trades?since=0 returns

`{"recentTrades":[{"timestamp":"2016-06-10T09:29:47.5270000Z","matchNumber":698158,"price":"579.48000000","amount":"0.08000000"},{"timestamp":"2016-06-10T09:29:47.2130000Z","matchNumber":698157,"price":"579.48000000","amount":"0.04000000"},{"timestamp":"2016-06-10T09:27:04.4130000Z","matchNumber":698156,"price":"579.10000000","amount":"1.00000000"},{"timestamp":"2016-06-10T09:27:04.2400000Z","matchNumber":698155,"price":"579.00000000","amount":"0.01000000"},`

There is `timestamp` attribute (ISO 8601 format) while `org.knowm.xchange.itbit.v1.dto.marketdata.ItBitTrade` expects `date` (being `long` – most probably millis from era).

mmithril:
Fix in merge request #1270

PR #1270, opened 2016-06-10
Fix for  #1269

mmithril:
Two things fixed:
- ISO8601 parsing – if text ends with 'Z', it is always in UTC 
- change in itbit API – returns timestamp as ISO8601 text instead of date representing millis from start of era.


timmolter:
Could you take a look at the merge conflicts on this one? I don't want to fix the conflicts in the wrong way.

mmithril:
Fixed. I should create feature branch for it...

PR #1280, opened 2016-06-20
Added getExchangeSymbols()

zholmes1:
getExchangeSymbols() was completely inaccessible from any generic
service.


timmolter:
Please see: Exchange Metadata Update and Discussion #1282 

The exchange symbols should be accessed through the `Exchange` object.

PR #1311, opened 2016-10-04
Switched the base and counter in PoloniexUtils

kennethjor:
I kept getting errors about unknown currency pairs on Polonies. PoloniexUtils has the base and counter the wrong way around.

### DIFF
--- a/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
+++ b/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
@@ -20,7 +20,7 @@ public class PoloniexUtils {
 
   public static String toPairString(CurrencyPair currencyPair) {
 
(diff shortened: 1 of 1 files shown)


kennethjor:
@timmolter Actually I think this may actually have been correct in the first place. My application is now getting weird ticks from Poloniex like `BTC/ETH -> bid=0.02125297, ask=0.02129996`, which is obviously not true.

Looking at the API docs https://poloniex.com/support/api/ there's an example of the returnTicker output:
```
{"BTC_LTC":{"last":"0.0251","lowestAsk":"0.02589999","highestBid":"0.0251","percentChange":"0.02390438",
"baseVolume":"6.16485315","quoteVolume":"245.82513926"},"BTC_NXT":{"last":"0.00005730","lowestAsk":"0.00005710",
"highestBid":"0.00004903","percentChange":"0.16701570","baseVolume":"0.45347489","quoteVolume":"9094"}, ... }
```
The BTC,LTC pair is at 0.0251, which must mean that these pairs on the Poloniex API are in fact in `counter_base` format.

PR #1315, opened 2016-10-08
Huobi: conclude host parameters inside initServices

jheusser:
Make BitVC work again via exchangeSpecificParams `use_bitvc=true`. This must have broken with the introduction of `initServices`.


jheusser:
Merged directly as only a tiny change.

PR #1328, opened 2016-10-27
Added new exchange - YoBit

atlet:
- Order Book
- Trades


xchange-examples/pom.xml
```diff
@@ -344,6 +344,11 @@
 			<artifactId>xchange-livecoin</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
```
timmolter: use the `${project.version}` etc. here instead.

PR #1332, opened 2016-11-02
[independentreserve] fixed the ISO8601 timestamp parsing

andre77:
use the data type converter in JAXB, since JAXB must be able to parse ISO8601 date string according to the XML Schema specification


xchange-quadrigacx/src/main/java/org/knowm/xchange/quadrigacx/QuadrigaCxUtils.java
```diff
@@ -28,9 +30,11 @@ private QuadrigaCxUtils() {
    * @return
    */
   public static Date parseDate(String dateString) {
-
     try {
-      return DATE_FORMAT.parse(dateString);
+        // SimpleDateFormat is not thread safe, there fore create a new instance every time
```
andre77: sorry, this change is not part of this pull request any more, was my fault...

quadriga is now on the other pull request
https://github.com/timmolter/XChange/pull/1333

but to answer your question: sore, one can use thread local here, but i'm not sure whether its worth it...
is the creation of SimpleDateFormat that expensive ?

andre77:
@timmolter the PR is merged, but for some reason not part of the last SNAPSHOT version
maybe because the other PR triggered the build process, and the changes were not detected after that?
could you trigger the build process manually ?

PR #1333, opened 2016-11-02
[quadrigacx] fixed the timestamp parsing by setting the time zone to UTC

andre77:


xchange-quadrigacx/src/main/java/org/knowm/xchange/quadrigacx/QuadrigaCxUtils.java
```diff
@@ -28,9 +30,11 @@ private QuadrigaCxUtils() {
    * @return
    */
   public static Date parseDate(String dateString) {
-
     try {
-      return DATE_FORMAT.parse(dateString);
+        // SimpleDateFormat is not thread safe, there fore create a new instance every time
```
kennethjor: @andre77 Continuing the conversation from #1332 here.

I haven't tested how expensive `SimpleDateFormat`, but it does have to parse the expression. Doing so on every single time you want to parse a date is probably not a non-trivial amount of overhead.

Issue #1357, opened 2017-01-09
PollingTradeService.openOrders - currency pair param


dozd:
Hi Timmolter,

I looked at `TheRockTradeService.getOpenOrders` and it's not implemented because `currencyPair` is needed. It's same in `ItBitTradeService` - there are all currency pairs loaded. In `BitstampTradeService` there is a "hack" that uses `getExchangeSpecificParameters()`.

What's the correct approach for this? I'd like to make PR for `TheRockTradeService` to implement this method.

timmolter:
I created a demo class for The Rock called "TheRockMetaDataDemo". That's the best way to get the currency pairs for the exchange. Then you can use those pairs to fetch the open orders and combine all the results.

dozd:
Yeah, that's similar to `ItBitTradeService` implementation i pointed out. But my question was more about what's the best approach. The problem I see in this implementation is that API is called many times for retrieval of open orders - API limits can be a problem here (rock has 14 currency pairs). 

Better way is to have a possibility to pass the argument. But I don't like "the Bistamp" way of implementation either.

As it's 3rd exchange (I didn't check many others) that needs this argument - is it possibility to change the interface and add optional arguments to this method?

timmolter:
It would only be possible if the exchange (The Rock) offers an API endpoint that returns all open orders without specifying a currency pair. Do they?

dozd:
No they don't. But I meant changing of `PollingTradeService` interface as `OpenOrders getOpenOrders()` is not sufficient at least for three exchanges.

timmolter:
Ah, ok. So you're proposing to add an additional method in `PollingTradeService` like: `OpenOrders getOpenOrders(CurrencyPair c)`? Or perhaps even something like: `OpenOrders getOpenOrders(OpenOrdersParams p)` like for `getTradeHistory`.

dozd:
Yeah. `OpenOrdersParams` looks like good approach - I like it. If you're ok with that I can propose PR.

timmolter:
Yeah, i think that'd be good.

PR #1358, opened 2017-01-10
New method getOpenOrders(OpenOrdersParams)

dozd:
Hi,

Implemented new method as discussed in #1357. I've implemented it in TheRockExchange as a reference and tested it towards live API through already present "integration tests".

I had a fast look over various TradingServices and it will be common for many exchanges to have currency pair attribute. But for now, new method is returning previous getOpenOrders until someone wants it rewritten.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
@@ -13,6 +13,7 @@ import org.knowm.xchange.exceptions.NotAvailableFromExchangeException;
 import org.knowm.xchange.exceptions.NotYetImplementedForExchangeException;
 import org.knowm.xchange.service.polling.BasePollingService;
 import org.knowm.xchange.service.polling.trade.params.TradeHistoryParams;
+import org.knowm.xchange.service.polling.trade.params.orders.OpenOrdersParams;
 
 /**
  * <p>
@@ -32,7 +33,7 @@ public interface PollingTradeService extends BasePollingService {
 
   /**
    * Gets the open orders
-   * 
+   *
    * @return the open orders, null if some sort of error occurred. Implementers should log the error.
    * @throws ExchangeException - Indication that the exchange reported some kind of error with the request or response
    * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the requested function or data
@@ -40,11 +41,26 @@ public interface PollingTradeService extends BasePollingService {
    *         implemented
    * @throws IOException - Indication that a networking error occurred while fetching JSON data
    */
-  public OpenOrders getOpenOrders() throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException;
+  OpenOrders getOpenOrders() throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException;
+
+  /**
+   * Gets the open orders
+   *
+   * @param params The parameters describing the filter. Note that {@link OpenOrdersParams} is an empty interface. Exchanges should implement its
+   * own params object.
+   *
+   * @return the open orders, null if some sort of error occurred. Implementers should log the error.
+   * @throws ExchangeException - Indication that the exchange reported some kind of error with the request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the requested function or data, but it has not yet been
+   *         implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  OpenOrders getOpenOrders(OpenOrdersParams params) throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException;
 
   /**
    * Place a market order
-   * 
+   *
    * @param marketOrder
    * @return the order ID
    * @throws ExchangeException - Indication that the exchange reported some kind of error with the request or response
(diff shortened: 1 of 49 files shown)


dozd:
Added cancelOrder implementation for TheRock exchange. I didn't know that GitHub add this commit to this pull request :-/.

xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
```diff
@@ -32,19 +33,34 @@
...
+   *
    * @return the open orders, null if some sort of error occurred. Implementers should log the error.
    * @throws ExchangeException - Indication that the exchange reported some kind of error with the request or response
    * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the requested function or data
    * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the requested function or data, but it has not yet been
    *         implemented
    * @throws IOException - Indication that a networking error occurred while fetching JSON data
    */
-  public OpenOrders getOpenOrders() throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException;
+  OpenOrders getOpenOrders() throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException;
```
timmolter: why remove `public` from all these methods?

xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
```diff
@@ -113,14 +129,21 @@ public boolean cancelOrder(String orderId)
...
    */
-  public TradeHistoryParams createTradeHistoryParams();
+  TradeHistoryParams createTradeHistoryParams();
+
+  /**
+   * Create {@link OpenOrdersParams} object specific to this exchange. Object created by this method may be used to discover supported and required
+   * {@link #getOpenOrders(OpenOrdersParams)} parameters and should be passed only to the method in the same class as the createOpenOrdersParams
+   * that created the object.
+   */
+  OpenOrdersParams createOpenOrdersParams();
```
timmolter: I'm not convinced we need this, since 99% of implementations return null. I'll recomment here once I'm completely through.

xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
timmolter: I was expecting that `getOpenOrders()` would cease to exist and be replaced by `getOpenOrders(OpenOrdersParams params)`.

dozd: Visibility modifiers have no effect in the interfaces so they were meaningless.

dozd: Yeah but that will break the API for users. What about marking getOpenOrders {@Deprecated} and mark for removal in the future?

xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
dozd: I'll remove it. I was following `TradeHistory` pattern but I agree with you.

xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
timmolter: TIL

timmolter: Yes.

dozd:
Sorry that I added two more commits for ItBit to this. Should I remove them and make separate PR?

timmolter:
In the future, would be nice, but it's OK for now. :)

xchange-core/src/main/java/org/knowm/xchange/service/polling/trade/PollingTradeService.java
timmolter: I think it might actually be a good thing. It forces the implementor, not the user of the exchange to think things out. in the end it's easier for the end user. What do you think?

timmolter: There really should be another `OpenOrdersParams` impl. that wraps the orderID and all (most) exchanges should be converted to that.

dozd: So you mean leaving it here? I made a mistake in `TheRockTradeService` not override returning object to `TheRockOpenOrdersParams` from `OpenOrdersParamCurrencyPair`. That will make things easier for user. I'll correct that.

Most exchanges use no parameters so I don't know what's good default. Null looked good for me.

timmolter: > There really should be another OpenOrdersParams impl. that wraps the orderID and all (most) exchanges should be converted to that.

Disregard that. I was getting that mixed up with something else.

dozd: So what's the final word - leave `createOpenOrdersParams` returning `null` or remove it?

timmolter: Go ahead and delete it. Just leave `getOpenOrders()` not deprecated too.

dozd:
Ok, i removed method `createOpenOrdersParams` and removed commits not related to openOrders (i create separate PR).

dozd:
@timmolter I have found that `createOpenOrdersParams` and `createTradeHistoryParams` is pretty good pattern and I'm missing that removed `createOpenOrdersParams` as a user of library.

Imagine you want to call `pollingTradeService.getOpenOrders` and don't know on which exchange. With `createOpenOrdersParams` you can call it like that:

```
PollingTradeService pollingTradeService = exchange.getPollingTradeService();
OpenOrdersParams openOrdersParams = pollingTradeService.createOpenOrdersParams();

if (openOrdersParams instanceof OpenOrdersParamCurrencyPair) {
  ((OpenOrdersParamCurrencyPair) openOrdersParams).setCurrencyPair(currencyPair);
}
// more if there for another params

return pollingTradeService.getOpenOrders(openOrdersParams);
```

You can add more IFs when you know that there are exchanges that need another params. Without `createOpenOrdersParams` i don't know what class i must create for the `getOpenOrders` method.

So it looks like i should follow pattern from trade history and move `createOpenOrdersParams` back. Or do you have any other suggestions?

timmolter:
I'm traveling. Will get back to you soon. Thanks for your efforts!!

timmolter:
@dozd Yeah, that makes sense. Go for it!

Issue #1298, opened 2016-08-16
200 mBTC Bounty - The Rock Trading not fully implemented


ICFiedler:
TheRock improvements:

please implement getOpenOrders, cancelOrder as well as getTradeHistory methods. It also seems that the current The Rock implementation still uses the old API address and not the new and more powerful API.

ICFiedler:
I would like to offer a 0.2BTC bounty on this one. How can I do this?

timmolter:
I've added the bounty tag and updated the title. If someone wants to complete this for the bounty, first claim it here and indicate you will work on it. Then @ICFiedler should send us (Knowm Inc.) the BTC and we will hold it until the code is complete and passes all functionality and quality checks. We will then forward the funds to the person carrying out the bounty.

zholmes1:
Is this bounty still available?

ICFiedler:
The devs of The Rock have taken care of the issue without claiming the bounty. But it is not yet fully delivered. Some small percent are still missing. I would not give 200mbt for it, though. I will come back to this with either a to-do description and a new bounty or the solution in 2-3 days.

ICFiedler:
Update: their API does not work with XChange: it needs a call to get all open orders. The Rock API only allows calls to query the open orders of one specific pair. So without them having their API changed things will not work. That will also make the boutny obsolete.

dozd:
Or better to change the XChange API as currency pair is needed on many exchanges for open orders ;-). Submitted PR.

npomfret:
Can this be closed?  Or is there still work need doing?  And is any bounty available?

timmolter:
Closing due to no response.

PR #1360, opened 2017-01-11
#1298 Implemented cancelOrder method for TheRock exchange.

dozd:


timmolter:
can you check the build errors?

PR #1367, opened 2017-01-24
LimitOrder.Builder does not set order status properly

dozd:
I've found that LimitOrder.Builder does not reflect status set to the builder when is building the order.

Is it bug? Attached failing unit test. Can attach fix if agreed.


timmolter:
Seems like a bug to me.

dozd:
Ok, i'll fix that. I don't get why this class is half immutable, half mutable.  I would make it whole immutable.

PR #1371, opened 2017-01-27
Initial support for trading futures on BitVc

jheusser:
Supporting trading futures on BitVc. Placing limit orders, open positions, and account info implemented to start with.


jheusser:
I'll fix that conflict later on.

PR #1382, opened 2017-02-06
Link to xchange-stream added.

dozd:
Can I have this small kind of advertisement? I've release first version 4.2.0 based on your release.


timmolter:
Definitely! Actually I'm really happy someone took on the challenge of making a nice streaming API. Good work!

PR #1384, opened 2017-02-08
introduce Coinfloor implementation

whitslack:
This introduces a Coinfloor module that adapts Coinfloor's official [Java client library](https://github.com/coinfloor/java-library) to XChange's API.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -91,6 +91,7 @@
 		<module>xchange-cexio</module>
 		<module>xchange-chbtc</module>
 		<module>xchange-coinbase</module>
+		<module>xchange-coinfloor</module>
 		<module>xchange-coinmate</module>
 		<module>xchange-core</module>
 		<module>xchange-cryptofacilities</module>
--- a/xchange-coinfloor/pom.xml
+++ b/xchange-coinfloor/pom.xml
@@ -1,0 +1,41 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.knowm.xchange</groupId>
+    <artifactId>xchange-parent</artifactId>
+    <version>4.2.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>xchange-coinfloor</artifactId>
+
+  <name>XChange Coinfloor</name>
+  <description>XChange implementation for the Coinfloor exchange</description>
+
+  <url>http://knowm.org/open-source/xchange/</url>
+  <inceptionYear>2012</inceptionYear>
+
+  <organization>
+    <name>Knowm Inc.</name>
+    <url>http://knowm.org/open-source/xchange/</url>
+  </organization>
+
+  <!-- Parent provides default configuration for dependencies -->
+  <dependencies>
+
+    <dependency>
+      <groupId>org.knowm.xchange</groupId>
+      <artifactId>xchange-core</artifactId>
+      <version>4.2.1-SNAPSHOT</version>
+    </dependency>
+    
+    <dependency>
+      <groupId>uk.co.coinfloor</groupId>
+      <artifactId>coinfloor-library</artifactId>
+      <version>1.0</version>
+    </dependency>
+
+  </dependencies>
+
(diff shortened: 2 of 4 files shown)


timmolter:
Please check on build failures. Also, would you consider splitting up the `CoinfloorExchange` class into separate classes like all the other exchange implementations?

whitslack:
> Please check on build failures.

Build failures occur because a dependency, `uk.co.coinfloor:coinfloor-library:1.0:jar`, is not available from any public artifact repositories. I was assuming users would fetch the sources from GitHub and build/install the artifact themselves. If that is unacceptable, then what do you suggest?

> Also, would you consider splitting up the `CoinfloorExchange` class into separate classes like all the other exchange implementations?

I don't understand. It does comprise separate classes. Do you want me to make separate source files?

timmolter:
## uk.co.coinfloor:coinfloor-library:1.0:jar

Yes, that's unacceptable. The project needs to build without any manual steps, both for devs but especially for CI and building/deploying xchange to Maven Central. Can you ask the provider of that jar too upload the jar to Maven Central? Otherwise, assuming an appropriate license, do you have access to the source code? You could paste it into xchange. 

## Separate Classes

Yes, separate source files.

whitslack:
> Can you ask the provider of that jar too upload the jar to Maven Central?

I am the author. I am unfamiliar with the process of uploading artifacts to Maven Central, but I will try to figure it out.

dozd:
@timmolter I started Coinfloor implementation but stopped because I decided I don't need this exchange in my project. That library is not needed at all. Just @whitslack does not follow any pattern in XChange library. I think this kind of PR is unacceptable.

@whitslack Go through https://github.com/timmolter/XChange/wiki/New-Implementation-Best-Practices and look at another exchange implementations. All exchanges I saw follow more or less these guidelines. Your commit is like from another universe in contrast to them.

timmolter:
Thanks @dozd, I agree. 

@whitslack in order to get your contribution into XChange, it needs to conform to the normal pattern of all the other exchanges. Someone familiar with XChange should be able to open up your module and be very comfortable with poking around and seeing everything is where it should be. I didn't even get to that point yet as all the code was stuffed into one class file. Depending on another jar is a bad idea because it will make maintainability almost impossible. Someone originally said that the coinfloor API was like the Bitstamp one. I would suggest cloning the Bitstamp module and adapting it to the coinfloor API.

whitslack:
@timmolter: Coinfloor offers a Bitstamp emulation layer, but it is necessarily less efficient than the native WebSocket API. I am implementing this XChange module in the most efficient way I can. I am presently in the process of splitting the code into various source files, though this is causing a lot of abstraction leaks (methods forced to become public that shouldn't be).

> Depending on another jar is a bad idea because it will make maintainability almost impossible.

This doesn't make sense. One can depend upon a specific version of an external artifact. That would never change.

@dozd: Coinfloor's native API is not REST; it is WebSocket. The "raw" implementation is simply the Coinfloor Java client library. A cooked implementation is what I am contributing to XChange.

timmolter:
It looks like there is a native HTTP (REST) API here: <https://github.com/coinfloor/API/blob/master/BIST.md>

Does your jar use the streaming API of Coinfloor and somehow turn it into a REST API? Why not go directly to Coinfloor's HTTP API?

>This doesn't make sense. One can depend upon a specific version of an external artifact. That would never change.

Exchanges change their APIs constantly. If we cannot for example tweak a URL param or adjust our DTOs to match returned JSON (because it's buried somewhere in a 3rd party jar), then the coinfloor module becomes broken and impossible to fix unless the provider of that jar does an update, which cannot be counted on.

whitslack:
> It looks like there is a native HTTP (REST) API here: https://github.com/coinfloor/API/blob/master/BIST.md

That's the emulation layer. It's built atop the WebSocket API.

> Does your jar use the streaming API of Coinfloor

Yes.

> and somehow turn it into a REST API?

I don't know what this would mean. My XChange module connects to Coinfloor's WebSocket API and submits appropriate commands to it when XChange API methods are called. Some XChange API methods (e.g., `MarketDataService.getTicker(CurrencyPair)`) require no commands to be sent to the WebSocket API since the information is available locally, having been cached as updates are received from the WebSocket.

> If we cannot for example tweak a URL param or adjust our DTOs to match returned JSON (because it's buried somewhere in a 3rd party jar), then the coinfloor module becomes broken and impossible to fix unless the provider of that jar does an update, which cannot be counted on.

I see. In that case, Coinfloor will host this XChange module themselves. (This is what I wanted to do from the beginning, but the CTO asked me to contribute the module to the XChange project for better visibility.)

PR #1386, opened 2017-02-16
Get all open orders on itbit of no pair is given

Achterhoeker:


dozd:
I would not include this kind of logic into the library (to keep it simple and consistent). You can easily call getOpenOrders multiple times from your application.

Achterhoeker:
@dozd, all exchanges give all open orders if no attributes are given. Itbit is the only exchange that is not doing this. If that shouldn't be done, it would be a api change from the past. (always had been this way)

PR #1402, opened 2017-03-15
BitcoinCore RPC interface to access wallet balance

ww3456:
I'm not sure whether this is something that fits into XChange or not but I wanted something that would fetch a BitcoinCore wallet balance in the same way as an exchange. Thoughts?


timmolter:
Yeah, this is great!

PR #1409, opened 2017-03-23
Bitfinex errors solved and added new coins

Achterhoeker:
Bitfinex order numbers are now to high to fit in a signed int. (resulting in errors at placing orders and retrieving open orders. Error at place order is a real problem, the order is executed, but result results in an error)
Increased order number to long type.
Added new coins, also conversion of dsh to dash

### DIFF
--- a/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/BitfinexAdapters.java
+++ b/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/BitfinexAdapters.java
@@ -48,6 +48,13 @@ public final class BitfinexAdapters {
 
   }
 
+  public static String adaptBitfinexCurrency(String bitfinexSymbol) {
+	  String currency = bitfinexSymbol.toUpperCase();
+	  if (currency.equals("DSH")) {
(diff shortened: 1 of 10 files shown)


ezhik97:
Why you use double for order num instead long?

Achterhoeker:
it was a mistake from me. Latest is already based on a long. Op 24 mrt. 2017 8:10 a.m. schreef ezhik97 <notifications@github.com>:Why you use double for order num instead long?

ezhik97:
Also it is needed to fix BitfinexTradeServiceRaw:
new BitfinexCancelOrderRequest(String.valueOf(exchange.getNonceFactory().createValue()), Integer.valueOf(orderId)));

orderId is not parse into Integer.

Achterhoeker:
@ezhik97 thanks, that was indeed also needed.

andre77:
in the class BitfinexTradeServiceRaw.java, there are still expressions like 
Integer.valueOf(offerId)

please replace the Integer with Long

Achterhoeker:
@andre77 you are right. I had now some more time to clean up the mess i made. Are you ok now with the latest state ? I think i've now transformed all id's into longs.

PR #1411, opened 2017-03-24
changed the bitfinex identifiers from int to long due to overflow > I…

andre77:
changed the bitfinex identifiers from int to long due to overflow > Integer.MAX_VALUE 2147483647


andre77:
looks like there is already a PR for that
https://github.com/timmolter/XChange/pull/1409
so this one can be declined

PR #1437, opened 2017-04-14
change the bitfinex ids from int to long

andre77:


timmolter:
there's a build failure.

PR #1449, opened 2017-04-27
removed javax.xml.bind. this will allow xchange-bittrex to be used on…

zholmes1:
… Android

### DIFF
--- a/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/v1/BittrexUtils.java
+++ b/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/v1/BittrexUtils.java
@@ -1,10 +1,9 @@
 package org.knowm.xchange.bittrex.v1;
 
-import java.util.Calendar;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/v1/BittrexUtils.java
```diff
@@ -26,10 +31,13 @@ public static String toPairString(CurrencyPair currencyPair) {
...
 
-  public static Date toDate(String date) {
-    Calendar cal = DatatypeConverter.parseDateTime(date);
-    cal.setTimeZone(TimeZone.getTimeZone(TIMEZONE));
-    return cal.getTime();
+  public static Date toDate(String dateString) {
+    try {
+      return DATE_PARSER.parse(dateString);
+    } catch (ParseException e) {
+      e.printStackTrace();
```
ww3456: I think it would be best to use a SLF4J logger here rather than print to stderr.

timmolter: Agreed.

PR #1453, opened 2017-05-02
Fix Kraken GNO parsing

### LINKED ISSUE #1452: Kraken GNO pairs are parsed incorrectly
The newly added Gnosis (GNO) does not have a X prefix (like all the others except DASH and USDT), and so gets mangled somewhere, causing incorrect pairs such as NOE/TH and NOX/BT to be created.

Here's the API call for asset pairs: https://api.kraken.com/0/public/AssetPairs

For DASH I found these two PR's:
https://github.com/timmolter/XChange/pull/1435
https://github.com/timmolter/XChange/pull/1438

I'm not sure if anything else needs fixing + some Kraken tests are failing to begin with, so I'd rather see if someone else does it.

HairyFotr:
Fixes #1452.

The existing code seems full of similar hacked corner cases, I'm sorry for one more.

### DIFF
--- a/xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenAdapters.java
+++ b/xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenAdapters.java
@@ -167,9 +167,10 @@ public class KrakenAdapters {
   }
 
   public static CurrencyPair adaptCurrencyPair(String krakenCurrencyPair) {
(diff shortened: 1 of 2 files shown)


timmolter:
Yeah, Kraken causes some ugly hacks for sure.

Achterhoeker:
@HairyFotr @timmolter I agree that the Kraken code base is getting uglier by all those hacks. Perhaps a redesign should be done. I expect that the asset pair list from Kraken is a start point. In this list the alt name seems always clean from the X or Z adds.

PR #1455, opened 2017-05-02
Add convenience methods to ExchangeFactory

HairyFotr:
Noticed these convenience methods could reduce the initialization boilerplate by a few lines for most users.

But this PR actually stems from a separate issue I ran into. By mistake I set the API keys only after constructing the `Exchange`, which resulted in a signing exception, since (it took me a while to find that) during initialization the secret key was not yet set when it was passed to `KrakenDigest` in the `KrakenBaseService` constructor. Other exchanges likely have the same issue.

A possible solution would be to create a `ExchangeSpecificationBuilder` class that returns an immutable `ExchangeSpecification`. That way I couldn't call a setter after the value had already been used, and it would allow moving the default specification out of the instance method, which may then enable simplification of `Exchange` construction.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/ExchangeFactory.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/ExchangeFactory.java
@@ -70,6 +70,71 @@ public enum ExchangeFactory {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


timmolter:
Good call. The builder idea would also be good!

PR #1471, opened 2017-05-17
[Huobi] Update api url

nxtpool:
Use correct api url. "http://market.huobi.com/staticmarket" is obsolete.


nxtpool:
https://github.com/huobiapi/API_Docs_en/wiki/REST-Candlestick-Chart

PR #1486, opened 2017-05-29
[Jubi] - allow to uce since parameter in get trade history method

nxtpool:


nxtpool:
Error not with my commit. It's about bleutrade exchange not jubi.

PR #1488, opened 2017-05-29
Implemented funding history for DSX

andre77:


andre77:
@Mikelle the package should be renamed from
org.known.xchange.dsx.service
to
org.knowm.xchange.dsx.service

PR #1490, opened 2017-05-30
xchange-okcoin: Use new URIs for okcoin.com futures

jheusser:
Support for okex.com, the futures exchange endpoint of okcoin.com.

Let's see if this is enough, or if okex.com requires its own module at some point.

https://github.com/timmolter/XChange/issues/1484


jheusser:
*Please wait with merge*

PR #1495, opened 2017-05-31
[bittrex] added support for paymentid (withdrawals)

andre77:


andre77:
@timmolter probably it would be a good idea to add the payment id to the method
org.knowm.xchange.service.account.AccountService.withdrawFunds

may be like this
```java

  default String withdrawFunds(Currency currency, BigDecimal amount,
          String address) throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException {
      return withdrawFunds(currency, amount, address, null);
  }
  
  String withdrawFunds(Currency currency, BigDecimal amount,
          String address, String paymentId) throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException;
```

PR #1496, opened 2017-06-01
[bitstamp] attempt to handle currencies in a more generic way

andre77:


andre77:
since bitstamp introduces new pairs including the xrp currency, current implementation does not work, if you try to fetch user trades, so i decided to rewrite the code, to be more robust if it comes to new currencies

PR #1515, opened 2017-06-10
[Poloniex] base/counter instead of counter/base

ikosyanenko:


timmolter:
On vacation. Will reply/ merge later...

xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexUtils.java
```diff
@@ -21,7 +21,7 @@
 
   public static String toPairString(CurrencyPair currencyPair) {
 
-    String pairString = currencyPair.counter.getCurrencyCode().toUpperCase() + "_" + currencyPair.base.getCurrencyCode().toUpperCase();
+    String pairString = currencyPair.base.getCurrencyCode().toUpperCase() + "_" + currencyPair.counter.getCurrencyCode().toUpperCase();
```
andre77: @ikosyanenko 
@timmolter 

what is this change about???
its rolled back in the current version, but still... not necessary stuff

timmolter:
@andre77 I already reverted it here: 55658f3a8a933c1dde2f92ec5a7e835584985a58
Please let me know if there are still issues...

ikosyanenko:
Hi guys,

Poloniex needs to get symbol like BTC_XMR not XMR_BTC

Thanks,
Ivan Kosianenko

andre77:
exactly my friend, poloniex has the pair BTC_XMR, where XMR is the base and BTC is the counter currency!!!
so the schema for poloniex is COUNTER_BASE! The implementation was correct, and you have broken it...

timmolter:
For Poloniex use BTC_XMR. This is the convention used for all modules in XChange, and we need to be consistent.

Issue #1231, opened 2016-04-10
Margin Lending Support


abelgo:
Do we have any thing that supports this at the moment. ?  
So we have sort of support for exchanges whit more than one wallet in AccountInfo, But then in the PolingAccountService we don't have and calls in the interface to do transfers between different wallets.

I saw the dto for LoneOrders but not sure that is usable with the PolingTradeService at this point ?

douggie:
okcoin support futures but this is handeled a little differently by the exchange as okcoin treats them as a seperate product within the same account.

Margin trading on other exchange support the same product funded from a different 'margin account'

Keen to get this working for poloniex.

Not sure what Tim thinks but may be something along the following lines

1) add a flag to the order type like 'marginFunded' in the order interfaces that will tell the exchange implementation to place a margin trade such as marginBuy on poloniex rather than just a buy.

marginBuy
Places a margin buy order in a given market. Required POST parameters are "currencyPair", "rate", and "amount". You may optionally specify a maximum lending rate using the "lendingRate" parameter. If successful, the method will return the order number and any trades immediately resulting from your order. Sample output:

{"success":1,"message":"Margin order placed.","orderNumber":"154407998","resultingTrades":{"BTC_DASH":[{"amount":"1.00000000","date":"2015-05-10 22:47:05","rate":"0.01383692","total":"0.01383692","tradeID":"1213556","type":"buy"}]}}

rather than a normal buy
buy
Places a limit buy order in a given market. Required POST parameters are "currencyPair", "rate", and "amount". If successful, the method will return the order number. Sample output:

{"orderNumber":31226040,"resultingTrades":[{"amount":"338.8732","date":"2014-10-18 23:03:21","rate":"0.00000173","total":"0.00058625","tradeID":"16164","type":"buy"}]}

2) add additional methods to account interfaces to support methods to move, add, withdrawal from margin account.

timmolter:
Feel free to add any functionality you'd like at the "Raw" level. If it's commonly used across multiple exchanges, we can talk about adding it tot the "generic" level.

douggie:
Hi Tim,

Quite a few exchanges support the concept of 'lending' a balance for your
to trade with, aka margin.

here are a couple I am aware of.

*bitfinex*
https://www.bitfinex.com/pages/howitworks
*poloniex*
https://poloniex.com/support/aboutMarginTrading/
*Kraken*
https://support.kraken.com/hc/en-us/sections/200560633-Leverage-and-Margin

[...]

PR #1517, opened 2017-06-11
#1231

douggie:


timmolter:
it would be nice if you updated your code formatter to match the [XChange Style](https://github.com/timmolter/XChange/wiki/Code-Style) so that the diffs are more readable.

douggie:
Hi Tim, so sorry since the projects moved to knowm I forgot to add back the project specifc formatters. should be neater now.

Issue #1528, opened 2017-06-17
Poloniex Base and Counter Currency Pair issue


SeanRobb:
It looks like the base and counter currency pair have changed in commit 62e027a54ed641ebb545f47ae0b4622882367438.

I am now having issues with retrieving Ticker information from Poloniex. The api call works well but the values are way off.  I was wondering why this was done?

timmolter:
Whoops. That's happened before and I forgot about it. Does anyone know why people keep submitting PRs with the bas and counter switched for poloniex?

douggie:
Nope, I am using polinex without issue the base and counter are the right
way round.

timmolter:
@douggie With the latest develop branch including the last change to Poloniex?

douggie:
I am one comit behind. Will check the number a bit later.

SeanRobb:
I started having the problem yesterday and it went away when I rolled back to a different snapshot of the same version in the project so I looked in the recent commits and saw that this was the only change.  Thank you @timmolter for your quick response and I just wanted to make you aware of the change that appears to return incorrect ticker values.

odaken:
I vote for @SeanRobb. I am also suffered getting ticker with error by the last change. I am getting tickers from several exchanges but the change was inconsistent with other exchanges.

douggie:
Hi,

I am using the branch with the last comitt of
3c0dcb07f2584609c1b86e3d04184c0db601690b
without issue.

douggie:
PR https://github.com/timmolter/XChange/pull/1533 submitted

timmolter:
OK, I'm going to look into this today and finalize a solution....

douggie:
I can't see why the need to reverse them. I thought it might be that
polonix provides inverse currencies i.e. eth.btc as well as btc.eth but
that does not seem the case.

timmolter:
The thing is, I keep getting pull requests that keep flipping the base and counter. I believe from what I saw, you are correct, but I want to 100% verify it and convince myself so I can correctly handle future pull requests that flip things.

timmolter:
Addressed in commit: 55658f3a8a933c1dde2f92ec5a7e835584985a58

douggie:
I think the comit has broken USD tether as we now have :

currencyPair.counter.getCurrencyCode().replace("USD","USDT").toUpperCase()

Submitted a PR to fix
Support currencies:
https://poloniex.com/public?command=returnCurrencies

[...]

PR #1533, opened 2017-06-19
#1528

douggie:


timmolter:
Thanks @douggie I already pushed this in: 55658f3a8a933c1dde2f92ec5a7e835584985a58

PR #1539, opened 2017-06-19
#1528 Reverted support for USD Tether

douggie:


andre77:
there is a PR already for this, see https://github.com/timmolter/XChange/pull/1537

PR #1545, opened 2017-06-21
fix https://github.com/timmolter/XChange/issues/1499

npomfret:
removed the static state from the YoBitOrderBookDeserializer which was causing orders from different markets to appear in the same order book


npomfret:
Hmm, my pr seems to have @timmolter's commits in it?  Not sure what I've done there.

timmolter:
@npomfret Could you close this and submit a new PR. You probable just need to get all synced up first.

npomfret:
Sure, have something else to add anyway

npomfret:
Hmm, I'm a bit stuck.  No matter what I do git wants to include those commits...

timmolter:
I'm not sure how to fix it. Maybe google has an answer?

PR #1546, opened 2017-06-21
2 fixes

npomfret:
* another change to prevent the destruciton of the static config
 * fixed https://github.com/timmolter/XChange/issues/1499


npomfret:
FFS

PR #1558, opened 2017-06-25
[dsx] fixed dsx trade amount field

andre77:


andre77:
@Mikelle for some reason the tradable amount for a trade comes in the field "volume", but not in the field "amount" like its described in the documenation, see https://api.dsx.uk/#reference/reports-and-history/trades-history/get-trades-request

could you please have a look at this, it seems to me you have a direct link to the dsx development team

Mikelle:
@andre77 It was mistake in api docs. Response should come with field "volume". I fixed docs.

PR #1561, opened 2017-06-26
[bitstamp] added LTC meta info

andre77:


andre77:
@Achterhoeker 
could you please have a look at the LTC values, i'm not 100% sure about it
thanks in advance

Achterhoeker:
@andre77 Bitstamp minimal order size is 5 usd. So minimal order size on current ltc prices is about 0.125. The 0.1 you gave to usd and eur are not enough. My advise would be to change them to 0.15 (some spare room if ltc prices is decreasing). The 200 for the ltc/btc pair is way to high. This can also be set to 0.15.

andre77:
@timmolter please give me a minute to make changes listed by @Achterhoeker

andre77:
ok, it should be fine now

PR #1564, opened 2017-06-27
[bitstamp] avoid an exception in case of unknown withdrawal request type

andre77:


andre77:
the documentation for "withdrawal type" at bitstamp is a bit incomplete...
it sais: 0 (SEPA), 1 (bitcoin), 2(WIRE transfer), 15 (litecoin).
but we've seen types 6, 7, 14
you can only guess the type... so imho its better to add and deal with "unknown" type than an exception at runtime

PR #1566, opened 2017-06-27
[ANX] implemented funding history for ANX

npomfret:


timmolter:
Thanks @npomfret 

In the future can you label your commits with the exchange name in the beginning such as:

[ANX] implemented funding history for ANX

npomfret:
Will do.

PR #1577, opened 2017-07-02
[kraken] added quick fix for EOS currency

andre77:
The way, how we deal with currencies at kraken is not the most elegant one...
I just added small changes to be able to handle the EOS currency properly. A quick and dirty way...
In the future maybe we should refactor the whole currency thing at kraken, maybe using this API information https://api.kraken.com/0/public/AssetPairs


timmolter:
Yeah, the Kraken code is hacky. Too bad Kraken makes up their own symbols. The `AssetPairs` symbols is used (99% sure) when generating the exchange metadata.

PR #1582, opened 2017-07-03
[hitbtc, bittrex, bluetrade] improvements to getTradeHistory and getOpenOrders

npomfret:
[bittrex] made getTradeHistory optionally support TradeHistoryParamCurrencyPair [bittrex] made getBittrexOpenOrders support currency pairs that already have BTC as the base currency (like BTC_XRP) 

[hitbtc] made getTradeHistory optionally support TradeHistoryParamPaging and TradeHistoryParamCurrencyPair

[bluetrade] implemented TradeHistoryParamCurrencyPair and changed the default trade history so it doens't return cancelled trades (as this isn't really supported in the UserTrade class)


npomfret:
thanks again @timmolter !

PR #1586, opened 2017-07-06
[Jubi] Add getAllTickers interface to fetch all e-coins' tickers by one request

DrawFun:
1. Move MarketData RestProxy from BaseService down to JubiMarketDataServiceRaw, which allow us to decouple different proxies(market data, account, trade...) in the next development. 
2. Add getAllTickers interface to fetch all e-coins' tickers by one restful request (Path：/api/v1/allticker/).

### DIFF
--- a/xchange-jubi/src/main/java/org/knowm/xchange/jubi/service/JubiMarketDataService.java
+++ b/xchange-jubi/src/main/java/org/knowm/xchange/jubi/service/JubiMarketDataService.java
@@ -13,7 +13,6 @@ import org.knowm.xchange.jubi.dto.marketdata.JubiTicker;
 import org.knowm.xchange.service.marketdata.MarketDataService;
 
 import java.io.IOException;
-import java.util.Map;
 
 /**
  * Created by Yingzhe on 3/17/2015.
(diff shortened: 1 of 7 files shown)


xchange-jubi/src/main/java/org/knowm/xchange/jubi/service/JubiMarketDataService.java
```diff
@@ -25,16 +26,18 @@ public JubiMarketDataService(Exchange exchange) {
...
+  public Ticker getTicker(CurrencyPair currencyPair, Object... args)
+          throws ExchangeException, NotAvailableFromExchangeException, NotYetImplementedForExchangeException, IOException {
     // Request data
     JubiTicker jubiTicker = getJubiTicker(currencyPair.base.getCurrencyCode(), currencyPair.counter.getCurrencyCode());
-
     // Adapt to XChange DTOs
     return jubiTicker != null ? JubiAdapters.adaptTicker(jubiTicker, currencyPair) : null;
   }
 
+  public Map<String, Ticker> getAllTicker(Object... args) throws IOException {
```
timmolter: Let's keep this out of the generic service for now. To access this, just cast to the raw service. in the future, if we add a method like this to the generic service, the map should probably be `<CurrencyPair, Ticker>`.

DrawFun: that's right not to mix some specific functions into generic services. i'll move them out.

PR #1592, opened 2017-07-08
Removing Logic added in Bittrex Utils at 916f3b32647f4d3255b5907a4d266d56d5fd05fa. 

SeanRobb:
This logic broke the ticker ability for Bittrex.


SeanRobb:
This logic reversed the base and counter pairs for the exchange.  This caused inaccurate ticker information.

PR #1601, opened 2017-07-13
[itbit, cexio, therock, bluetrade, bitfinex, gdax, quadriga] multiple improvements - mostly support for more parameters in getTradeHistory

npomfret:


npomfret:
Not sure what's going on here.  The tests are passing for me.

timmolter:
It looks like the dates in the test are off by one hour. The test is not local-timezone friendly I guess.

npomfret:
Ah! Ok, thanks.  (Maybe thats why the original implementation left them out!)

timmolter:
there's a way to properly do it, but I forgot. I'm 99% sure some other tests do this properly.

timmolter:
I think you have to set the timezone on your Formatter. See other examples. (I'm recalling from memory, could be wrong)

npomfret:
How often is the snap-shot built?

timmolter:
On every commit or accepted pull request. 

I set up Travis CI to do it.

npomfret:
Thanks.

xchange-quadrigacx/src/main/java/org/knowm/xchange/quadrigacx/service/QuadrigaCxTradeServiceRaw.java
```diff
@@ -64,10 +63,16 @@ public boolean cancelQuadrigaCxOrder(String orderId) throws IOException {
         orderId);
   }
 
-  public QuadrigaCxUserTransaction[] getQuadrigaCxUserTransactions(CurrencyPair currencyPair, Long numberOfTransactions) throws IOException {
-
-    return quadrigacxAuthenticated.getUserTransactions(QuadrigaCxUtils.currencyPairToBook(currencyPair),
-        exchange.getExchangeSpecification().getApiKey(), signatureCreator, exchange.getNonceFactory(), numberOfTransactions);
+  public QuadrigaCxUserTransaction[] getQuadrigaCxUserTransactions(CurrencyPair currencyPair, Long numberOfTransactions, long offset, String sort) throws IOException {
```
andre77: What is the purpose of having two identical methods here ???
the only difference, i see, is the "offset" perameter being long/Long

npomfret: that's not intentional, thanks for pointing it out

andre77: create a PR for that already
https://github.com/timmolter/XChange/pull/1604

Issue #1607, opened 2017-07-17
Ensure bitfinex submits correct side when exiting positions


douggie:
Xchange support EXIT_ASK and EXIT_BID order types for exiting positions, when margin trading bitfinex adapter only support BID and ASK, so it might make sense to add the types of EXIT_BID and EXIT_ASK.
update
  String type = limitOrder.getType().equals(OrderType.BID) ? "buy" : "sell";
  to 
 String type = (limitOrder.getType().equals(Order.OrderType.BID) || limitOrder.getType().equals(Order.OrderType.EXIT_ASK)) ? "buy" : "sell";

PR #1613, opened 2017-07-17
#1607

douggie:


douggie:
Ensure bitfinex submits correct side when exiting positions

Xchange support EXIT_ASK and EXIT_BID order types for exiting positions, when margin trading bitfinex adapter only support BID and ASK, so it might make sense to add the types of EXIT_BID and EXIT_ASK.
update
String type = limitOrder.getType().equals(OrderType.BID) ? "buy" : "sell";
to
String type = (limitOrder.getType().equals(Order.OrderType.BID) || limitOrder.getType().equals(Order.OrderType.EXIT_ASK)) ? "buy" : "sell";

PR #1615, opened 2017-07-18
luno - initial implementation

andre77:


timmolter:
don't forget to add it to parent `pom.xml` as a module.

PR #1625, opened 2017-07-20
[therocktrading] added a CanelOrderParams interface that can be by therock to pass through the required currency parameter

npomfret:


npomfret:
Hey @timmolter thanks again.

I think we need a similar change for `withdrawFunds` because some types of wallet (XRP for example) are more than just a single string. So would it be agreeable if I added something like

    // the existing one
    public String withdrawFunds(Currency currency, BigDecimal amount, String address) {...

    // the new one
    public String withdrawFunds(WithdrawFundsParams ) {...
    

Again, I think for ease of use, perhaps keep the old method and delegate everywhere.

timmolter:
Yeah, go for it!

PR #1637, opened 2017-07-25
ripple withdrawls - introduced concept of WithdrawFundsParams to support more complex addresses

npomfret:
I've not yet implemented an XRP withdrawal, but the API at least supports it now.

[GDAX] previous 'moveFunds' functionality now in its own method
[GDAX] fixed withdraw funds


npomfret:
@timmolter this fixes https://github.com/timmolter/XChange/issues/1636

npomfret:
Have moved the 'moveFunds' functionality out of the withDraw method.  It's still there in case people need to use it (I assume it was written for a reason)

npomfret:
@timmolter thanks for merging.  it seems to have broken the build though.  i can't tell from the build output whats wrong

timmolter:
@npomfret Hey, thank you! It looks like a failure to deploy the snapshots, no fault of yours. I've restarted it.

PR #1643, opened 2017-07-28
[hitbtc, bleutrade, poloniex]

npomfret:


npomfret:
@timmolter the Bluetrade change is a little controversial.  Basically the exchange will only supply 1 api key.  So allowing people to create 2 nonces (by creating 2 instances of the exchange) is invalid.  

Thoughts?

I've requested they change their policy, because only having one api-key is too limiting.

PR #1676, opened 2017-08-09
Update HitbtcTradeServiceRaw.java

cymp:
lots sizes are now dynamically retrieved from Hitbtc at trade service creation


npomfret:
Are the hard coded lots now unused?

cymp:
Wel, they should be unused now. I left them just in case the dynamic fetching of the data fails, in this case the hard coded values are here as a backup.

npomfret:
I think they should be removed. I guarantee it will cause confusion in the future. Either your change works or it doesn't (I'm sure it does!)

cymp:
Ok I can do that. Anyways the future of this part of the library is the new API, but no one seems to be happy to take my bounty and implement it :)

npomfret:
I might also be able to take a look at implementing the new api - will see how much work it is first.

PR #1679, opened 2017-08-09
[yobit, dsx] 

npomfret:
[yobit] first stab at getting trade and account service implemented.  work in progress
[dsx] made `cancel` behave better


timmolter:
Can you check the failure out?

npomfret:
Ooops, forgot to commit a file.

npomfret:
@timmolter can you try it again?

PR #1689, opened 2017-08-13
[btcmarkets, quadriga, bitstamp, coinflooor] 

npomfret:
* [btcmarkets] fix currency for trade fees
* [quadriga] make getOpenOrders honour params
* [quadriga] fix wallets so that currencies are correctly using the constant (uppercase) values 
* [bitstamp] added support for ripple withdrawal
* [bitstamp] implemented LTC withdrawals 
* [bitstamp] implemented ETH withdrawals 
* [bitstamp] fixed modelling of error responses for withdrawals
* [coinfloor] added support for new markets
* Added IOTA currency
* And added some metadata for various exchanges, although its causing some upset!


Achterhoeker:
@npomfret I do not agree with the json changes for some exchanges.
For Bitstamp BTC/USD you are giving a minimal amount 5 BTC ? That is a minimal order amount off +/- 20.000,00 USD. That seems incorrect to me. :-)

npomfret:
@Achterhoeker thanks for the feedback.  Do you know how bitstamp calculate their min order size?

Achterhoeker:
@npomfret The docs are correct. But we need the minimal order size based on the base currency. So BTC/USD must be minimal 5 usd. With the current btc value this is +/- 0.00125. Is is not possible to give a minimal order value based on the counter currency. So this minimum is a always changing value. 
For crypto/crypto it is also a minimal value off 5 euro. (although, that is a assumption from my side that always worked)

Achterhoeker:
@npomfret Minimal order sizes on bitfinex are also changed ? they are documented on: https://api.bitfinex.com/v1/symbols_details

npomfret:
Right.  So we have to enter a value which can be used to **calculate** the min order size, no?  

The only value the exchange gives us if 5 (USD/GBP/EUR).  So I guess we have to enter 5 for the min order size everywhere and it's up to the developer to calculate how that translates into the order size for that currency pair.

It would be a nice addition to this API to have a method that uses the static data to return the min order size. The calcs vary from exchange to exchange. Sometimes it's easy, it's just a static value in the base ccy.  But often it's calc based on the static value and the trade direction, sometimes its expressed in the base and sometimes the counter ccy.

Bitstamp seems to make this calc particularly complicated.  It seems that if the currency pair contains a fiat currency then the min order size is just 5 of that currency (which is always the counter).  However, if it's crypto/crypto we need to take one of those currencies and use another exchange rate to calculate that trade value in EUR.

npomfret:
@Achterhoeker thanks for the bitfinex url.  I'll update the code so these are taken from the exchange rather than the static data.

Achterhoeker:
@npomfret I do not agree with that. All exchanges are giving the minimal order amount based for the base currency. If it is not a fixed size (like bitstamp) it is a calculation made once and put in the json file. If it changes to much somebody will change it if they need to. 
With your approach it is not clear what to do with the minimal order amount so it is unusable. (every xchange can differ. I think that the usability off the xchange project is to give a common api)

Achterhoeker:
@npomfret updating bitfinex with the remote data would be great, then the json file will be obsolete. (for bitfinex)
Another question, why changed the bleutrade json ? ltc/btc is already dynamicly loaded from the exchange. So adding to the json file give confusion ?

npomfret:
@Achterhoeker  But you can't enter a min amount for bitstamp in a crypto/crypto market and achieve an accurate min order size.  It changes as the exchange rates change - if the EUR/USD exchange rate changes then the min order size for LTB/BTC will change also.  

Is your suggestion to just calculate the current value and hope the exchange rate doesn't change too much?  And when it does, just update the value?

My last point about adding a `minOrderSize` method (that returns an amount AND currency) would take away any confusion about how to calculate a min order size.  The calculation differs from exchange to exchange, encapsulating it in a common method call on the API would remove that confusion.  

Having a single numeric value in the static data doesn't cover the complexity.

Achterhoeker:
I Agree that the current approach isn't perfect. But changing it requires an api change. Just setting the minimal order amount for example btc/eur to 5 isn't a solution for the current approach. For example my trading bot is following the current approach were the base currency is the minimal order amount. If i pull your pull request my trading bot will never trade again on bitfinex.
So what i'm asking is that please, don't change the current approach without discussion. (and indeed, if an exchange rate changes, the minimal amount changes. That is already the case more many years i'm using this project for some exchanges. Currently it is not possible to indicate for what currency the minimal order amount is mend. So a trading bot isn't able to use it if is set differently )
An other approch for this would be great, make an api suggestion and start the discussion. :+1:

npomfret:
> But changing it requires an api change

The suggestion is to add a new method, so would be backward compatible.  The static data is sufficient and wouldn't change if we added this new method.  it would combine the static data with the order direction and size to produce the true min order size and currency.

> Just setting the minimal order amount for example btc/eur to 5 isn't a solution for the current approach

I'm afraid it is exactly the correct solution for bitstamp according to the exchange documentation (see above).

> For example my trading bot is following the current approach were the base currency is the minimal order amount

I'm afraid you bot has a bug.  We can't avoid correcting this project because of that.  Why not just fix you bug? The other option is just create your own version of the static data file and put it in your main `resources` directory.  The JVM will pick up your version over the one that comes with the project (assuming your class path is set up correctly).  That way you can specify your own min order sizes that play better with your bot's assumption that all min order sizes are expressed in terms of the base ccy.

Achterhoeker:
@timmolter Can you give your opinion about this issue ? If the statements from npomfret is true, (minimal order size reported in the xchange market assetpairs information is exchange dependant. So no common code any-more usable because it needs to know how to handle it) I will make my own json files and start using those. (and not any longer update the json files from this project)

lapoor:
@npomfret this change you are proposing also gives problems with my trading bot as Im also calculating the min order size like @Achterhoeker is doing this worked for years now for me so i see no point in you breaking it now by changing the json.

timmolter:
I'm at a conference right now, but I'll take a close look at this ASAP.

npomfret:
@timmolter do you have any thoughts on this metadata issue? 

It seems a few users have assumed that the min order size in the metadata is always an absolute value expressed in the base currency.  This assumption isn't correct for a number of exchanges, but is for a few.  In truth I think an ideal solution would be to add an API method that calculates the min order size based on the metadata, the currency pair, the trade direction and the order price... but that's a separate discussion I guess.

One of the metadata changes in this PR will adversely affect these users because of that assumption about the min order size (their bots will stop trading on bitstamp).

I've suggested 2 solutions for them:

1) don't could not take the snapshot update until they fix their bug
2) they could just provide their own copy of the metadata files, as long as their classpath is set up correctly they will be unaffected by this change

Of course the other option is to roll back the bitstamp metadata change, but I do think it's _more correct_ than what was there before.  There are a lot of other fixes here and it seems a shame to delay their addition for a simple disagreement about some metadata.  Happy to do whatever.

timmolter:
@npomfret I appreciate your and everyone's efforts and patience on this issue as well as everyone's other contributions in this pull request, past ones, and discussions. I looked in the particular minimum fee issue, and I can see the point of views from all sides after looking at the existing code and familiarizing myself again with this, as I don't actually use this feature of XChange regularly.

First off, let's make sure we all agree what "base" currency means. If we look at the `CurrencyPair` constructor:

```java
 /**
   * <p>
   * Full constructor
   * </p>
   * In general the CurrencyPair.base is what you're wanting to buy/sell. The CurrencyPair.counter is what currency you want to use to pay/receive for
   * your purchase/sale.
   *
   * @param base The base currency is what you're wanting to buy/sell
   * @param counter The counter currency is what currency you want to use to pay/receive for your purchase/sale.
   */
  public CurrencyPair(Currency base, Currency counter) {

    this.base = base;
    this.counter = counter;
  }
```

, we see that it's the first currency, the thing we're wanting to buy/sell. 

Given this fact, and looking through existing metadata json files, I found that the `bitstamp.json` file was originally correct in the sense that the number is given in terms of BTC (the base) in the case of `BTC/USD` for example. If we put 5 there like you're saying then it would mean a minimum order of 5 BTC. The current value of 0.01, is, as you pointed out wrong, and at today's prices is around $40.

The original intent in this case is to override `remoteInit()` like shown in `HitbtcExchange`, load the json file, call remote methods, and override values. There one could call the ticker data and calculate the min order sizes in the base currencies, given a $5 minimum amount.

However, maybe there is a better way, and we should discuss it. Let's move it over to <https://github.com/timmolter/XChange/issues/1544>. For now I will merge this PR to pull in all your updates and manually revert those few json files and to keep it from breaking existing bots, until we figure out a better solution to the metadata.

I hope this is a good compromise and no one gets upset. I feel everyone just wants the cleanest and simplest least-confusing solution to all this, and we can make that happen for sure.

npomfret:
I've contacted bitstamp to get clarity on this issue and they've confirmed that you need to translate any order into either a USD or EUR amount, and the value of that order needs to be at least 5.  It's crazy complicated.

Now we have this information, I think every min order size in the bitstamp .json file should be set to 5.  I might add a method (for bitstamp only for now) that calculates the correct min order size by making an api call first which get's the appropriate exchange rate.  I hope that would be useful and maybe something we can add to the api at a later date.

In the mean time, to those users adversely affected by this change;  you can put a copy of the `bitstamp.json` file in your `main/resources` directory.  It should appear in your classpath before the one in the jar file, and so you can just put you own values in there... or just don't take the snapshot until your code is fixed.

I do appreciate you guys looking at my PR's and pointing out the impact of my changes.

timmolter:
FYI, I updated bitfinex accoring to: <https://support.bitfinex.com/hc/en-us/articles/115003283709-What-is-the-minimum-trade-order-size->

timmolter:
@npomfret The original intent in this case is to override remoteInit() like shown in `HitbtcExchange`,  so I would suggest implementing it like that. I'm sure everybody would be happy to see that additional feature. Then in the logic, calculate each minimum amount by a hardcoded 5 and update the minimum values.

npomfret:
I'm not sure.  As exchange rates are always changing, in this case maybe it has to be done as a method call so you get an accurate calc.  My bot for example can run for days, if we took the exchange rate when it started it could be way out soon enough and the min order size might be too small.

timmolter:
Agreed. I'm just saying that the original idea was to do it that way. But I think a better solution for the future would be to be a change in the metadata like you suggested where the currency is set as well for the minimum amount and an additional method is defined. But still since prices are constantly changing, this conversion rate would need to be called all the time before each order? I'm not so sure actually.

timmolter:
For your long-running bot you can always just periodically call `remoteInit()` (if also implemented like in `HitbtcExchange`). Actually that's probably the best way to handle this issue.

PR #1690, opened 2017-08-14
Issue 1633 update hitbtc

EricAnderson1000:
*** WORK IN PROGRESS ***


xchange-core/src/main/java/org/knowm/xchange/ExchangeFactory.java
```diff
@@ -150,7 +150,7 @@ public Exchange createExchange(String exchangeClassName) {
 
   public Exchange createExchange(ExchangeSpecification exchangeSpecification) {
 
-    Assert.notNull(exchangeSpecification, "exchangeSpecfication cannot be null");
+    Assert.notNull(exchangeSpecification, "exchangeSpecification cannot be null");
```
EricAnderson1000: Spelling fix

EricAnderson1000:
@timmolter Since this is my first Pull Request. Before I got too far I thought I would ask for some feedback. 

Are there any code styles I should be following etc?

Do we want this to be backwards compatible? If so what approach should I take with it? Ie there are some new calls that are not available in the new API (getServer time).

timmolter:
Please read <https://github.com/timmolter/XChange/wiki/New-Implementation-Best-Practices> and see step 10 about code formatting. 

I'm not toooo concerned about backwards compatibility, but try to minimize it when possible and use deprecated tags as well.

In the Raw layers you can add whatever you want. In the generic layer, additions are generally discussed beforehand.

Thanks!

xchange-hitbtc/pom.xml
```diff
@@ -30,6 +30,13 @@
             <version>4.2.2-SNAPSHOT</version>
         </dependency>
 
+        <!--For String Joining, etc. -->
+        <dependency>
```
EricAnderson1000: Added for String manipulation. Is this OK?

xchange-yobit/src/main/java/org/knowm/xchange/yobit/YoBitAdapters.java
```diff
@@ -131,7 +132,7 @@ public static String adaptCcyPairsToUrlFormat(Iterable<CurrencyPair> currencyPai
       pairs.add(adaptCcyPairToUrlFormat(currencyPair));
     }
 
-    return String.join("-", pairs);
+    return StringUtils.join(pairs, "-");
```
EricAnderson1000: String.join is only available in 1.8.
[Javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-)

 Since this is 1.7 project...

xchange-yobit/pom.xml
```diff
@@ -23,6 +23,14 @@
             <artifactId>xchange-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <!-- For Joining Strings -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.6</version>
```
timmolter: remove the version here as it's defined in the parent pom

xchange-hitbtc/pom.xml
timmolter: Yes, it's already defined in the parent pom.xml

timmolter:
Can you please merge with the most up to date changes on `develop`?

PR #1697, opened 2017-08-18
add null check for minimum amount (for poloniex)

gpwclark:
Poloniex's currency pair meta data doesn't seem to ever have a minimum amount. This means that the verifyOrder method typically fails on Poloniex. I'm happy to make the pull request better, but there doesn't seem to be any other meta data about the currency pair that one could check to stay in line with the spirit of the verifyOrder method. Poloniex does tend to have the price scale int value, making a call to metaData.getPriceScale() valid but that seems like it would return an int value of price scale for the quote currency. I'm not sure we can infer that the tradeable amount's scale must be less than (or equal to?) the quote currency's price scale. If that is a good assumption I'm happy to put it in and throw an IllegalArgumentException.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/service/BaseExchangeService.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/service/BaseExchangeService.java
@@ -57,11 +57,12 @@ public abstract class BaseExchangeService {
 
     BigDecimal amount = tradableAmount.stripTrailingZeros();
     BigDecimal minimumAmount = metaData.getMinimumAmount();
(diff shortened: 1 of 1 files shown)


gpwclark:
After talking to a subject matter expert, it seems like the best way to check the possible scale the tradeable amount can be is from the getScale method of the CurrencyMetaData object found off the given base currency. Happy to check against that in a revised commit.

timmolter:
OK, so if no minimum amount is given, it just let's it through. Seems like a good compromise. I googled for Poloniex minimum amount and couldn't find anything.

PR #1699, opened 2017-08-18
[bitstamp] support eth and ltc withdrawals

andre77:


npomfret:
Looks like I have the same change in https://github.com/timmolter/XChange/pull/1689 + ripples

timmolter:
Closing because you said it was handled in #1689, which I just merged.

PR #1711, opened 2017-08-22
Updated kraken.json

Achterhoeker:
Kraken removed some pairs: 

In order to help mitigate the strain on our platform caused by recent exponential
growth, we are delisting some of our illiquid trading pairs and temporarily
suspending all advanced order types. *These changes will go into effect on Tuesday,
August 22 around 6 am UTC (Monday August 21 around 11 pm Pacific).*

# The following pairs will be delisted with all outstanding orders being canceled

* XBT/GBP
* ETH/GBP
* XRP/CAD
* XRP/JPY
* EOS/EUR
* EOS/USD
* XLM/EUR
* XLM/USD
* GNO/EUR
* GNO/USD
* REP/USD

# The ability to create new advanced orders will also be temporarily suspended (so
you will only be able to create basic limit and market orders)

Open orders will remain open until cancelled or executed, but you will no longer be
able to create new orders in any of the following advanced types:

* Stop Loss
* Take Profit
* Stop Loss, Take Profit
* Stop Loss, Take Profit Limit
* Stop Loss Limit
* Take Profit Limit
* Trailing Stop
* Trailing Stop Limit
* Stop, Limit

This suspension is temporary only. Advanced orders will return in a month or two as
soon as we are better able to process them. We know that many of our clients like
the advanced orders, but they are used much less frequently than basic limit and
market orders while taking more system resources to process.

*The above changes will go into effect on Tuesday, August 22 around 6 am UTC (Monday
August 21 around 11 pm Pacific).*

Thank you for choosing Kraken,

The Kraken Team

### DIFF
--- a/xchange-kraken/src/main/resources/kraken.json
+++ b/xchange-kraken/src/main/resources/kraken.json
@@ -2,49 +2,45 @@
   "currency_pairs": {
     "BTC/EUR": {
       "price_scale": 3,
-      "min_amount": 0.002000000
+      "min_amount": 0.002000000      
     },
     "BTC/USD": {
       "price_scale": 3,
-      "min_amount": 0.002000000
+      "min_amount": 0.002000000      
     },
     "BTC/CAD": {
       "price_scale": 3,
-      "min_amount": 0.002000000
-    },
-    "BTC/GBP": {
-      "price_scale": 3,
-      "min_amount": 0.002000000
+      "min_amount": 0.002000000      
     },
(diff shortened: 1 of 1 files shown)


timmolter:
It looks like there is a json parsing error.

Achterhoeker:
@timmolter There was indeed a , to much at the end. Sorry for that. I encountered a diff when i updated the json file. (was updated between my pull and push)  I've merged it manually and missed that ',' .

PR #1715, opened 2017-08-22
Issue 1633 -- Update hitbtc with backwards compatible

EricAnderson1000:
Please provide feedback on this approach.

@timmolter Reviewer
@npomfret Reviewer

This is a backwards compatible change that leaves all existing calls in place. New services will as best as possible keep existing interfaces.. but some method signatures will indeed change with the differences in the APIs.
What do you think about this approach? 

Adding tests to new services in package 'v2'

I could also commit smaller PRs for review with this approach.


xchange-hitbtc/src/main/java/org/knowm/xchange/hitbtc/v2/HitbtcExchange.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private final SynchronizedValueFactory<Long> nonceFactory = new CurrentTimeNonceFactory();
+
+  private HitbtcMetaData hitbtcMetaData;
+
+  @Override
+  protected void initServices() {
+
+    marketDataService = new HitbtcMarketDataService();
+    tradeService = new HitbtcTradeService(this);
```
EricAnderson1000: Borrowing old services here until new ones are implemented.

xchange-hitbtc/src/main/java/org/knowm/xchange/hitbtc/v2/HitbtcExchange.java
```diff
@@ -0,0 +1,70 @@
...
+  @Override
+  public SynchronizedValueFactory<Long> getNonceFactory() {
+
+    return nonceFactory;
+  }
+
+  @Override
+  public void remoteInit() throws IOException {
+
+    HitbtcRestClient.INSTANCE.init(this);
```
EricAnderson1000: Moving the actual rest proxy to a singleton. In the older version we are creating this object 3 times for each service.

npomfret:
I'll should have time to take a look tomorrow.

EricAnderson1000:
@npomfret @timmolter Do you have time to look over this PR today?

timmolter:
Can you not use Guava? We've managed to keep it out of XChnage for many years now. What features do you need?

EricAnderson1000:
>Can you not use Guava? We've managed to keep it out of XChnage for many years now. What features do you need?

It's scope is test, which means it shouldn't be part of the jar artifact right? I am using it for it's functional map transformation in a test. If Java 8 were used, it wouldn't be needed for sure. If you still prefer it not be used in tests, I can remove it. Just let me know.

timmolter:
I'd be open for updating to Java 8. Let me open up a specific issue to get people's opinion. 

Is this ready to be merged now? @npomfret comments?

EricAnderson1000:
test this please

EricAnderson1000:
If it's OK with everyone, lets merge this, and I will keep adding features to the new Hitbtc v2 package.

npomfret:
Ok with me.  But can we be very quick to fix any regressions?

EricAnderson1000:
>Ok with me. But can we be very quick to fix any regressions?

Of course. But I would be surprised if there were any.  Since existing code is left in place and only deprecated.

PR #1724, opened 2017-08-26
Add theOtherType method to OrderType

Yuki-Inoue:
Sometimes, I want to get the `ASK` for `BID` OrderType and vise versa.  Rather than writing switch statements or ifs on many places, I'd prefer the OrderType enum can calculate the other type of itself.

So, wrote such method.
... Any opinion?


timmolter:
I don't understand what this is for. Seems like you can have this method in your own client code.

Yuki-Inoue:
Just to make it easy, I want it on the library.  But guess you're right.  I can have this on my client.

Issue #1398, opened 2017-03-12
Quoine: Authentication failed


wuqianming:
Caught: org.knowm.xchange.exceptions.ExchangeException: API Authentication failed
org.knowm.xchange.exceptions.ExchangeException: API Authentication failed
	at org.knowm.xchange.quoine.service.QuoineAccountServiceRaw.getQuoineFiatAccountInfo(QuoineAccountServiceRaw.java:32)
	at org.knowm.xchange.quoine.service.QuoineAccountService.getAccountInfo(QuoineAccountService.java:45)
	at org.knowm.xchange.service.account.AccountService$getAccountInfo.call(Unknown Source)

jheusser:
@wuqianming currently you need to set KEY_TOKEN_ID and KEY_USER_SECRET in the exchange specific parameters. The two values you get from the API key you generate on Quoine.

```
exchange.getExchangeSpecification().getExchangeSpecificParameters().put(QuoineExchange.KEY_TOKEN_ID), "your token id");
exchange.getExchangeSpecification().getExchangeSpecificParameters().put(QuoineExchange.KEY_USER_SECRET, "your secret");
```
I will make this more user friendly.

wuqianming:
@jheusser  I've tried but got this,seems like it's a readonly property:

Caught: groovy.lang.ReadOnlyPropertyException: Cannot set readonly property: KEY_TOKEN_ID for class: org.knowm.xchange.quoine.QuoineExchange
groovy.lang.ReadOnlyPropertyException: Cannot set readonly property: KEY_TOKEN_ID for class: org.knowm.xchange.quoine.QuoineExchange

wuqianming:
Thank you so much man, I've figured it out.
BTW may I have your e-mail address ? @jheusser

npomfret:
I'm getting a 401 too.  Is there a trick to this? I've tried:

```
specification.setExchangeSpecificParametersItem(QuoineExchange.KEY_TOKEN_ID, keys.apiKey);
specification.setExchangeSpecificParametersItem(QuoineExchange.KEY_USER_SECRET, keys.secret);
```

jheusser:
@npomfret This is correct. How do you instantiate the `Exchange` instance?

npomfret:
`ExchangeFactory.INSTANCE.createExchange(specification)`

It's a bit confusing known which bit of the text provided by the exchange is the api key and which bit is the secret.

npomfret:
[...]

PR #1726, opened 2017-08-27
[quoine] use getApiKey and getSecretKey

Yuki-Inoue:
Currently, xchange-quoine uses it's own specification parameter for setting api(or, token, in the terms of quoine exchange) and secret key.  This does not follow convention, so I thought it would be better to use the ordinal interface of `ExchangeSpecification#getApiKey()` and `ExchangeSpecification#getSecretKey()`.

(Can revent new people stumbling on this peculiarity, e.g. #1398 , IMO.. )


npomfret:
I agree, no idea why it was implemented in the way it was.

PR #1743, opened 2017-08-31
Update btctrade.json

glebreutov:
currency list fixed


glebreutov:
and LakeBTC migrated to v2. I've tested public services only, but v1 not supported in any case

PR #1757, opened 2017-09-09
[kraken] Adapt limit order

MartianTech:
Update Kraken's adaptLimitOrder to use original amount and filled amount. Notice there is a proposed change of LimitOrder to include 
```
OriginalAmount
FilledAmount
RemainingAmount
```
While we are still waiting for the upper level change, I think this is good for now. 

This PR also add `adaptOrderStatus`.


MartianTech:
Some build error outside of my change, hitbtc?

timmolter:
I've renamed some test classes to not run as normal tests. Please pull changes and update your fork. It should trigger a re-run.

MartianTech:
Does this work now?

Issue #1767, opened 2017-09-18
createExchange method with generic


rostuha1:
I think it will be better to use generic instead of String exchangeClassName.

For example we can replace this method ->
public Exchange createExchangeWithoutSpecification(String exchangeClassName)

with it ->

    public Exchange createExchangeWithoutSpecification(Class<? extends Exchange> exchange) {
        try {
            return exchange.newInstance();
        } catch (InstantiationException e) {
            throw new ExchangeException("Problem creating Exchange (instantiation)", e);
        } catch (IllegalAccessException e) {
            throw new ExchangeException("Problem creating Exchange (illegal access)", e);
        }
    }

And call: Exchange exchange = createExchangeWithoutSpecification(BittrexExchange.class); without getName() and possible exceptions with it

timmolter:
I considered this, but I'm not so sure it's a problem needing fixing. The most common way to create an Exchange is to use the method with an ExchangeConfiguration, which uses a String for the exchange class name. I think this is advantageous for serialization or creating instances in the most generic way possible, from a config file for example.

I did add overloaded methods with this argument type though!

Issue #1714, opened 2017-08-22
open orders: tradableAmount


npomfret:
I know this question was discussed recently but I can't find it, sorry.

What is the `tradableAmount` field _supposed_ to represent in a open order?  I guess it could be any of:

 * the original order size (before any fills happened)
 * the current order size (the original minus the total quantity filled already), aka _amountRemaining_
 * the amount filled already, aka _amountFilled_

cymp:
I am 99% sure it is not the _amountFilled_ already, from what I observed in my software.
My understanding is that it corresponds to the _amountRemaining_.

MartianTech:
This has been creating inconsistency across exchanges. I propose we change them to 
```
private final BigDecimal originalAmount;
private final BigDecimal remainingAmount;
private final BigDecimal executedAmount (or filledAmount);
```

timmolter:
@MartianTech I think that's a great idea.

npomfret:
+1 for this.  Big change though.

I'd prefer to name the 3rd one `filledAmount`, that's the terminology I'm familiar with from working in finance.  I think _executedAmount_ could be mistaken for _originalAmount_ by some.

timmolter:
I like `filledAmount` better as well. We can keep the existing getter and setter (but delete the field) and just set/return whatever field we think is the standard one, in addition to the new three.

Achterhoeker:
This is what i've already started with. See https://github.com/timmolter/XChange/issues/1693
I'll add an getter for filledAmount (doesn't need a field, it can be calculated from originalAmount and remaining amount)
I'll also rename the current amount field to original amount and maintain the getter/setter for it.
But the problem is time to do this. Next Wednesday is the first day i have some spare time for this.

timmolter:
deprecated `getTradeableAmount` also now removed.

Issue #1693, opened 2017-08-15
[Kraken] adaptLimitOrder tradableAmount


MartianTech:
```  
public static LimitOrder adaptLimitOrder(KrakenOrder krakenOrder, String id) {

    KrakenOrderDescription orderDescription = krakenOrder.getOrderDescription();
    OrderType type = adaptOrderType(orderDescription.getType());
    BigDecimal tradableAmount = krakenOrder.getVolume().subtract(krakenOrder.getVolumeExecuted());
    CurrencyPair pair = adaptCurrencyPairOpenOrder(orderDescription.getAssetPair());
    Date timestamp = new Date((long) (krakenOrder.getOpenTimestamp() * 1000L));

    return new LimitOrder(type, tradableAmount, pair, id, timestamp,
        orderDescription.getPrice());
  }
```

Most exchanges define `tradableAmount` in a limit order to be the entire order size client placed. However this Kraken's adapter define it to be the order size minus the partially filled size. 

It's not a bug accidentally introduced because the author intentionally implemented this way. However this is not consistent with most other exchanges' implementation. Should we change this?

npomfret:
How can you see the amount remaining typically?

MartianTech:
Do you mean the unfilled amount in an order? Then it's the order size minus partially filled size just like the above implementation. However the problem here is that exchanges are defining `tradableAmount` differently.

Achterhoeker:
@MartianTech Which exchanges do not return the open amount for the order ? 

As far as I known the exchanges Bitfinex, CexIO, Bitstamp, Kraken, Bittrex, Polo and Bleutrade return the amount still open for open open orders. (and not the original amount).
This is also discussed a long time ago, then the decision was made that the amount still open needs te be returned.

MartianTech:
@Achterhoeker Sorry if I incorrectly used the term "most exchanges" but to name a few, Gdax, Okcoin and Bitfinex, which you stated the opposite - 
```
  public static OpenOrders adaptOrders(BitfinexOrderStatusResponse[] activeOrders) {

    List<LimitOrder> limitOrders = new ArrayList<>(activeOrders.length);

    for (BitfinexOrderStatusResponse order : activeOrders) {

      OrderType orderType = order.getSide().equalsIgnoreCase("buy") ? OrderType.BID : OrderType.ASK;
      OrderStatus status = adaptOrderStatus(order);
      CurrencyPair currencyPair = adaptCurrencyPair(order.getSymbol());
[...]

PR #1768, opened 2017-09-18
issue #1767: Bitfinex open orders adaptation

pavlo-pastushok:
I propose set remaining amount instead of original to LimitOrder on adapting Bitfinex open orders (reason: amount in open orders never change until full order operation)


MartianTech:
See discussion here #1714, #1693 and PR #1761

timmolter:
Are there any objections to this? 

@MartianTech

MartianTech:
I'm not using Bitfinex but I'd suggest that we use the new limit order field to include original, remaining and filled amount.

PR #1770, opened 2017-09-19
[Core] Fix getFilledAmount [Gemini] [Gdax] Adapt open order

MartianTech:
- Fix LimitOrder getFilledAmount. 

- Update Gemini and Gdax's open order adapters to include original, remaining and filled amount. 

- Use `AtomicLongCurrentTimeIncrementalNonceFactory` instead of `AtomicLongIncrementalTime2013NonceFactory` because according to Gemini, "e recommend using a timestamp at millisecond or higher precision"
https://docs.gemini.com/rest-api/#private-api-invocation
`AtomicLongIncrementalTime2013NonceFactory` is rounding down the precision.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/trade/LimitOrder.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/trade/LimitOrder.java
@@ -109,7 +109,7 @@ public class LimitOrder extends Order implements Comparable<LimitOrder> {
 
 	if (remainingAmount != null)
 	{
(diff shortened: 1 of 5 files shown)


timmolter:
Attention: This changes `LimitOrder` behavior. Any objections to this??

MartianTech:
The part changing LimitOrder is a fix. It was 
`return getOriginalAmount().min(remainingAmount);` and using `min` was clearly not correct.  

It should be `return getOriginalAmount().subtract(remainingAmount);`

PR #1785, opened 2017-10-01
Added additional coins - MLN, ETC, ZEC, EOS, DASH, GNO, ICN

jcuypers:
Additional coins added to support Kraken Currency pairs


jcuypers:
need to combine aal changes

PR #1786, opened 2017-10-01
Additional trading pairs for kraken based on MLN/ETC/ZEC/EOS/DASH/GNO…

jcuypers:
…/ICN


jcuypers:
need to combine with previous pull

PR #1795, opened 2017-10-04
coinmarketcap read-only data / base CurrencyAmount class

allenday:
added read capability from coinmarketcap.com API.

Also introduced a CurrencyAmount class to xchange-core, I noticed something like this is needed in the xchange-coinbase, and I need it for some other work I'm doing that is not exchange related, so I thought best to pull up to parent project.  Happy to discuss further, thx.


timmolter:
How is the `CurrencyAmount` class used? It doesn't seem necessary.

allenday:
Hi @timmolter,
RE `CurrencyAmount`, take a look at this class in `xchange-coinbase`
https://github.com/timmolter/XChange/blob/develop/xchange-coinbase/src/main/java/org/knowm/xchange/coinbase/dto/marketdata/CoinbasePrice.java
making use of `CoinbaseMoney`.  I need a similar concept in an app I'm building with XChange, but I need it general purpose across all exchanges.

The concept comes up in the context of fees, and importantly also in technical analysis where it's necessary to be able to reason about quantities of objects generically.

timmolter:
I've removed it for now. If there is a general need for it in xchange-core, let's reconsider in a separate issue.

PR #1798, opened 2017-10-07
add httpConnTimeout setting and apply both timeouts in several exchanges

johnzweng:
## Overview:
This pull request adds the support to set HTTP connection timeout and the read timeout separately per exchange (via ExchangeSpecification). As we experience different behavior of different exchanges during times of high traffic, it might be useful to set different values at runtime (per exchange).

## Details:
The first commit adds `setHttpConnTimeout()` and `getHttpConnTimeout()` to `ExchangeSpecification` (similar to methods for httpReadTimeout which were already present.

The second commit adds the application of **both** timeouts (`httpConnTimeout` and `httpReadTimeout`) to several exchange subprojects:

In detail:
- Bitfinex
- Bitstamp
- Bittrex
- C-Cex
- CEX.io
- Coinbase
- GDAX
- HitBTC
- Kraken
- Luno
- Poloniex
- Ripple
- TheRock

I searched for all occurrences of `RestProxyFactory.createProxy(..)` in these projects and made sure, that both timeouts will be applied if they are set in `ExchangeSpecification`.

### DIFF
--- a/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/service/BitfinexBaseService.java
+++ b/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/service/BitfinexBaseService.java
@@ -5,6 +5,7 @@ import org.knowm.xchange.bitfinex.v1.BitfinexAuthenticated;
 import org.knowm.xchange.service.BaseExchangeService;
 import org.knowm.xchange.service.BaseService;
 
+import si.mazi.rescu.ClientConfig;
 import si.mazi.rescu.ParamsDigest;
 import si.mazi.rescu.RestProxyFactory;
 
@@ -24,7 +25,18 @@ public class BitfinexBaseService extends BaseExchangeService implements BaseServ
 
     super(exchange);
 
-    this.bitfinex = RestProxyFactory.createProxy(BitfinexAuthenticated.class, exchange.getExchangeSpecification().getSslUri());
+    // allow HTTP connect- and read-timeout to be set per exchange
+    ClientConfig rescuConfig = new ClientConfig(); // default rescu config
+    int customHttpConnTimeout = exchange.getExchangeSpecification().getHttpConnTimeout();
+    if (customHttpConnTimeout > 0) {
+      rescuConfig.setHttpConnTimeout(customHttpConnTimeout);
+    }
+    int customHttpReadTimeout = exchange.getExchangeSpecification().getHttpReadTimeout();
+    if (customHttpReadTimeout > 0) {
+      rescuConfig.setHttpReadTimeout(customHttpReadTimeout);
+    }
+
+    this.bitfinex = RestProxyFactory.createProxy(BitfinexAuthenticated.class, exchange.getExchangeSpecification().getSslUri(), rescuConfig);
     this.apiKey = exchange.getExchangeSpecification().getApiKey();
     this.signatureCreator = BitfinexHmacPostBodyDigest.createInstance(exchange.getExchangeSpecification().getSecretKey());
     this.payloadCreator = new BitfinexPayloadDigest();
--- a/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampAccountServiceRaw.java
+++ b/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampAccountServiceRaw.java
@@ -20,6 +20,7 @@ import org.knowm.xchange.currency.Currency;
 import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.exceptions.ExchangeException;
 
+import si.mazi.rescu.ClientConfig;
 import si.mazi.rescu.RestProxyFactory;
 import si.mazi.rescu.SynchronizedValueFactory;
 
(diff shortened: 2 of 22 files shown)


johnzweng:
Maybe we should avoid adding the same timeout-setting code in each service class and instead add a getClientConfig() method to the BaseExchangeService superclass, which returns a ClientConfig object with generic settings (like the timeout settings) already set.

Inherited classes could overload the method or create their own ClientConfig if they want or need to.

Of course it would still be the responsibility of the  inherited service classes to actually use this ClientConfig when creating the RestProxy (but wie would avoid repeating setting the timeout in each service class).

timmolter:
Sorry I'm taking so long on this. When I have time, I'll get to it. Thanks!

johnzweng:
No problem, take your time. If I can do something let me know. I will happily provide you with a new pull-request, if I should change something.

timmolter:
OK, I think the idea is good. However, it looks like there is roughly the same copy-pasted code in all the *BaseService classes. Can you move it up a level to `BaseExchangeService` perhaps?

johnzweng:
Hi!

Yes, this was my thinking too. I restructured the change now in the ["add_http_timeouts_base_service" branch in my repo](https://github.com/johnzweng/XChange/commits/add_http_timeouts_base_service):

Please look there into the last 4 commits.

The first 2 commits add the connection timeout to the `ExchangeSpecification` and the `BaseExchangeService`

The [3rd commit](https://github.com/johnzweng/XChange/commit/9b1d6b2fddfb303df25087a8879da46c7295396b) then makes use of these additions and includes the RestCU client config in every extending classes which creates a RestProxy.

So after this change (almost) all exchange implementations will make use of exchange-specific connection- and read-timeouts.

I say almost here, as not all exchanges extend the BaseExchangeService class. In the following 5 classes, which are not extending the `BaseExchangeService` also RestProxy instances are created, but I didn't change anything there. So these 5 classes currently will still not respect the timeout settings:

- [BitcoinChartsFactory:24](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-bitcoincharts/src/main/java/org/knowm/xchange/bitcoincharts/BitcoinChartsFactory.java#L24)
- [CoinMarketCapMarketDataServiceRaw:25](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-coinmarketcap/src/main/java/org/knowm/xchange/coinmarketcap/service/CoinMarketCapMarketDataServiceRaw.java#L25)
- [CryptopiaAccountServiceRaw:29](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-cryptopia/src/main/java/org/knowm/xchange/cryptopia/service/CryptopiaAccountServiceRaw.java#L29)
- [CryptopiaTradeServiceRaw:31](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-cryptopia/src/main/java/org/knowm/xchange/cryptopia/service/CryptopiaTradeServiceRaw.java#L31)
- [BitVcFuturesServiceRaw:16](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-huobi/src/main/java/org/knowm/xchange/huobi/service/BitVcFuturesServiceRaw.java#L16)

Also the `LunoExchange` was a little bit different than the others. It creates its RestProxy in [LunoAPIImpl:45](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-luno/src/main/java/org/knowm/xchange/luno/LunoAPIImpl.java#L45). This class is used by [LunaExchange](https://github.com/johnzweng/XChange/blob/add_http_timeouts_base_service/xchange-luno/src/main/java/org/knowm/xchange/luno/LunoExchange.java#L31) (which unfortunately also doesn't extend `BaseExchangeService` but already included code to add the httpReadTimeout to the ClientConfig). So for now I just kept it as it was and simply added code for setting the connect timeout in the same manner as it was for the read timeout in [this commit](https://github.com/johnzweng/XChange/commit/f09848d72c248975f7d9588886521d67cfee0c4f). 


Tell me what you think or if I should change something. If it's fine for you I will send it as a pull-request.

kyr7:
Maybe we could add http proxy parameters to ClientConfig the same way?

johnzweng:
Hi @kyr7! I've seen your issue #1823 on this topic. Sounds reasonable to me. If all exchanges will use the ClientConfig object already configured in the super class (what my branch tries to do), then it will be easy to add support for proxy host/port in a single place.

But let's wait for @timmolter 's response what he thinks.

timmolter:
I'm looking at this issue today... Thanks for your ideas!

npomfret:
Is it possible to be able to pass a ClientConfig object into the exchange?  It would allow adhoc / per-exchange http configuration.  Might be useful if for example one exchange had particularly poor performance.

timmolter:
It looks like what @johnzweng did was add the individual config params to the `ExchangeSpecification`, which I like. If none are provided it simply falls back on the default.

@johnzweng I will update those 5 or 6 classes to fall in  line with the normal class hierarchy convention. Could you create a PR then with your ClientConfig implementation?

johnzweng:
Sure, just let me know, when you are ready. 👍

timmolter:
OK, preparation is finished! I think if you submitted a new clean PR, it'd be best.

kyr7:
Great. Then I'll wait till this PR is merged and will add proxy settings on top.

johnzweng:
Ok, I am preparing a new pull request. Therefore closing this one.

johnzweng:
Ok, here we go, new pull request over there: #1835 :-)

johnzweng:
@kyr7 My other pull-request has been merged into develop. :-) I think you can start your work now on top of the "develop" branch. 

You just need to set the needed options on the RestCU client config in the `getClientConfig()`method in `BaseExchangeService` and it will be used in all current exchange implementations.

kyr7:
@johnzweng, I'm on it.

PR #1805, opened 2017-10-10
hitbtc v1 api balance total fixed

cymp:


cymp:
PRed too fast, sorry.

PR #1811, opened 2017-10-13
various fixes

npomfret:


timmolter:
@npomfret What's the current situation with BTC-E? I thought they shut down. Is it run somewhere else under a different name now?

cymp:
@timmolter  They are now https://wex.nz/

timmolter:
Just curious - were all user accounts upheld? How was the transition?

cymp:
I'm not sure everyone got their money back : https://www.coindesk.com/btc-e-back-online-refund-efforts-anything-smooth/

cymp:
There is a bitcointalk topic somewhere as well.

Achterhoeker:
@timmolter  Some of the assets are converted in debt-tokens. (like bitfinex did in the past)
So there is a possibility everything will be returned. (The assets confiscated by the fbi are converted in debt tokens) My trading bot is running good on the new exchange. (like it always did on btc-e)

timmolter:
Thanks, everyone.

timmolter:
Should we change the module name, namespace and class names etc. to WEX? I can do it...

npomfret:
I guess we should.

PR #1817, opened 2017-10-16
[hitbtc] funds conversion fix

andre77:


xchange-hitbtc/src/main/java/org/knowm/xchange/hitbtc/v2/internal/HitbtcAdapters.java
```diff
@@ -241,21 +242,70 @@ public static ExchangeMetaData adaptToExchangeMetaData(List<HitbtcSymbol> symbol
...
+    if (transaction.getIndex() != null) {
+        description += ", index: " + transaction.getIndex();
+    }
+    if (transaction.getHash() != null) {
+        description += ", hash: " + transaction.getHash();
+    }
+    if (transaction.getPaymentId() != null) {
+        description += ", paymentId: " + transaction.getPaymentId();
+    }
+
```
npomfret: Can we use `transaction.getHash()` for the externalId in fundingRecord?  I think that's what a few other implementations have done where it's available?

andre77: sure, we can.
there is nothing to be said against it.

PR #1819, opened 2017-10-16
Bitfinex Lending history implemented

dave0688:
I implemented to fetch the lending history for Bitfinex. 
I also tested it (and left a test in the bitfinex-examples) and it works.

The only thing I'm not sure: The Bitfinex docs state that there's a RateLimit of 45reqs/min. Does that have to be considered somewhere? 
And, just out of curiousity: Is the Ratelimit per Api Key or per Server?

Please, since it's my first commit to this project, doublecheck it. If approved, please merge it into develop :)

Kind regards,
David


timmolter:
Thanks! The rate limit doesn't need to be considered anywhere in the XChange API. Probably applied to an API key, but I'm not really sure.

PR #1828, opened 2017-10-19
Support for Binance Exchange

andre77:
Support for the Binance Exchange https://www.binance.com
API description: https://www.binance.com/restapipub.html


andre77:
as i write, the API documentation is being updated by binance, they are adding support for withdrawals/fundrecords
as far as its done, i will implement is in a separate PR

xchange-binance/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+        <url>http://knowm.org/open-source/xchange/</url>
+    </organization>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.knowm.xchange</groupId>
+            <artifactId>xchange-core</artifactId>
+            <version>4.3.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
```
timmolter: This can and should be left out, it gets added to all submodules automatically

xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAuthenticated.java
```diff
@@ -0,0 +1,231 @@
...
+     * @param newClientOrderId optional, a unique id for the order. Automatically generated if not sent.
+     * @param stopPrice optional, used with stop orders
+     * @param icebergQty optional, used with iceberg orders
+     * @param recvWindow        optional
+     * @param timestamp
+     * @return
+     * @throws IOException
+     * @throws BinanceException
+     */
+    BinanceNewOrder newOrder(@FormParam("symbol") String symbol
```
timmolter: Do you have the Eclipse or IntelliJ formatter that makes the params get formatted like this? I like it.

xchange-binance/pom.xml
andre77: do you mean the version tag? or the whole artifact as dependency?
its actually the same as in every other module

xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAuthenticated.java
andre77: i use eclipse, but this is my own preferred formatting

xchange-binance/pom.xml
timmolter: Sorry it marked the wrong line. I mean the junit dependency

PR #1835, opened 2017-10-20
add getClientConfig to BaseExchangeService and make use of it

johnzweng:
## Overview:
This is a rework of pull-request #1798. Please look over there for full discussion history.

## Details:
1) This PR adds a method `getClientConfig()` to the `BaseExchangeService` which returns a RestCU configuration object, which has already the timeout values from the `ExchangeSpecification` set (uses defaults if they are not set).

2) Also in `ExchangeSpecification` a method for setting the `httpConnTimeout` was added (the `httpReadTimeout` was already included before).

3) Over all sub-projects (exchange implementations) in all subclasses of `BaseExchangeService` every call to `RestProxyFactory.createProxy(..)` was modified, to make use of the newly added `getClientConfig()` method (so that the set timeout values will be used).


johnzweng:
👍

PR #1839, opened 2017-10-23
[DSX] Fixed market order placing

Mikelle:


xchange-dsx/src/main/java/org/knowm/xchange/dsx/DSXAdapters.java
```diff
@@ -237,6 +237,6 @@ private static BigDecimal withScale(BigDecimal value, int priceScale) {
   public static LimitOrder createLimitOrder(MarketOrder marketOrder, DSXExchangeInfo dsxExchangeInfo) {
     DSXPairInfo dsxPairInfo = dsxExchangeInfo.getPairs().get(getPair(marketOrder.getCurrencyPair()));
     BigDecimal limitPrice = marketOrder.getType() == OrderType.BID ? dsxPairInfo.getMaxPrice() : dsxPairInfo.getMinPrice();
-    return LimitOrder.Builder.from(marketOrder).limitPrice(limitPrice).build();
+    return LimitOrder.Builder.from(marketOrder).limitPrice(limitPrice).timestamp(new Date()).build();
```
timmolter: `timestamp` should be null if there is none given by the server. This has been the standard convention in XChange since day one.

timmolter:
@Mikelle Could you check out why the checks are failing?

PR #1857, opened 2017-11-01
[binance] deposit/withdrawal support

andre77:


andre77:
please do not merge yet, i need to add a couple of things

andre77:
ok, it should be ok for now
still the api documentation for deposits/withdrawals at binance is very thin... so the current implementation is not very stable. its likely, that the api will be changed soon. in that case, i will update the implementation here

PR #1862, opened 2017-11-03
Adding price to update order for Hitbtc

EricAnderson1000:
@timmolter 

Sorry one more update.

Also a question as well. How do you feel about mocking out API server responses with a framework like WireMock? I could keep these tests running if I did that. But it would mean if the APIs changed, running the test would not identify that change.


timmolter:
No worries!

I'm not a big fan of mocking out the server responses, mainly because the coverage will be extremely low across the entire project. But if you want to go ahead and do it for one or more, go for it! Someone added `powermock` to the pom and used it for a few implementations a while ago, so maybe you could take a look at that and try to stick with what they did.

Issue #1866, opened 2017-11-05
Travis builds not running unit tests


ww3456:
Have the unit tests been disabled on purpose? e.g. 

https://travis-ci.org/timmolter/XChange/builds/297084703#L454
```
$ mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```

KrakenAdaptersTest.testAdaptCurrencyPairs will need fixing if they are going to be turned back on.

timmolter:
Do you know where that's coming from? I'm confused about how that it even there.

timmolter:
Our `.travis.yml` config is this:

```
language: java
sudo: false

before_install: "git clone -b travis `git config --get remote.origin.url` target/travis"
script: "
if [ ${TRAVIS_PULL_REQUEST} = 'false' ];
then 
    mvn clean deploy --batch-mode --settings target/travis/settings.xml; 
else 
    mvn clean verify --batch-mode --settings target/travis/settings.xml;
fi"

jdk:
  - oraclejdk8

# whitelist
branches:
  only:
    - develop

env:
  global:
    - secure: "pDTp3TujySuBn3IV5QzgYSsLdLHRDApm2dvKT5P8aHgLDF2tM59a5fWYuA+MiTrZdpbxsUXzxrl4l5+cQyLGI10/hu4R7uft4k46Da78xNA5REki+YU7rjCA6RXdWUuXyhLNFYPWVNt3mgi7YfsyNwRrtS3FQQZn1BOfDJ0s5Vo="
    - secure: "YwWFnNl4cBEKzRiMjKlfp1VieMkWp4LGENMKzlVs7hy3TWFz4A2pJpRUX76qqWV/6h6vOqY6cogql3NCietzJbspSgwl7MkboPVXYn9HgQkPwrcdQh/h9PzsC63iOW4jyIqHzjVT36MYLYrXTvRrqdMHUjFJmObiNt1KY674d7g="

matrix:
  - LC_ALL=C
  - LC_ALL=pl_PL
  - LC_ALL_zh_CN
```

ww3456:
No, I'm afraid not. It is also skipping the javadoc, as though it is running in a configuration to speed up the build.

timmolter:
OK, looks like Travis CI changed something on their end. I think it's fixed now.

<https://stackoverflow.com/questions/31945809/skip-first-mvn-install-in-travis-ci>

Issue #1430, opened 2017-04-07
High Priority one! BitfinexOrderStatusResponse cannot handle the ID returned by the exchange.


kisapmata:
This is quite a big issue. Since march 24, order ID generated by Bitfinex is > Integer.MAX_VALUE. This means, orders are executed on the exchange but we cannot process the response due to json parsing exception. This puts the user at risk in many ways like hedging and trade reconciliation as the user doesn't even know that he just executed an order. Please note that this can happen to all the Exchange Packages that uses INTEGER for ID instead of LONG or whichever is appropriate.

kisapmata:
forgot to mention that this impacts also order cancellation and retrieval of order status etc.... anything that references to the order id.

jheusser:
already fixed in develop. please pull latest version before reporting
issues.

On 7 Apr 2017 7:35 am, "kisapmata" <notifications@github.com> wrote:

forgot to mention that this impacts also order cancellation and retrieval
of order status etc.... anything that references to the order id.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/timmolter/XChange/issues/1430#issuecomment-292453849>,
or mute the thread

kisapmata:
i see. gotcha. I've been having problem syncing my forked version recently so I missed out this changes. But I have to say that Bitstamp (and others) is going the same path but is not critical at this moment.

kisapmata:
closing this issue

PR #1313, opened 2016-10-07
Gemini support

jheusser:
Initial commit of Gemini REST API.

As Gemini REST is 100% compatible with Bitfinex this is an initial dirty copy&paste job. Some features of Bitfinex are not supported on Gemini which can be taken out, however the Gemini market data, account, and trading functionality work.

PR #1867, opened 2017-11-05
[Gemini] changed order ID to a long since it now exceeds Integer.MAX_VALUE

ww3456:
This was causing deserialization exceptions. e.g. 
```
si.mazi.rescu.HttpStatusIOException: Can not deserialize value of type int from String "2147483855": Overflow: numeric value (2147483855) out of range of Integer (-2147483648 - 2147483647)
at org.knowm.xchange.gemini.v1.service.GeminiTradeServiceRaw.getGeminiOpenOrders(GeminiTradeServiceRaw.java:39)
at org.knowm.xchange.gemini.v1.service.GeminiTradeService.getOpenOrders(GeminiTradeService.java:51)
```
I've also removed a couple of compiler warnings, and fixed a Kraken unit test (example-assetpairs-data.json had changed 3 months ago) for for #1866.

### DIFF
--- a/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/dto/trade/GeminiLimitOrder.java
+++ b/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/dto/trade/GeminiLimitOrder.java
@@ -10,6 +10,7 @@ import org.knowm.xchange.dto.trade.LimitOrder;
  * Poloniex order response contains details of any trades that have just executed in the order entry return value. If a LimitOrder of this type is
  * supplied to the trade service orderEntry method it will be populated with this information.
  */
+@SuppressWarnings("serial")
 public class GeminiLimitOrder extends LimitOrder {
 
   private GeminiOrderStatusResponse response = null;
--- a/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/dto/trade/GeminiCancelOrderRequest.java
+++ b/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/dto/trade/GeminiCancelOrderRequest.java
@@ -13,7 +13,7 @@ public class GeminiCancelOrderRequest {
 
   @JsonProperty("order_id")
   @JsonRawValue
-  private int orderId;
+  private long orderId;
 
   /**
    * Constructor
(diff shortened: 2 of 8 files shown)


npomfret:
Could we change it to a string?  Id's aren't numbers.  Just because they are numeric to look at, that doesn't make them a number.  You can't add two id's together for example, they have no numeric properties, and as such I think a string is a more suitable type (unless we introduce an Id class one day).

ww3456:
In general terms, I don't see a fundamental problem with a (trading or otherwise) system using numeric IDs, some might choose do this for performance reasons, for example. 

It would seem reasonable that the exchange specific DTOs should use the same type as the exchange, and this can be mandatory when sending data to the exchange, depending on how type-safe their server side is. The Gemini API seems slightly confused about whether their order IDs are numeric or strings, on order placement is a returned in the [status response](https://docs.gemini.com/rest-api/#order-status) as a string, but on both the order status query and the [cancel](https://docs.gemini.com/rest-api/#order-status) method it seems to want a number. 

Obviously for the XChange common interfaces (e.g. return value of TradeService.placeLimit) the order ID is represented as a string to support the range of different exchanges.  

The last similar issue I remember was caused by Bitifinex (#1409)  and since the Gemini implementation was based on Bitfinex (#1313) I don't suppose this particular situation should come as much of a surprise. 

I suppose if it is a concern you could check if any exchanges of particular interest that are storing IDs as ints and potentially change them to longs, but I'm not sure a wholesale switch of all IDs to strings is either necessary or the correct solution, indeed what is the problem you are trying to address? If it is just integer overflow then changing to long will perhaps move the problem to the exchange side, where it fundamentally belongs if they are using numbers for IDs, e.g. what are they going to do when their identifiers wrap. 

(As metioned by @kisapmata in #1430 the Bitstamp order ID is currently an int.)

xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/dto/trade/GeminiLimitOrder.java
```diff
@@ -10,6 +10,7 @@
  * Poloniex order response contains details of any trades that have just executed in the order entry return value. If a LimitOrder of this type is
  * supplied to the trade service orderEntry method it will be populated with this information.
  */
+@SuppressWarnings("serial")
```
timmolter: in general, I don't like these `suppressWarnings` in the code, as you can configure your IDE to not show warnings.

timmolter:
Many times IDs are indeed integers or longs, i.e. primary keys in databases. If an exchange is obviously using integers or longs the raw impl. should use integers or longs. However, the generic API remains a String as it needs to handle the case for exchanges using other schemes such as UUIDs, etc.

PR #1870, opened 2017-11-05
Implement remainingAmount via cumulativeAmount

EzBar:
Order class already holds cumulativeAmount, so remainingAmount can be directly retrieved via originalAmount.subtract(cumulativeAmount). 
- Order.getRemainingAmount() method for that purpose. LimitOrder.remainingAmount was removed.
- C'tors and Builders were updated accordingly
- All relevant exchanges adapters were updated accordingly.

NOTE: Consider removing LimitOrder.getFilledAmount() which is the same as Order.getCumulativeAmount and causes confusion.


timmolter:
OK, I decided to leave `LimitOrder.getFilledAmount()` for convenience. Also you can use the LimitOrder builder with remaining amount for convenience. Only cumulative amount is an actual class member though. I guess it could even be made cleaner somehow.

PR #1873, opened 2017-11-07
Fix Bitstamp funding API and add Ripple support for Bittrex

roeiba:


roeiba:
I am afraid bitstamp request LTC address still doesn't work.

PR #1880, opened 2017-11-11
[cexio, bitbay]

npomfret:


timmolter:
We have an Intellij formatter in `etc` that would be preferable.

npomfret:
ok, thanks.  i didn't realise - I'll check it out

npomfret:
Done.  Only one file changed.  I guess the formatter isn't that different from the default.

PR #1896, opened 2017-11-21
rest calls implementation for the websocket api

cymp:
this only allow the user to command the websocket with api calls (start, keep alive and close websocket stream)


cymp:
wrong files committed, working on a fix

PR #1908, opened 2017-11-26
Add support for Liqui.io

cerko2:


timmolter:
@cerko2 Could you fix the conflicts? Should be easy. Then I can pull...

cerko2:
@timmolter They should be resolved now.

PR #1941, opened 2017-12-05
fix minSize size scale computation

iddqd1:
fix minSize size scale computation


npomfret:
I din't think this is the correct solution. `product.getQuoteIncrement().scale()` isn't what rounding is for, it's the minimum tick size.  I think it should just be removed and the code should be:

        BigDecimal minSize = product.getBaseMinSize();
        BigDecimal maxSize = product.getBaseMaxSize();

PR #1948, opened 2017-12-05
[resolves #1947] Version mixup with fasterxml-jackson

### LINKED ISSUE #1947: Version mixup with fasterxml-jackson
```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ xchange-coinbase ---
[INFO] org.knowm.xchange:xchange-coinbase:jar:4.3.1-SNAPSHOT
[INFO] +- org.knowm.xchange:xchange-core:jar:4.3.1-SNAPSHOT:compile
[INFO] |  +- com.github.mmazi:rescu:jar:2.0.1:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.9.1:compile
[INFO] |  |  |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.9.0:compile
[INFO] |  |  |  \- com.fasterxml.jackson.core:jackson-core:jar:2.9.1:compile
[INFO] |  |  +- javax.ws.rs:javax.ws.rs-api:jar:2.1:compile
[INFO] |  |  +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO] |  |  +- net.iharder:base64:jar:2.3.9:compile
[INFO] |  |  \- oauth.signpost:signpost-core:jar:1.2.1.2:compile
[INFO] |  |     \- commons-codec:commons-codec:jar:1.3:compile
[INFO] |  \- commons-io:commons-io:jar:2.6:compile
[INFO] +- org.slf4j:slf4j-api:jar:1.7.25:compile
```

Also fix the funky import ordering issue in coinbase

tdiesler:


timmolter:
A couple questions:

1. How does this fix the Coinbase test issue?
1. Why did you duplicate the junit dep?
1. Why did you remove the test scope from the mock deps?

tdiesler:
Hi Tim, do you perhaps also hang out on some IRC channel?

1. It fixes the compile issue in coinbase that relies on a specific import ordering

2. & 3. AFAIK it is good practice to define all deps in dependency management including their respective versions and potential excludes. The scope however is defined at the individual modules level. So junit is not really duplicated, it is first defined and then added as a dependency for all modules.

Later perhaps, it'd be good to move the dependency management section to a BOM module, that can be imported by other projects.

timmolter:
> It fixes the compile issue in coinbase that relies on a specific import ordering

How did you fix it? The fasterxml deps version or in the code? What was the trick?

tdiesler:
The diff is hard to read. It however only moves the the CoinbaseOrderStatusDeserializer a scope higher i.e. out of the enum. Then the corresponding import can be removed & and the compiler is happy.

Not sure if this is a compiler bug.

timmolter:
Thanks for finding and fixing that! It was bothering me big time. 

BTW, I reverted some of your changes to the pom. I hope you're not offended...

Thanks again!

PR #1953, opened 2017-12-06
api v2 on authenticated api calls

cymp:


cymp:
Sorry for the mess, it should be fine now.

PR #1956, opened 2017-12-07
Respect Bittrex' usage of BCC for BitcoinCash

Tillerino:
XChange detected BitcoinCash as BitConnect for Bittrex. These two commits should fix that.

The `Currency` instance for BitcoinCash will still return the currency code "BCC" when created by a call to the Bittrex Service.

(Sorry about replacing the entire file for `BittrexUtils`. It was originally using CRLF.)

### DIFF
--- a/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/BittrexAdapters.java
+++ b/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/BittrexAdapters.java
@@ -45,6 +45,13 @@ public final class BittrexAdapters {
 
   }
 
+  public static Currency adaptCurrency(String bittrexCode) {
+    if (bittrexCode.equals("BCC")) {
+      return Currency.BCH.getCodeCurrency("BCC");
+    }
+    return Currency.getInstance(bittrexCode);
+  }
+
   public static List<CurrencyPair> adaptCurrencyPairs(Collection<BittrexSymbol> bittrexSymbol) {
 
     List<CurrencyPair> currencyPairs = new ArrayList<>();
@@ -55,10 +62,18 @@ public final class BittrexAdapters {
   }
 
   public static CurrencyPair adaptCurrencyPair(BittrexSymbol bittrexSymbol) {
+    return adaptCurrencyPair(bittrexSymbol.getBaseCurrency(), bittrexSymbol.getMarketCurrency());
+  }
 
-    String baseSymbol = bittrexSymbol.getMarketCurrency();
-    String counterSymbol = bittrexSymbol.getBaseCurrency();
-    return new CurrencyPair(baseSymbol, counterSymbol);
+  public static CurrencyPair adaptCurrencyPair(String bittrexBaseCurrency, String bittrexMarketCurrency) {
+    Currency baseCurrency = adaptCurrency(bittrexMarketCurrency);
+    Currency counterCurrency = adaptCurrency(bittrexBaseCurrency);
+    return new CurrencyPair(baseCurrency, counterCurrency);
+  }
+
+  public static CurrencyPair adaptExchangeString(String exchangeString) {
+    String[] currencies = exchangeString.split("-");
+    return adaptCurrencyPair(currencies[0], currencies[1]);
   }
 
   public static List<LimitOrder> adaptOpenOrders(List<BittrexOpenOrder> bittrexOpenOrders) {
@@ -75,8 +90,7 @@ public final class BittrexAdapters {
   public static BittrexLimitOrder adaptOpenOrder(BittrexOpenOrder bittrexOpenOrder) {
 
     OrderType type = bittrexOpenOrder.getOrderType().equalsIgnoreCase("LIMIT_SELL") ? OrderType.ASK : OrderType.BID;
-    String[] currencies = bittrexOpenOrder.getExchange().split("-");
-    CurrencyPair pair = new CurrencyPair(currencies[1], currencies[0]);
+    CurrencyPair pair = adaptExchangeString(bittrexOpenOrder.getExchange());
 
     return new BittrexLimitOrder(type, bittrexOpenOrder.getQuantity(), pair, bittrexOpenOrder.getOrderUuid(),
         BittrexUtils.toDate(bittrexOpenOrder.getOpened()), bittrexOpenOrder.getLimit(), bittrexOpenOrder.getQuantityRemaining(),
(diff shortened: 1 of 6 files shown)


npomfret:
I'm quite concerned about this change (sorry!).  

Some exchanges just use a different symbol for BitcoinCash (BCH v's BCC).  It's unfortunate and needs to be addressed, but I think we need a solution that covers all exchange implementations.  Mangling the currency name in just one exchange could cause significant confusion.

timmolter:
So, if I understand this correctly, you want to call methods using BCH, but then for Bittrex it converts it to BCC before sending off the request?

timmolter:
@npomfret My opinion on this matter is that if Bittrex (or whatever) uses BCC for Bitcoin Cash, then client code should just use BCC. One can design the client code to elegantly handle this.

Tillerino:
@timmolter Correct. You call the methods with `BCH` or `BCH.getCodeCurrency("BCC")` or `BCC`. You _get_ `BCH.getCodeCurrency("BCC")` in all objects returned from bittrex. That way `equals` returns true for `BCH` and false for `BCC`. `getCurrencyCode` will still return "BCC" like it did before. The only behaviour that breaks is if you are checking equality from a returned currency against `BCC` using `Currency.equals`.

I think that the point of XChange - rather than a collection of JAX-RS interfaces, which by itself is already heaven-sent :) - is to abstract away from the exchanges and provide a standardized interface. IMO that means that these inconsistencies need to be handled in the `Exchange` implementation, not client-side.

I agree that the perfect way would be to solve this for all currencies and all exchanges simultaneously but I also fear that this will simply not happen because it is such a large and annoying task. I think the only way to tackle this is to go exchange-by-exchange.

@npomfret Setting aside that this only fixes one currency for one exchange: do you agree with the proposed usage of `Currency` and it's alternative codes?

npomfret:
@timmolter i agree, handling the BCC/BCH _madness_ for the moment is best done in the client code.  I think the current `Currency` class might have better been named `CurrencyCode` (or something).  The currency we're discussing ( _BitcoinCash_) isn't adequately represented in the `Currency` class in it's current form - we're missing an abstraction.  Anyway, just some food for thought.

@Tillerino I don't know if I understand the impact of this change:

```public static final Currency BCH = createCurrency("BCH", "BitcoinCash", null, "BCC");```

Is it not confusing to have BCC in here?  Should BCC be another top level Currency?  It is in my app at the moment.

**Comment by Tillerino** [comment]
@npomfret Yes `BCC` is a top level Currency, but it denotes BitConnect, as can be seen in the attributes:

```java
public static final Currency BCC = createCurrency("BCC", "BitConnect", null);
```

The change you mentioned adds "BCC" as an alternative code to `BCH`, which is accurate. Since `Currency` instances are compared based on their attributes, not on the actual code used, this is enables us to carry a `BCH` instance with the "BCC" symbol. It looks to me like the author of the `Currency` class gave us all the tools to deal with the symbol confusion.

With the change I pushed, you can have "BCC" both BitConnect and BitcoinCash:

```java
Currency bcc = Currency.BCC;
Currency bch = Currency.BCH;
Currency bchMod = Currency.BCH.getCodeCurrency("BCC");
Currency bccInstance = Currency.getInstance("BCC");

System.out.println(bchMod.getCurrencyCode()); // outputs "BCC"
System.out.println(bch.equals(bchMod)); // returns true
System.out.println(bcc.equals(bchMod)); // returns false
System.out.println(bccInstance == bcc); // return true
```

The way `bchMod` is constructed is exactly how `BittrexAdapter` now handles the symbol "BCC". IMO this behaviour makes sense and is how the author of `Currency` intended it to be used.

timmolter:
I think the alternative currency code was intended for a different use-case in mind, for example  `BTC` and`XBT`. In this case, the symbol is universally accepted and not disagreed upon between different exchanges, like the case for BCH and BCC. 

I'm sorry to close this, but my decision is based on the fact that since this can be easily handled in client code, by choosing BCC for bitcoincash for the bittrex exchange, that's a much simpler solution and keep XChange code simpler, easier to understand and eliminates surprises for client code. While BCC is technically bitconnect according to our Currency class, it's still bitcoin cash according to bittrex. 

The Currency class is at its root just a way to provide a symbol. Maybe it's over engineered and simple Strings would suffice (as it was originally, IIRC).

npomfret:
Sorry for the slow reply.

My view is that we need to have a symmetric api - so if I place a trade using the BCH constant and then ask for my open orders I will have one for BCH (and not BCC in the case where an exchange uses BCC).  I don't think this will be the case with the proposed change and so I agree with @timmolter that this isn't the right way to go for now.

However, I have been giving it some thought and if we wanted to better encapsulate the concept of currency v's exchange specific currency codes then I'd be happy to implement it.  It's quite a big change though.  

My initial thoughts are that we need an interface which converts between exchange specific codes and the Currency class - both ways.  Each exchange should have an instance (most can use a default one) in their base class and should use it to convert from Currency to a string, and from a String to a Currency.  

Use of `Currency.getInstance` would be strongly discouraged, as would `currency.getCurrencyCode()` or `currency.toString()`. Instead each exchange should use the interface, which I imagine would look something like:

    interface CurrencyConverter {
        String toCurrencyCode(Currency currency);

        Currency fromCurrencyCode(String currency);
    }

    class DefaultCurrencyConverter implements CurrencyConverter {

        @Override
        public String toCurrencyCode(Currency currency) {
            return currency.getCurrencyCode();
        }

        @Override
        public Currency fromCurrencyCode(String currency) {
            return Currency.getInstance(currency);
        }
    }
    
    class WeirdExchangeCurrencyConverter extends DefaultCurrencyConverter {
        @Override
        public String toCurrencyCode(Currency currency) {
            if(currency.equals(Currency.BCH))
                return "BCC";
            return super.toCurrencyCode(currency);
        }

        @Override
        public Currency fromCurrencyCode(String currency) {
            if(currency.equalsIgnoreCase("BCC"))
                return Currency.BCH;
            return super.fromCurrencyCode(currency);
        }
    }

The proposed `CurrencyConverter` class could also be used to deal with all the lower-casing that goes on in some exchanges.

Step 1 would be to introduce the class, and just make all exchanges use the default implementation - this would be a non-function change.  It would be done over time and wouldn't need to be checked in as a single commit.

Step 2 would be to implement exchange specific CurrencyConverters (for Bittrex for example) that need to somehow mangle the currency code that get's returned by the Currency class.  So long as all conversions go via the converter this should solve the problem globally.

PR #1970, opened 2017-12-11
Introduce ticker information into Independent Reserve market plugin

perlboy:
Hi there, this introduces the standardised Ticker information for the Independent Reserve exchange based in Australia. No much changes here probably not perfect but functions. IR provides everything except Open price. Content lifted from Bitstamp and BTCMarkets plugins.

PR #1971, opened 2017-12-11
Introduce Ticker support or Independent Reserve

perlboy:
Resubmit of #1970 into develp branch with no stray pom version


perlboy:
Apologies, new to pull requests. Not sure if these post merge's are now a problem?

perlboy:
Have fast-forwarded my fork so that the overall commit changes are just this change. Round about way of doing things but this should be sufficient?

Issue #1991, opened 2017-12-16
Support OKEX Futures Price Limits


douggie:
OKEX imposes max and min price limits, above which longs cannot be placed, and below which shorts cannot be placed.

In rapidly rising market, the max price limit can be below the current best bid/best ask in an  to cool down the market, also in a rapidly failing market, the min price limit can be above the current best bid/ask. In such instances placing limit orders at the current market, get rejects as do to market orders. The only price you are able to place at is at or below the max price for longs, or at or above the min price for shorts, which requires leveraging the Get Limit High Price or Limit Low Price (https://www.okex.com/api/v1/future_price_limit.do?symbol=[symbol]&contract_type=[contract]).

In rapidly moving markets, it would be useful to know the max/min prices via the xchange libs and as this is just an okex thing, it seems to make sense to add it to either the  OkCoinFuturesMarketDataService or OkCoinFuturesTradeService. I am thinking the OkCoinFuturesTradeService given that you will probably have a trade service object that you want to place the order with, but might want to check before, or after the order is rejected what the max/min price limits are.

Any concerns with adding this to the OkCoinFuturesTradeService?

PR #1997, opened 2017-12-17
#1991

douggie:


douggie:
Was formatted with XChange_Eclipse_2017_04.xml	formatter, but still shows some formatting differences.

PR #2002, opened 2017-12-18
Fix typo in Binance getFundingHistory

jamding:
The Binance implementation commonly uses this pattern to check types when accepting an instance of `TradeHistoryParams`, but for `AccountService#getFundingHistory` there's a typo which causes all calls to `getFundingHistory` to either throw a (userland)`RuntimeException` or real `RuntimeException` when you can't cast params as a `TradeHistoryParamCurrency` (line 85).

See https://github.com/timmolter/XChange/blob/develop/xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java#L96-L98 for intended behavior.


jamding:
@timmolter

PR #2008, opened 2017-12-19
[Bitmex] Initial partial implementation

cristianlucaci:


timmolter:
Can you please remove the lombok dependency?

cristianlucaci:
@timmolter  yes, I can, but arent we allowed to use 3rd party libraries? I didnt see anything about that in "Implementation Best Practices"

timmolter:
My apologies, I didn't add a note in there about that yet, but I will, including the reasoning. The reason  we want to keep the number of 3rd party deps to an absolute minimum is 2-fold:

1. As a maintainer, I cannot possibly learn all the new/additional deps that people want to add. It's already a monumental task to keep up with things as it is.
1. For new people and existing contributors wanting to go in and hack/improve the different modules, needing to deal with unfamiliar deps is a barrier to entry.

I understand the tradeoffs, don't get me wrong. :)

cristianlucaci:
lombok was removed

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexAccountService.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public TradeHistoryParams createFundingHistoryParams() {
+    return null;
+  }
+
+  @Override
+  public List<FundingRecord> getFundingHistory(TradeHistoryParams params) throws IOException {
+    return null;
```
timmolter: Instead of returning null for all these, throw a NotYetImplementedException or NotAvailableFromExchangeException

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexAccountServiceRaw.java
```diff
@@ -0,0 +1,81 @@
...
+      throw new ExchangeException(e);
+    }
+  }
+
+  public List<BitmexMarginAccount> getBitmexMarginAccountsStatus() throws IOException {
+
+    try {
+      return bitmex.getMarginAccountsStatus(apiKey, exchange.getNonceFactory(), signatureCreator);
+    } catch (BitmexException e) {
+      throw new ExchangeException(e);
```
timmolter: You should handle the exceptions with a handle exceptions method in BitmexBaseService. See GDAX impl. on develop. That way, it's easier for clients to handle certain types of exceptions.

timmolter:
Looks great so far, just see my two comments.

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexAccountService.java
cristianlucaci: done, will do this for all exchanges going forwards.

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexAccountServiceRaw.java
cristianlucaci: done

jnorthrup:
i am having issues with the bitmex progress to date.  i ran the swagger generation from bitmex to fix the stackoverflow in getAccountInfo.  the bitmex branch in my tree holds the results.  the swagger appears to under-perform on the necessary api and secret signature features, and documentation, however the bitmex api is absolutely comprehensive.

that said, on the matter of deps, this codebase lacks good interface abstractions, so wrapper interfaces around impl drop-in replacements is not an option.  such like AccountInfo being a final concrete class requires drastic measures  to do what better abstractions would enable.

PR #2009, opened 2017-12-20
Added recvWindow to binance

KevinJannis:
The recvWindow parameter can be set using:

`specification.setExchangeSpecificParametersItem("recvWindow", 10000L);`


timmolter:
Thanks. Please use [binance] at the start of the commit messages for binance in the future. :)

PR #2011, opened 2017-12-20
[BitFlyer] Partial Implementation

cristianlucaci:


xchange-bitflyer/src/main/java/org/knowm/xchange/bitflyer/service/BitflyerBaseService.java
```diff
@@ -29,4 +34,15 @@ public BitflyerBaseService(Exchange exchange) {
         exchange.getExchangeSpecification().getApiKey());
   }
 
+  protected ExchangeException handleError(BitflyerException exception) {
+    if (exception.getMessage().contains("Insufficient")) {
```
timmolter: The strings here need to be specific to bitflyer according to their API docs, not just copy-pasted from GDAX. I'll accept it for now, but I probably won't work as intended, FYI.

timmolter:
Can you resolve the conflict? You'll need to update your upstream with the timmolter fork and merge conflicts. The re-push.

xchange-bitflyer/src/main/java/org/knowm/xchange/bitflyer/service/BitflyerBaseService.java
cristianlucaci: Sorry. but I couldnt find much information about the specific errors that BitFlyer will send, except one very generic error response. Will add the errors once encountered while testing. 

Error Response:
{
  "status": -700,
  "error_message": "This account has not yet been authenticated",
  "data": null
}

timmolter: sounds good!

PR #2019, opened 2017-12-21
Implementing equals() and hashCode() for UserTrade.

stevenuray:
I'm a user of XChange but this is my first contribution. 

I would benefit from these modifications, and I thought others would too. 

If I am not following accepted conventions, proper PR protocol, etc please let me know and I'll try to make necessary changes.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/trade/UserTrade.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/trade/UserTrade.java
@@ -1,14 +1,15 @@
 package org.knowm.xchange.dto.trade;
 
-import java.math.BigDecimal;
-import java.util.Date;
-
 import org.knowm.xchange.currency.Currency;
 import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.dto.Order.OrderType;
 import org.knowm.xchange.dto.marketdata.Trade;
 import org.knowm.xchange.service.trade.TradeService;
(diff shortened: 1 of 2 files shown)


andre77:
just out of curiosity, why do you need those methods?

stevenuray:
@andre77

I have an object in my application that has a UserTrade as a property and has other properties that add data about the UserTrade that is specific to my application.  

I want to implement equals() and hashCode() on my custom object so that two objects with the same values are considered equal, and this would be made easier/cleaner for me if UserTrade implemented equals() and hashCode() so that two UserTrades with equal values are considered equal. 

I assumed other users would desire the same behavior from UserTrades. 

Perhaps this was an incorrect assumption.

If this isn't going to be a net positive for the community, I'll close the PR.

andre77:
@stevenuray 
i do not mind to add those methods, totally fine to me
in your case, you probably just could extend the existing UserTrade and create a new class containing all the properties required by your application. In that new class you can than add what ever you like, the hash and equals method, which would use the properties from the super classes etc.

just my 2c

Marry Christmas! :-)

timmolter:
I'm not yet against it, but haven't had time yet to take a close look. Please don't close this PR yet. Thanks!

stevenuray:
@andre77 

I agree extending the class would work too, in this case.

Thanks for the reviews/merge, and also all the work on the XChange project!

Happy Holidays!

PR #2062, opened 2017-12-28
[resolves #2061] [Binance] Add support for Stop-Limit orders

### LINKED ISSUE #2061: [Binance] Add support for Stop-Limit orders

tdiesler:


timmolter:
Is a stop limit order different than a stop loss order (which is in the end a market order)?

tdiesler:
Yes, it is different. Have a look at [this](https://support.binance.com/hc/en-us/articles/115003372072-How-to-use-Stop-Limit-Function)

The enum already has STOP_LOSS, so it might get supported in the future. For now however I could not get it to work.  The [api docs](https://www.binance.com/restapipub.html#user-content-enum-definitions) are outdated.

douggie:
hi tim, they are indeed different stop loss becomes a market order when
triggered, where as stop limit becomes a limit order at the trigger price
once triggered.

timmolter:
OK, so I understand this correctly, you can place a limit order along with a stop limit order at the same time? Is this essentially the same as placing 2 limit orders one after the other?

My gut feeling is that this should be a new class, perhaps one that extends `LimitOrder`, but I'm not yet sure either way.

tdiesler:
Unfortunately, on Binance every order operates on a distinct set of coins. So you cannot place a sell order on all available coins and another sell order with different parameters on the same set of coins.

A stop-limit order is a two step process. When the market prices reaches the stop price, a limit order is placed with potentially a limit price that is different from the stop price.

tdiesler:
Closing in order to wait for other binance PRs

PR #2088, opened 2018-01-01
Added timeDelta for withdrawal-requests endpoint on bitstamp and using V2 now.

tonyskulk:


tonyskulk:
Seems like travis has failed to build. How can I start it again?

timmolter:
You cannot restart the build, but I just did.

timmolter:
Looks like there's a compile error.

PR #2103, opened 2018-01-03
[BitFinex] Implemented proper wallets adaptation

### LINKED ISSUE #1535: [Bitfinex] (DEV)  AccountInfo .getWallets() not returns wallet names  like deposit, exchange,..
AccountInfo .getWallets() not returns wallet names  like deposit, exchange,.. 
It does return some info like total balance value usd and how much is used and is free but it doesn't differenciate between  deposit, exchange, margin .anymore.

hurricup:
Wallets are now separated by types

Fixes #1535

### DIFF
--- a/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/BitfinexAdapters.java
+++ b/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/BitfinexAdapters.java
@@ -1,21 +1,8 @@
 package org.knowm.xchange.bitfinex.v1;
 
-import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.Collection;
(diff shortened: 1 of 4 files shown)


hurricup:
Do not merge!!! Need fixes

npomfret:
can you explain what the change in behaviour for this pr is?

hurricup:
There are 3 different wallets on BitFinex:

- Exchange wallet is used for exchanging funds between our supported currencies.
- Margin wallet is used for trading leveraged short and long positions.
- Funding wallet is used for providing margin funding/financing to other traders.

atm. XChange just sums balances on these and creates one wallet object. But this makes a little sense.
I've implemented 3 wallets with type as id.

npomfret:
@hurricup that sounds good.  thanks

PR #2105, opened 2018-01-04
[cexio] Completing implementation of requestDepositAddress.

bryantharris:
Fleshing out the implementation of requestDepositAddress.


timmolter:
There's a compile error.

bryantharris:
What is the error, maybe I missed a file when I forked and copied it from the original branch.

bryantharris:
Sorry, the cexio package itself uses the prefix CexIO and sometimes Cexio for class names and it looks like when I was going through the files in an attempt to match pre-existing coding/name conventions I did so without also updating the file name.  Should be fixed.

bryantharris:
@timmolter Created new pull request, you can close this one.

PR #2109, opened 2018-01-04
[Bit-Z] New Exchange Implementation

Flemingjp:
A basic implementation of the [Bit-Z Exchange](https://www.bit-z.com/api.html), as per this request [https://github.com/timmolter/XChange/issues/2049](https://github.com/timmolter/XChange/issues/2049)

Currently offering an implemented MarketDataService with appropriate units tests. Scope of the implemenation is now onto implementing a Trading Service.

One ticker demo is given.


timmolter:
I think you forgot to add the new module to the examples pom.xml.

pom.xml
```diff
@@ -152,7 +152,8 @@
         <module>xchange-vircurex</module>
         <module>xchange-yobit</module>
         <module>xchange-wex</module>
```
timmolter: please move this in alphabetical order in the list

PR #2112, opened 2018-01-05
[Binance] Add openPrice to ticker

KevinJannis:


KevinJannis:
`ticker.getOpen()` is now always null for binance.

PR #2121, opened 2018-01-08
hitbtc trade history fix and various minor non-functional changes

npomfret:


tofti:
Any fixes that address the history, and pagination are +1 from me. Some of this functionality regressed quite significantly in versions that supported the v2 version of hitbtc, from versions that supported v1 of the API.

Issue #2129, opened 2018-01-11
java 9 java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter     at org.knowm.xchange.binance.service.BinanceHmacDigest.digestParams (BinanceHmacDigest.java:63)


jnorthrup:
same as https://github.com/jexler/grengine/issues/5

java.lang.NoClassDefFoundError: javax/xml/bind/DatatypeConverter
    at org.knowm.xchange.binance.service.BinanceHmacDigest.digestParams (BinanceHmacDigest.java:63)

rafalkrupinski:
Can this be closed now?

PR #2135, opened 2018-01-11
jdk9 runtime modules error

jnorthrup:
addresses  #2129


timmolter:
Can you check out `DateUtils` in `xchange-core` and use that instead?

jnorthrup:
as requested -- I'm not a rebase ninja, but I would suggest a rebase to remove the intermediate commits.  

DateUtils seems inconsistent wrt libraries/sdk utilized but my goal was simply to start and run without care and feeding of java 9 modules, not changing the codebase.   my intellij uses jdk8, commandline uses jdk9, java9 is the release, and 99.9% of github java and 100% of all kotlin is jdk 1.8 or less.

PR #2138, opened 2018-01-12
[Binance] Currency returned in balance cannot use ==

### LINKED ISSUE #2137: [Binance] Currency returned in balance cannot use ==
It seems that balance contains new currency instances instead of using the available constants. Also, the AccountService does not use the Exchange's time stamp

tdiesler:
resolves #2137

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/BinanceBalance.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/BinanceBalance.java
@@ -2,19 +2,39 @@ package org.knowm.xchange.binance.dto.account;
 
 import java.math.BigDecimal;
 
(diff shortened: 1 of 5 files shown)


uwemaurer:
I think it would be better to avoid this extra request to get the time before doing the actual request.
Is it really needed?

tdiesler:
I believe so. If the client's system clock is slightly out of sync with that of the binance system, these requests will fail. Hence relying on these clocks to be in sync introduces another factor of uncertainty.

uwemaurer:
yes true. A disadvantage is the extra roundtrip and that the rate limit is reached faster then. Maybe the time request could be done once during initialization to calculate the delta between client and server time. And for a new request it could add the delta then to the clients time to adjust it to the server time.

tdiesler:
Yes, that is good idea - let me see to it

Issue #2148, opened 2018-01-13
[35 mBTC Bounty] Implement Coinegg


cymp:
Hello,

I am paying a 35mBTC bounty to whomever fully implements the [Coinegg ](https://www.coinegg.com/explain.api.html)before January 21st. 25mBTC otherwise.

Thanks.

Flemingjp:
I can take a look at implementing this exchange also.

cymp:
Please proceed, thanks!

Flemingjp:
@cymp  See #2156 As with your other request, this API is also half finished, I check this at first with their customer support team. I've implemented everything I possibly can with the appropriate tests. Currently they have no working Trade service or Account service and I've implemented the data structures for these services to allow it to be quickly implemented when they sort it out.

cymp:
wow this is really annoying, because it is completely useless for me like that... Well thank you I can pay you 18 mBTC now and the 17 remaining when you can finish this. I'm trusting you on this part and will test the whole stuff once it is finished.

Flemingjp:
@cymp I just message the exchange again. It's difficult due to language barriers. They've sent me some test PHP code which I'm testing, so I may be able to push through. I'll update ASAP

Flemingjp:
@cymp Can confirm the PHP code works. Their API documentation is wrong in comparison to their code. Will update in a few hours

Flemingjp:
@cymp I've now implemented the basic Trade Service functions and the Accounts service. Please begin testing them ASAP and report and issues and bugs to me. Feedback with some code samples would be very useful

cymp:
Thanks, will do asap.

Flemingjp:
@cymp Any updates?

cymp:
Sorry I was away these last days, I will be back tomorrow and work on it.

Flemingjp:
@cymp No worries. I'll be busy the next few days also. Will be able to fix anything over the weekend

cymp:
@Flemingjp I am starting to work a bit on integrating your work, do you have an idea what can be used to get all our open orders?

Flemingjp:
@cymp use the getCoinEggTradeView from the CoinEggTradeServiceRaw

cymp:
@Flemingjp ok thanks. By the way, I noticed that you extract the order books by calling "/api/v1/orders/", which returns, if I am correct reading the documentation, "Return the most recent 100 transactions", which does not look like an orderbook. Did I miss something?

[...]

PR #2156, opened 2018-01-14
Implementation of CoinEgg Exchange

Flemingjp:
Implemenation as of request #2148. This is as far as it can be currently implemented

- Fully implemented MarketData Service
- Partial implementation of Account Service
- DTO Objects and Deserializaition of Trade Service Objects


xchange-coinegg/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    </organization>
+
+    <!-- Parent provides default configuration for dependencies -->
+    <dependencies>
+
+        <dependency>
+            <groupId>org.knowm.xchange</groupId>
+            <artifactId>xchange-core</artifactId>
+            <version>4.3.3-SNAPSHOT</version>
+            <exclusions>
```
timmolter: There should be no dependency on pusher in this project, so you can delete this exclusion.

xchange-coinegg/src/main/resources/logback.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
timmolter: This file shouldn't be necessary. Please delete.

PR #2159, opened 2018-01-14
Implemented getOrder(...) in several repositories

traviscollins:
[GDAX] Implemented getOrder(...) for market and limit orders with tests.


traviscollins:
Added Kraken getOrder(...) implementation with unit test for market order. It had LimitOrder partial done only for the getOpenOrders(...) methods. Expanded it to include LimitOrder for the getOrder(...) method.

traviscollins:
I'm done for now. Will look into Poloniex and Binance in the future.

PR #2160, opened 2018-01-15
Binance and Gate.io updates

cymp:
- Gate.io: fixed  API endpoint name
- Binance: added exchange info API call


timmolter:
Can you please sync up and merge conflicts? Sorry, not your fault...

cymp:
@timmolter Should be fine now.

PR #2164, opened 2018-01-15
[abucoins] new module

bryantharris:
I'm working with Abucoins and so I created this module based on the style/api of your other XChange modules.


bryantharris:
There's one area of the API I wasn't sure exactly how to handle.

From the Abucoins API documentation
> before and after cursors are available via response headers ac-before and ac-after. Your requests should use these cursor values when making requests for pages after the initial request.

As you can see it returns cursor related information in the HTTP header of the response.  I wasn't sure how to access the HTTP response so that I could add these fields to the POJO that currently just represents the JSON portion of the response.

The module is still very usable, but if you have an example of grabbing header fields I doubt it would take me too long to adapt the approach to add full cursor support.

timmolter:
Hmmmm, I'm not sure about grabbing the header values. @mmazi might know or take a look at ResCU...

mmazi:
Unfortunately, rescu currently only supports reading response headers for [exceptional responses](https://github.com/mmazi/rescu/wiki/Exception-handling) -- see [code here](https://github.com/mmazi/rescu/blob/develop/src/main/java/si/mazi/rescu/RestInvocationHandler.java#L131).

It seems this could easily be adapted to support regular response model classes that extend `HttpResponseAware` -- I'll happily accept a pull request with adequate tests.

You can also try replacing rescu in your module with a more flexible client library, like the one in [Resteasy](https://docs.jboss.org/resteasy/docs/3.1.4.Final/userguide/html/RESTEasy_Client_Framework.html#d4e2024) or [Jersey](https://jersey.github.io/apidocs/latest/jersey/index.html). This will probably add many new dependencies, however.

mmazi:
I've implemented my above suggestion (response model classes that extend `HttpResponseAware`) in rescu; feel free to use it.

PR #2165, opened 2018-01-15
[cexio] Adding implementation for  request deposit address

bryantharris:
Creating new pull request per request.


xchange-cexio/src/main/java/org/knowm/xchange/cexio/service/CexIOAccountService.java
```diff
@@ -40,8 +41,11 @@ public AccountInfo getAccountInfo() throws IOException {
 
   @Override
   public String requestDepositAddress(Currency currency, String... arguments) throws IOException {
+    CexIOCryptoAddress cryptoAddress = getCexIOCryptoAddress(currency.getCurrencyCode());
+    if ( cryptoAddress.getOk().equals("ok"))
+      return cryptoAddress.getData();
 
-    throw new NotAvailableFromExchangeException();
+    throw new IOException(cryptoAddress.getE() + ": " + cryptoAddress.getError());
```
timmolter: This should be an `ExchangeException`

pom.xml
```diff
@@ -421,4 +421,4 @@
         </plugins>
     </build>
 
-</project>
\ No newline at end of file
+</project>
```
timmolter: This seems fishy.

timmolter:
Could you create a new PR after syncing up? There are too many commits in this PR not related to your code changes.

pom.xml
bryantharris: On the updated branch I didn't include this as it seems unnecessary.

bryantharris:
Okay, created a new branch and tried to bring just the files across.  I apologize, I'm obviously not super advanced with GitHub.

After making the Abucoins implementation I also moved the error check/exception throw into the 'Raw' file as that seems like a better place.  Otherwise the implementation is the same.

PR #2167, opened 2018-01-15
[Bitcointoyou] Add Bitcointoyou Exchange

daniloguimaraes:
@timmolter, please review my code. Here is the implementation of Bitcointoyou Exchange, as requested on issue #1692.
I've tested the most of Bitcointoyou API end-points and features, including the authenticated ones.
I couldn't test any order related end-point (create and cancel) because they don't provide any development or test environment.
But order book and order searching and listing it's working pretty well.

Special thanks to:
@jonathascarrijo who helped me with the implementation, 
@brunokarpo who helped me with testing and 
@andrehortab2u who helped opening the issue and providing the Bitcointoyou API documentation.


xchange-bitcointoyou/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+	</organization>
+
+	<!-- Parent provides default configuration for dependencies -->
+	<dependencies>
+	
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>xchange-core</artifactId>
+			<version>${project.version}</version>
+			<exclusions>
```
timmolter: Don't need this.

xchange-bitcointoyou/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+				<exclusion>
+					<groupId>org.java-websocket</groupId>
+					<artifactId>Java-WebSocket</artifactId>
+				</exclusion>
+			</exclusions>
+		</dependency>
+
+		<dependency>
```
timmolter: Isn't pusher for streaming? XChange doesn't support streaming - only REST.

xchange-bitcointoyou/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+		<artifactId>xchange-parent</artifactId>
+		<version>4.3.3-SNAPSHOT</version>
+	</parent>
+
+	<artifactId>xchange-bitcointoyou</artifactId>
+
+	<name>XChange BitcoinToYou</name>
+	<description>XChange implementation for the BitcoinToYou Exchange</description>
+
+	<url>http://xeiam.com</url>
```
timmolter: in 2015 Xeiam was acquired by Knowm, so this info should be changed.

xchange-bitcointoyou/pom.xml
daniloguimaraes: Done!

xchange-bitcointoyou/pom.xml
daniloguimaraes: Yeah, unuseful dependency. Removed!

xchange-bitcointoyou/pom.xml
daniloguimaraes: Sorry about that. Done!

PR #2169, opened 2018-01-16
[Binance] Improve support for candle stick market data

### LINKED ISSUE #2139: [Binance] Improve support for candle stick market data
The kline interval needs a better abstraction as well as the returned kline price

tdiesler:
resolves #2139


xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
```diff
@@ -18,6 +18,7 @@
  * This class is immutable.
  * </p>
  */
+@SuppressWarnings("serial")
```
timmolter: Please remove this.

PR #2182, opened 2018-01-17
[Binance] Add support for Stop-Limit orders

### LINKED ISSUE #2061: [Binance] Add support for Stop-Limit orders

tdiesler:
resolves #2061

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
@@ -3,7 +3,7 @@ package org.knowm.xchange.binance.dto.trade;
 import com.fasterxml.jackson.annotation.JsonCreator;
 
 public enum OrderType {
-  LIMIT, MARKET;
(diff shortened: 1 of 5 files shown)


timmolter:
Should we maybe have separate order objects instead? 

* MarketOrder
* LimitOrder
* StopLossOrder
*StopossLimitOrder

tdiesler:
Your call. IMHO Stop-Limit is a simple enough extension to LimitOrder. Stop-Limit may also conceptually not be a clean extension of Stop-Loss

theo1985:
@timmolter I implemented the StopOrder in #2239.

Stop orders are not only used to exit a position.  They may also be used to enter a position, and as such should be logically separated from LimitOrder (not extended).  This is seen in GDAX, for example.

Maybe we can link a Stop/Limit to an Order by calling: Order.addStopLoss(StopOrder) and/or Order.addTakeProfit(LimitOrder); so as to simulate StopLoss and TakeProfit(StopLossLimit).  

This design would also cover OCO (known as ParentOrders in OkCoin).

timmolter:
I merged #2239, as I liked the implementation better. 

@theo1985 Yeah, the linking of the orders seems like a good idea.

PR #2184, opened 2018-01-17
Added 'fee' to Order, and integrated each exchange that has an obvious fee return value

traviscollins:
fee value in the raw response object. For others, the value is null
or BigDecimal.ZERO following whatever was already present for other
constructor arguments. This is all behind the scenes adapter conversions,
so I don't think it will impact downstream projects using the existing
xchange interfaces.


traviscollins:
I'm done on this for now.

traviscollins:
I added some fixes for Gemini and Bitstamp json files based on the price scales seen in their documentation and in practice.

traviscollins:
Closing this pull request to split out the concerns.

PR #2186, opened 2018-01-18
[abucoins] Updating metadata on currency pairs and currencies from abucoins

bryantharris:
Updating metadata on currency pairs and currencies from ab

Also added Nem, Stratis, and Golem as currencies to base package.


xchange-core/src/main/java/org/knowm/xchange/currency/Currency.java
```diff
@@ -487,6 +491,7 @@ public int compareTo(Currency o) {
   }
 
   private static class CurrencyAttributes implements Serializable {
+    private static final long serialVersionUID = -9010771808048500252L;
```
timmolter: Please remove this UID.

bryantharris: Okay, reverted that commit.

PR #2189, opened 2018-01-19
[kucoin] Implements tickers

janakerman:
Adds Kucoin support for single ticker requests. My plan is to slowly add support for this exchange over time. My first contribution so expecting feedback.

I plan to add all ticker call to the raw exchange next. Will improve support for other endpoints gradually.

(Great contribution guide by the way - Love The library!)


xchange-examples/pom.xml
```diff
@@ -110,6 +110,12 @@
             <artifactId>xchange-kraken</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <!-- XChange Kraken support -->
```
timmolter: Comment is wrong - not Kraken

PR #2193, opened 2018-01-19
Order Status Fee

traviscollins:
Added "fee" to the Order object, and implemented the adapters where the existing raw service already provided the fee (didn't go through documentation for each exchange to modify the raw objects). 

This change is mostly internal to the xchange library and the adapters. The only external facing change is the addition of the fee property to Order (getter and setter).


traviscollins:
Done here.

timmolter:
Could you please fix the conflicts? Then I will merge this....

timmolter:
nevermind, I fixed it.

PR #2194, opened 2018-01-19
updated exchange URL

sirilicious:


theo1985:
This might be invalid, the specification is already conditionally available using:

exchangeSpecification.setExchangeSpecificParametersItem("Use_Intl", true);

sirilicious:
Ok, thanks @theo1985, this works. However, I'm wondering why not using the international setting as default?

theo1985:
I agree with you; as OkEx has the most instruments.  

However, I would also add support for OKCoin CNY, OKCoin USD, OKLink, Jixianyuan and Bafang?

The API seems the same, what changes is the supported symbol list.

PR #2208, opened 2018-01-21
[bitbay] Bitbay withdraw support

bryantharris:
Adding support for 'withdraw' and 'transfer' which correspond to fiat withdrawal and crypto withdrawal.


bryantharris:
Bitbay has support for requesting fiat withdrawals via Bic/Swift.

As a suggestion there should be a default BicSwift class derived from WithdrawFundsParams to generically describe a Fiat currency withdrawal that different exchanges may or may not support.

Ahead of a decision like that I created a class <code>BitbayWithdrawFundsSwiftParams</code> which allows <code>AccountService.withdrawFunds</code> to make this Bic/Swift request.

Just a suggestion, 5 minute change if you decide to go that route.

PR #2212, opened 2018-01-22
[bitbay] I have corrected a bug in transactions method of BitbayTradeService

makarid:
Hello, i have change the currencyPair parameter of transaction method in order to send a "currency.base-currency.counter" message on the bitbay private API.I am posting the API documentation about that to see the problem:

(optional)market: (e.g. BTC**-**USD) if set the result list would be limited to transactions with given currencies. Correct format is cryptoCurrencyShortcut-priceCurrencyShortcut. If parameter not set method returns all transactions ordered by date in descending order.

They want to send a currencyPair with "-" between them and not a currencyPair with "/" as was before.This was the problem that i couldn't get a userTrade list with only one currencyPair.


makarid:
Please check it and add it to master branch.It works from me.I am waiting for your response

Issue #2074, opened 2017-12-30
Implement API for kucoin


jamding:
API docs: https://kucoinapidocs.docs.apiary.io/#introduction/language-parameters

goleary91:
seconded

brunocvcunha:
+1

makarid:
+100

Mawster:
On it

jamding:
Especially the get funding and get trade history APIs :D, lmk if u need
help--might have cycles in next few weeks

odrotleff:
I started off almost at the same time as Mawster, although I'm not sure how fast I will be able to finish. Current progress in my fork.

timmolter:
merged #2235

PR #2235, opened 2018-01-29
[Kucoin] add Kucoin exchange

odrotleff:
Needs (a lot) more tests and exception handling, but should integrate Kucoin as best as their API allows (hopefully).


timmolter:
Looks like something isn't compiling. Can you look into it?

odrotleff:
Yeah I think I will finish with the basics this weekend. Still had some stuff on my todo but noticed that someone else started on the Kucoin implementation. So I wanted to get it out there that I work on it so no work is duplicated.

odrotleff:
#2074

Still have to write some tests and there are probably a few bugs here or there (cant really test extensively if I lose money doing it). Is the code format checked during CI? Because I noticed just now that there is also an eclipse configuration in the parent project and used google style for all of this.

timmolter:
Formatting isn't checked, but we ask that you do use the supplied formatters.

Issue #2063, opened 2017-12-28
BitMex Symbol Adapters


douggie:
Bitmex support the following prompts for the derivatives contracts (https://www.bitmex.com/api/v1/instrument/activeIntervals)

monthly
quarterly
biquarterly
perpetual
weekly

Any objections to using a base.quote.prompt format that would be adapted into the local bitmex symbol as per the following:

XMR.BTC.MONTHLY   -> XMRZ17
XMR.BTC.QUARTERLY ->  XMRH18
DASH.BTC.MONTHLY ->  DASHZ17
DASH.BTC.QUARTERLY ->  DASHH18
XRP.BTC.MONTHLY ->  XRPZ17
XRP.BTC.QUARTERLY ->  XRPH18
BCH.BTC.MONTHLY ->  BCHF18
ZEC.BTC.MONTHLY ->  ZECZ17
ZEC.BTC.QUARTERLY ->  ZECH18
XTZ.BTC.MONTHLY ->  XTZZ17
XTZ.BTC.BIQUARTERLY ->  XTZZ17
BTC. XXX.MONTHLY ->  B_BLOCKSZ17
XBT.USD.MONTHLY ->  XBTZ17
XBT.USD.QUARTERLY ->  XBTH18
XBT.USD.PERPETUAL ->  XBTUSD
XBT.JPY.MONTHLY ->  XBJZ17
XBT.JPY.QUARTERLY ->  XBJZ17
ETH.BTC.MONTHLY ->  ETHZ17
ETH.BTC.QUARTERLY ->  ETHH18
ETC.BTC.WEEKLY ->  ETC7D
LTC.BTC.MONTHLY ->  LTCZ17
LTC.BTC.QUARTERLY ->  LTCH18

Would make sense to create the above as a  BiMap (https://google.github.io/guava/releases/19.0/api/docs/com/google/common/collect/BiMap.html)  create by parsing the response from https://www.bitmex.com/api/v1/instrument/active and https://www.bitmex.com/api/v1/instrument/activeIntervals,

douggie:
[...]

PR #2236, opened 2018-01-30
Fix for old guava-collections include from recent bitmex merge

traviscollins:
Moved guava-collections from xchange-core to xchange-bitmex (the only place it is used) and upgraded it to the latest version of guava. But I suggest that this dependency be removed, as it's a rather large one.


traviscollins:
This is in response to issue #2063

Issue #2177, opened 2018-01-16
[GDAX] Stop Orders


Ailanz:
Hi, Thanks for the amazing work, but I can't figure out how to place stop orders with GDAX. I only see Limit and Market. Can someone please assist? Thanks!

Flemingjp:
It appears that placing a stop order is not currently possible with the GDAX implementation, only market and limit orders are. It will require a pull request to add this feature

Ailanz:
Oh Thank you very much for the quick response

Flemingjp:
@alianz if you want I could try add it in. May take  day or two given my current free time, or lack there of

Ailanz:
@Flemingjp  I would greatly appreciate it if you could do that. Thanks so much!

badgerwithagun:
I've just spotted that getOrder/getOrders doesn't currently return stop orders.  Will add a PR shorly.

badgerwithagun:
#2283 is the main bit that was missing I think.  The other being that the limit price for stop/limit orders isn't being returned at the moment.  Not done anything about that yet.

PR #2239, opened 2018-01-30
[core-all] Implemented StopOrder and added placeStopOrder to TradeService.

theo1985:
[core-all] Implemented StopOrder and added placeStopOrder to TradeService.  
[gdax] Implemented placeStopOrder
[gdax] added (optional) Sandbox support

extends #2234 and #2177


timmolter:
I'll look at this in a few days hopefully....

PR #2245, opened 2018-02-01
Add maven wrapper.

scionaltera:
Similar to the gradle wrapper, the maven wrapper allows developers
to run maven without having to install anything on their systems.
If you run `./mvnw` instead of `mvn` the wrapper will attempt to
download the correct version of maven if necessary and run it.

The project can specify the version of maven it wants and is
significantly easier to set up for CI/CD and for developers (like me)
who have many maven based projects that require different 
versions of the tool.

Please see the project's home page here:
https://github.com/takari/takari-maven-plugin


timmolter:
I'm rejecting this for now as the gains don't outweigh the extra baggage. Couldn't you just run `/path/to/your/maven/version clean install` on your CI server?

scionaltera:
Sure you can. The wrapper just inverts control so the project can specify which version it wants to be built with, and eliminates the dependency on software installed locally. In a way it turns maven itself into a managed dependency. The CI server or the developer runs `./mvnw clean install` and the project knows how to get the right version and run it without depending on what's already installed.

If you have 50 projects on your CI server that all want different versions of maven, the wrapper is a good way to manage that without having to manually install all those different versions and constantly switch between them. If you have a hundred developers contributing to your project and you want to make sure they can all build successfully, the wrapper ensures they all use the same version of maven to build.

If you want to switch to a different version of maven due to a bug or a new feature, all you have to do is edit `.mvn/wrapper/maven-wrapper.properties` and change the URL in there. Now everybody builds with the version you changed to without having to do anything.

It's really just a convenience I've gotten used to, so I figured I'd suggest it. At work I've got more than a hundred projects checked out dating from various times over the past several years. There have been a lot of versions of maven over that time but I haven't had to manually install any of them. Each project just knows what version it wants. The same command builds each project regardless of the version, and it's one less moving part that can break over time.

PR #2248, opened 2018-02-03
Fixing some issues with getOrder on GDAX

badgerwithagun:
Specifically:

* A division by zero if an order has not been filled
* Some `BigDecimal.equals()` misuses
* Wasn't handling the pending state
* Wasn't handling the settled state
* Added `OrderStatus.UNKNOWN` for where we can't really tell what status the order is in.


xchange-gdax/src/main/java/org/knowm/xchange/gdax/GDAXAdapters.java
```diff
@@ -232,23 +237,29 @@ public static Order adaptOrder(GDAXOrder order) {
...
+    }
+
+    if(order.getStatus().equals("done") || order.getStatus().equals("settled")) {
 
       if(order.getDoneReason().equals("filled"))
         return OrderStatus.FILLED;
 
-      return null;
-
+      return null; // FIXME could be STOPPED perhaps. GDAX documentation doesn't really cover the possible
```
badgerwithagun: Log an issue?

xchange-gdax/src/main/java/org/knowm/xchange/gdax/GDAXAdapters.java
```diff
@@ -232,23 +237,29 @@ public static Order adaptOrder(GDAXOrder order) {
...
+    if(order.getFilledSize().compareTo(BigDecimal.ZERO) == 0)
       return OrderStatus.NEW;
 
-    if(order.getFilledSize().compareTo(new BigDecimal(0.0)) > 0
+    if(order.getFilledSize().compareTo(BigDecimal.ZERO) > 0
             && order.getSize().compareTo(order.getFilledSize()) < 0)
       return OrderStatus.PARTIALLY_FILLED;
 
-    return null;
+    return null; // FIXME probably need a better response than this.
```
badgerwithagun: Suggestions?

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXTradeService.java
```diff
@@ -95,7 +97,21 @@ public TradeHistoryParams createTradeHistoryParams() {
...
+        final GDAXOrder order = super.getOrder(orderId);
+        orders.add(GDAXAdapters.adaptOrder(order));
+      } catch (final ExchangeException e) {
+        // We get 404 for cancelled orders. This is a bit of a pain....
+        if ( e.getCause() instanceof HttpStatusExceptionSupport && ((HttpStatusExceptionSupport)e.getCause()).getHttpStatusCode() == 404) {
+          // We can't actually tell, if we get a 404, whether it was a market or limit order, or
+          // whether it was a bid or an ask and so on. This makes things very tricky.  Let's
+          // put a take in the ground and say you get get neither, and lots of nulls. At least
+          // the caller gets a consistent behaviour they can code to.
+          orders.add(new Order(null, null, null, orderId, null, null, null, Order.OrderStatus.CANCELED) { });
```
badgerwithagun: Faintly horrified by this TBH, but from the caller perspective it actually felt _least_ unpleasant.  Perhaps replace some nulls with zeros, add an `UNKNOWN` `OrderType`... not sure.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/GDAXAdapters.java
timmolter: maybe we should have an UNKNOWN status or similar.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXBaseService.java
```diff
@@ -15,7 +15,7 @@
 
 public class GDAXBaseService extends BaseExchangeService implements BaseService {
 
-  protected final GDAX gdax;
+  private final GDAX gdax;
```
timmolter: Doing it this way diverges from the way all other modules are implemented in XChange, so I'd rather not see this here and the related gdax() method.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXTradeService.java
timmolter: The ExchangeException should be caught and handled by the client code. That's how every module is implemented.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/GDAXAdapters.java
badgerwithagun: Yeah, that was my thinking. Happy to add that if you're comfortable with the API impact.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXBaseService.java
badgerwithagun: OK, since we don't need the 404 handling code, I don't need this.  It was only here to allow test mocking.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXTradeService.java
badgerwithagun: Fair enough. I guess it just means client code supporting GDAX will have to just know that exceptions can be a valid response.

badgerwithagun:
OK, added an UNKNOWN order status, rebased to pull some overlapping changes and removed the unnecessary changes.

PR #2250, opened 2018-02-04
nulls in bad places

jnorthrup:
using kotlin and gson (and various other pojo tools) creates  conditions where nulls exist in places they didnt before.  this fix is not a cure for the problem but it reduces the occasion where something is computed and then the accessor returns a value where the pojo still stores a NULL and emits nulls when using bean pattern code.


timmolter:
Can you resubmit this without @scionaltera's commits?

PR #2256, opened 2018-02-05
bitmex swagger Generated API

jnorthrup:
I have active operations with kucoin, and have no issues with odrotleff progress on wallet usecase.

i did not roll back my kucoin merge before tackling bitmex, but I would say it's ready for others to beat on, it WFM. 

I have attempted Bitmex wallet usecases and there is a stackoverflow recursive call in getAccountInfo and most methods are still stubs.  i found a way to move forward.

I found the swagger.json on bitmex's site and generated the api therein

the bitmex codebase is incomplete, but sort of works with XChange.   the swagger code IS complete, but this exchange is obviously a ferrari and the XChange controls are for a prius with no common design or porting done yet.  

if there's a bounty, I could be encouraged to look into this further.  

Thanks


timmolter:
I'm not sure if I'm a fan of all the swagger generated code. 264 new files? Since Kucoin and bitmex are already implemented, why is this better? Sorry for the Swagger ignorance.

jnorthrup:
im normalizing, i can split out the jkucoin and revise.  the docs is included in the patch request, the generated artifacts are one per api, but small.  if i could find a swagger->mmazi solution i would re-issue this patch generating thru that. 

the tradeoffs for the generated file s(and docs) is that these are code-generation, and the technical debt is low, if the api changes, the codebase will not be far behind the republished swagger IDL.



the existing bitmex stopped my progress because:
a) use base64 which is unspecified anywhere, and conflicts with all 8+  other ecxhanges i store api keys for causing failure on load
b) never implemented accountService which was what prompted me to find an implementation strategy (calling accountInfo method causes stack overflow!)
c) the impl's in place are hand built for some personal usecases that have no resemblance to anywhere else in the codebase, especially the bimap wierdness. 
noone has ever implemented the signature header (until yesterday?) so im not sure what level of implementation or rigor was applied.
d) numerous brittle inventions (guava, bimaps anyone?) focussed on futures.

jnorthrup:
im detecting some new conflicts i hadnt yet seen before.  
will revisit.

jnorthrup:
"placeStopOrder" has been problematic for getting this merge in.  it builds/tests fine if i dont merge upstream.

jnorthrup:
this is the third or fourth time I have stepped away from a working patch set and returned to find conflicts.  the bottom line is that the bitmex client is vastlky more complex than what is implemented in the bitmex attempts to date, and also comes with a code generation option.  it appears there are 3 concurrent attempts to solve the same problems, access to the online api's.  

i've got no time to keep patching conflicts.  i will disable bitmex from my codebase if the features don't work.

PR #2257, opened 2018-02-05
BitMEX client

ryanfox:
This PR fixes a couple issues with the BitMEX adapter:
- BitMEX's API key scheme is URL-encoded.
- BitMEX allows testing via `testnet.bitmex.com`. Testnet lists different instruments than the production site, but the currency pairs were locked in from first connecting to the live site. Connecting to testnet would not update the currency pairs, and raise an exception.

### DIFF
--- a/xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/Bitmex.java
+++ b/xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/Bitmex.java
@@ -1,6 +1,7 @@
 package org.knowm.xchange.bitmex;
 
 import java.io.IOException;
+import java.util.Date;
 import java.util.List;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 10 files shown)


jnorthrup:
I have been running the following exchanges:
BitmexExchange
BinanceExchange
PoloniexExchange
YoBitExchange
HitbtcExchange
KucoinExchange

with the exception of Bitmex, they all conform to the sample code and DO NOT use base64 to encode the ApiSecret.  further, it appears that base64 has been removed from the default BaseDigest but the method paramters retain the name.

@timmolter can you weigh in on this? 

I also have had no success using the Bitmex-client where the other exchange implementations do report user wallet info successfully.

#2256 is now working to my satisfaction, with biotmex's published swagger api being wrapped where I'm using something that isn't yet implemented in bitmex.  i can't merge this PR into my own work thru github due to conflicts but I'd be happy to incroprate whatever benefits (though I fail to understand why bitmex is unique with base64 encoing).

timmolter:
@jnorthrup It's been a while since I stumbled upon the Base64-related code. IIRC we used to have a 3rd party base64 class, which was then replaced with a JDK implementation. I think some exchanges require the encoding while others don't, but I'm not 100% sure.

ryanfox:
The first update to the signing mechanism was just to prevent a crash. The auth mechanism has been reworked in 480231 (latest commit as of now) to actually conform to BitMEX's auth scheme.

jnorthrup:
Ryan if you are committing for bitmex why aren't you using the bitmex swagger API?

i just got done porting jaxrs-spec from swagger on top of the existing handwritten XChange conversions.

ryanfox:
@jnorthrup I was using the REST API to avoid an external dependency for XChange. It would be possible though.

PR #2261, opened 2018-02-07
Add support for Bitflyer's "me/getbalance" endpoint.

scionaltera:
Is it better to do this bit by bit or wait until I've got a larger change? I'm exploring using Bitflyer for a trading bot so I'll need to implement a lot of the functionality that is currently missing. I prefer small PRs but I don't want to be a nuisance. :)


timmolter:
Small PRs are better.

PR #2263, opened 2018-02-07
[dsx] fixed the "since" and "end" timestamp parameters for fetching funding records

andre77:


npomfret:
Are you sure this change is correct?  The DSX timestamps on my fundings look fine with the current implementation.  I think I remember that the DSX documentation about the timestamp format is wrong (it's unix time, not mills since the epoch).

andre77:
its not about the timestamp of the funding, but about the parameters "since" and "end" passed in the request
see the method description
https://dsx.docs.apiary.io/#reference/reports-and-history/transfers-history/get-transfers-request

"since" - number - Show transfers at or after this timestamp (UNIX time milliseconds). If this value is not null order will become "ASC"

"end" - number - Show trades at or before this timestamp (UNIX time milliseconds). If this value is not null order will become "ASC"

npomfret:
Right.  But that's not what this PR does.  You've changed [one line of one file](https://github.com/timmolter/XChange/pull/2263/files), the effect of which is to (incorrectly?) turn a date from unix time to epoch mills.  No?

None of the changes you describe have been committed to the PR.

And the PR name (_fixed the timestamp of funding records_) is confusing if the intent is to implement more parameters in the transfer history call.

andre77:
puh, i changed the implementation of the method "nullSafeUnixTime", which now converts the time to UNIX time millis. The former implementation was converting the time to UNIX time seconds.

the private method is used exactly two times, [see lines 98 and 99](https://github.com/timmolter/XChange/blob/develop/xchange-dsx/src/main/java/org/knowm/xchange/dsx/service/DSXAccountService.java#L98)
there are parameters processed, like i already described

> None of the changes you describe have been committed to the PR.
> And the PR name (fixed the timestamp of funding records) is confusing if the intent is to implement more parameters in the transfer history call

the name of PR was changed hours ago, it was a bit misleading, yes
have u actually seen the code?

npomfret:
Ok, I understand.  But the current name is still confusing.  The FundingRecord object has a timestamp and you have not changed it.  You have changed the time unit that is passed into the call to `getDSXTransHistory` from unix time to mills...  If this is your intent I think should rename the method because it is no longer returning unix time.

andre77:
>  If this is your intent I think should rename the method because it is no longer returning unix time.
its returning "UNIX time milliseconds"
you can have "UNIX time seconds" and "UNIX time milliseconds"

andre77:
i simply removed the method "nullSafeUnixTime"
hope its ok now
PR will be renamed

PR #2265, opened 2018-02-07
[GDAX] added /users/self/verify endpoint for authenticating websocket connections

ww3456:


makarid:
Hello, can we use this method to connect with Websocket instead of REST API?

ww3456:
The XChange project is for REST APIs only, if you are interested in websocket connections have a look at https://github.com/bitrich-info/xchange-stream. 

This return value of this method is intended to be used for setting up an authenticated GDAX websocket connection, but this project does not contain the actual implementation for doing this.

makarid:
Thanks for the reply!

PR #2293, opened 2018-02-12
[LakeBTC] Fixing ticker information based on real world data

bryantharris:
The documentation on LakeBTCs website doesn't seem to match real world data.  Their documentation matches the XChange implementation, but then fails when running live against the servers.

The changes I see are as follows:
1) I eliminated the object LakeBTCTickers and replaced it with a Map object.  The example in their documentation shows USD and CNY.  The LakeBTCTickers object was hardcoded to expect this output.

In reality a large number of currency tickers are returned so a Map seemed a more natural fit and would be more future proof.

2) The structure of the keys themselves seems to have changed.  Instead of 'USD' we see 'btcusd'.  CNY doesn't seem to be returned anymore, I'd guess based off the data they've switched to HKD (returned as 'btchkd').

These changes at least get the ticker information coming in again.

One change I'd propose.  There's a bit of an impedance mismatch between the generic MarketData API and the LakeBTCMarketData API.

MarketData provides a per-CurrencyPair API.  LakeBTC returns all currency pairs at once.  This means when using the MarketData API you'll be making multiple calls to LakeBTC per currency pair even though all can be retrieved in a single server call.

I toyed around with caching the results and clearing them after a couple of seconds, but didn't want to get too far ahead of myself.  If that approach is liked I'm happy to include it.


npomfret:
I think the lakebtc implementation here is one version behind what's live

bryantharris:
I pulled upstream, that may or may not have been a problem.

The compilation problem was due to my unfamiliarity with the examples subproject.

PR #2296, opened 2018-02-13
[Cex IO] Adding a remote init to get the correct minimumAmounts

brilliantnut:
This irked me a bit - cex.io actually has a fairly varied set of limits, and the included json was static data, so needed to fix.

### DIFF
--- a/xchange-cexio/src/test/java/org/knowm/xchange/cexio/service/marketdata/RemoteInitTest.java
+++ b/xchange-cexio/src/test/java/org/knowm/xchange/cexio/service/marketdata/RemoteInitTest.java
@@ -1,0 +1,49 @@
+package org.knowm.xchange.cexio.service.marketdata;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Test;
+import org.knowm.xchange.Exchange;
+import org.knowm.xchange.ExchangeFactory;
+import org.knowm.xchange.cexio.CexIOExchange;
+import org.knowm.xchange.cexio.dto.marketdata.CexIOCurrencyLimits;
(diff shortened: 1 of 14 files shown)


xchange-cexio/src/test/java/org/knowm/xchange/cexio/service/marketdata/RemoteInitTest.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Test
+  public void integrationTest() throws IOException {
+    final Exchange exchange = ExchangeFactory.INSTANCE.createExchange(CexIOExchange.class);
+
+    assertThat(exchange).isNotNull();
+    assertThat(exchange.getExchangeMetaData().getCurrencyPairs()).isNotEmpty();
+
+    final CexIOCurrencyLimits currencyLimits = ((CexIOMarketDataService) exchange.getMarketDataService()).getCurrencyLimits();
```
timmolter: this class makes a remote call to an exchange right? Please rename it to *Integration.java so that it's run as an integration test and not a unit test.

https://github.com/timmolter/XChange/wiki/New-Implementation-Best-Practices#step-9

brilliantnut: I was wondering if I should separate the unit & Integration tests into separate classes for this reason. What are your thoughts?

brilliantnut: Addressed.

brilliantnut:
@timmolter Any thoughts?

PR #2313, opened 2018-02-15
Merge travis branch

rafalkrupinski:
Keep the configuration in a single place.

Also should make travis builds a bit faster


rafalkrupinski:
travis branch can be removed

Issue #2308, opened 2018-02-14
Withdrawal of ripple from BitFinex not working.


herofe:
There is a list of available withdrawal types not kept up to date with the BitFinex API documentation in  class BitfinexUtils called convertToBitfinexWithdrawalType.

If you try to use a currency not in the list you will get an BitfinexException("Cannot determine withdrawal type.") thrown.

The current API documentation states;
withdraw_type
string
required
can be one of the following ['bitcoin', 'litecoin', 'ethereum', 'ethereumc', 'mastercoin', 'zcash', 'monero', 'wire', 'dash', 'ripple', 'eos', 'neo', 'aventus', 'qtum', 'eidoo']

And the current implementation have not included the last six currencies in this list.

I will check with bitFinex support if any more currencies are supported in the api and update the list in the convert function.

PR #2315, opened 2018-02-15
[bitfinex] Adding withdrawal types missing from current API doc.

herofe:
Fix for #2308


herofe:
Some of the added currencies do not exist in org.knowm.xchange.currency, should I add them there also?

timmolter:
yes.

Issue #2260, opened 2018-02-06
[50 mBTC Bounty] Implement Bibox


cymp:
Hello,

I am paying a 50mBTC bounty to whomever fully implements the [Bibox API](https://github.com/Biboxcom) before Februrary 13rd. 35mBTC otherwise until I close the issue.

Top priority implementations required:
- Place/cancel orders
- Get order books
- Get user balances
- Get user active orders
- Get tickers
- Get symbols (available markets)

Thanks.

Bounties I already paid:  [HitBTC](https://github.com/timmolter/XChange/issues/1633), [YObit](https://github.com/timmolter/XChange/issues/1653), [Livecoin](https://github.com/timmolter/XChange/issues/1730), [Liqui](https://github.com/timmolter/XChange/issues/1825), [Bit-Z](https://github.com/timmolter/XChange/issues/2049), [Coinegg](https://github.com/timmolter/XChange/issues/2049)

odrotleff:
I'll have a look

odrotleff:
Current progress in my fork. Testing would be greatly appreciated :)

odrotleff:
now done with 
- Place/cancel orders
- Get order books
- Get user balances

I might be able to do the rest of your list by this time tomorrow.

odrotleff:
didnt quite finish, but now also done:
- Get user active orders
- Get tickers

odrotleff:
Withdrawal seems impossible to do. It requires the Google Authenticator code :-1:

odrotleff:
Will implement the funding history this evening. Is there anything else you expect for the bounty?

cymp:
Have you implemented the getSymbols? If so, I think this is perfect. I am away for the rest of the week (sorry that was unexpected) but I will test all this as soon as I come back. Please give me an address for the full 50mbtc bounty.

odrotleff:
I dont really find a getSymbols call in their API, XChange also doesnt require such call (I'm guessing you mean a list of all currencies on the exchange). The AccountService.getAccountInfo call returns all currencies with your balance.
I could also fake it by just taking all unique currencies from the available pairs? Do you just need all symbols or what information do you need for the symbols?

odrotleff:
address is 3Q7vE3H3F8bZ7zywP2RpDzAkC8cVF5e7Um
<3

cymp:
I meant getting all the markets, all the currency pairs available for trading.

odrotleff:
yeah thats in. :)

[...]

PR #2327, opened 2018-02-19
[Bibox] retrieve all order books

odrotleff:
refers to comment in #2260


odrotleff:
hey how do I restart CI? pretty sure this was just a random error.

PR #2335, opened 2018-02-21
idex, bitcoin.co.id, and stocks.exchange

jnorthrup:
…case)

 * Idex
 * Bitcoin.co.id
 * Stocks.Exchange
 * updated bitmex with new auth

SmokeTests presently fail for Stocks.Exchange it appears cloudflare is
dominating even the sample code they provide, they have been notified.

these share the same pom, but the custom Exchange wrapper is kotlin
which compiles to java classes anyways.  (my external codebase is
kotlin, apoligies, it doesn't prevent interop)


jnorthrup:
progress 2/3 against https://github.com/timmolter/XChange/issues/2322 

need further discussion about signing ethereum trades and direct wallet -- this is outside of the typical xchange modules scope.

timmolter:
I have many issues with this PR, mainly:

1. It should be Java
1. There are way too many new dependencies whereas zero new dependencies should need to be introduced
1. The [best practices](https://github.com/timmolter/XChange/wiki/New-Implementation-Best-Practices) should be more closely followed

jnorthrup:
I understand your reservations, you host a project which i use as a base platformn for a kotlin codebase and I have contributed 3 exchange implementations back to the community here.  

the implementation strategy is respectful to the development process at the level of keeping maintenance low and corresponding to the build and the jvm.  i have a bounty client of this PR who is satisfied with the results, and when i promised it was painless to open the pom with intellij, [the customer] did in fact have a painless experience. at this point there are two users of the fork including the bounty issuer to get IDEX and I have no problems porting the swagger to other platforms whereas I am locked into using the other XChange implementations that lack an IDL skeloton.  for this reason i suggest you keep an open mind wrt to alternative implementation strategies.  

I have another exchange to contribute soon, and ongoing, using this methodology, and hope you can look at the bigger picture, 
 * you're defining web API clients in a space where java is a tiny fraction of the implementations, 
 * the XChange-core does not utilize interfaces and in many cases has final classes.  
 * plenty of annotation magic to come up to speed on, JAXRS yes, but still spring-like.
 * swagger dominates web and REST api promulgation,
 * oracle has landlocked java so badly with patents that google has adopted kotlin as the language moving forward for android.
 * java 9 is release, but java 8 is still the python 2.7 of the jvm, because of modules, which affects XChange as well.

at any rate 

I made the modules distinct the pom level artifact names, they sort to the bottom intentionally.  i suggest you allow the developers of change modules to utilize proto, swagger, WSDL, and other tools that make software API's more repeatable, and if there was a codegenerator for MAZI i'd spend significant cycles making that work.  

what's the resolution you suggest?

PR #2336, opened 2018-02-21
[itbit, coinmate] fixes

npomfret:
[itbit] merges multiple fills for a single order into a single UserTrade and uses the order id to represent the trade id (because itbit doesn't have the concept of a trade id).

[itbit] translates XBT to BTC for the itbit response data.

[itbit] Added some checks in `getTradeHistory` to avoid potential invalid ClassCastExceptions  

[coinmate] more statuses for funding


npomfret:
This PR merges multiple fills for a single order into a single UserTrade and uses the order id to represent the trade id (because itbit doesn't have the concept of a trade id).

PR #2348, opened 2018-02-23
[itbit] fix trade tid to matchNumber

devinrsmith:
Via https://api.itbit.com/docs#market-data-get-recent-trades-get:

```json
{
    count: 3,
    recentTrades: [
        {
            timestamp: "2015-05-22T17:45:34.7570000Z",
            matchNumber: 27475,
            price: "351.45000000",
            amount: "0.00010000"
        },
        {
            timestamp: "2015-05-22T17:01:08.4270000Z",
            matchNumber: 27474,
            price: "352.00000000",
            amount: "0.00010000"
        },
        {
            timestamp: "2015-05-22T17:01:04.8630000Z",
            matchNumber: 27473,
            price: "351.45000000",
            amount: "0.00010000"
        }
    ]
}
```

### DIFF
--- a/xchange-itbit/src/main/java/org/knowm/xchange/itbit/v1/ItBitAdapters.java
+++ b/xchange-itbit/src/main/java/org/knowm/xchange/itbit/v1/ItBitAdapters.java
@@ -102,15 +102,15 @@ public final class ItBitAdapters {
   public static Trades adaptTrades(ItBitTrades trades, CurrencyPair currencyPair) throws InvalidFormatException {
 
     List<Trade> tradesList = new ArrayList<>(trades.getCount());
-    long lastTradeId = 0;
+    long lastMatchNumber = 0;
     for (int i = 0; i < trades.getCount(); i++) {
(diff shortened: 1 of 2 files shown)


npomfret:
I'm not sure it's an id, just an increasing number.  I guess it could be used as an ID, but for what purpose?

devinrsmith:
I'm not sure what the distinction between an increasing number and an ID is. The field can be used to limit the results of future calls to the trades endpoint (`https://api.itbit.com/v1/markets/XBTUSD/trades?since=27472`). The `ItBitAdapters` seems to assume that this field exists, and passes it along... and it seems reasonable to pass along the match number to the clients so it can be used on the next call to `getItBitTrades`...

If it also makes sense to change the name of the field from `tid` to `matchNumber` (and corresponding getter), that seems reasonable.

timmolter:
Chnaging the field name and getters/setters makes sense to me.

devinrsmith:
Updated the field / getter to `matchNumber`.

PR #2340, opened 2018-02-22
[bitfinex] a better remoteInit

brilliantnut:

PR #2349, opened 2018-02-23
[ANXPRO, bitfinex] multiple changes

brilliantnut:
…nteger

the amount_int parameter actually gets created by multiplying the exact amount with 1000000000, so flows out of range in int, for large number of XRP.


brilliantnut:
Also fixes issues reported in #2340

PR #2354, opened 2018-02-27
[bitfinex] Adding bitcoin cash as withdrawal type.

herofe:
Got confirmation from developers via customer support on what withdrawal type to use for bch.


brilliantnut:
Yaay for undocumented API parameters!

In #2332, I literally found BTG = bgold through trial-and-error.

PR #2314, opened 2018-02-15
Test with java 9

rafalkrupinski:
Add jdk9 run to travis.

Fixes #2337

PR #2356, opened 2018-02-28
[ANX, binance] multiple changes

brilliantnut:
### [ANX] now populates fee in FundingRecord
Essentially, the walletHistory from ANX provides separate records for trades, deposits, withdrawals and fees. Fees can be matched to appropriate transactions using the transaction id.

### [binance] now has minNotional in currencyMetadata
binance has a minNotional order size filter, which blocks orders, even if they fall between the min and max order quantity. Adding a way to get this for using a CurrencyMetadata subclass.

### DIFF
--- a/xchange-anx/pom.xml
+++ b/xchange-anx/pom.xml
@@ -30,5 +30,18 @@
             <version>4.3.4-SNAPSHOT</version>
         </dependency>
 
+        <!-- Test dependencies -->
+        <dependency>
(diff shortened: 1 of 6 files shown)


brilliantnut:
@timmolter I meant to add the second change as a different PR, but it got merged into the same.

Based on my last comment in #2314, I'm not removing PowerMockito from this right now. Please review and provide feedback if any.

PR #2359, opened 2018-02-28
gdax - historical candles support

jan-x-marek:
Hi Tim,

I just started playing with your library, and it looks great, but I miss there one substantial feature - access to aggregated historical data (open, high,low, close, volume). Not every exchange provides this service, but some do.

I am going to work with GDAX, so I integrated historical rates access on GDAX. To start off, I did not modify the core API, as it would have a huge impact. I just added a new method to GDAXMarketDataService, and it can be used like this:

		Exchange gdax = ExchangeFactory.INSTANCE.createExchange(GDAXExchange.class.getName());
		GDAXMarketDataService mds = (GDAXMarketDataService) gdax.getMarketDataService();
		GDAXCandle[] candles = mds.getGDAXHistoricalCandles(CurrencyPair.BTC_USD, "2018-02-01T00:00:00Z", "2018-02-01T00:10:00Z", "60");
		System.out.println(Arrays.toString(candles));

I believe I am not the only one who can benefit from the feature.

Best Regards

Jan


timmolter:
Thanks for this. There is an open issue somewhere I think about adding this to the generic level API of XChange, which will hopefully be implemented soon.

PR #2366, opened 2018-03-02
Ticker bidSize askSize

devinrsmith:
Some exchanges provide the size in addition to the price for ticker calls. The size information should be passed along when available. Implemented the additional fields for itBit.


timmolter:
what exactly do those fields represent?

aleks-azen:
@timmolter on the exchanges that send this data, the represent the lot size of the best bid and best ask

devinrsmith:
`bidSize` represents the instantaneous size at price `bid`
`askSize` represents the instantaneous size at price `ask`

timmolter:
it'd be nice if the description of the fields were added as Javadoc comments.

devinrsmith:
I've updated the constructor javadoc to be more explicit. If there is a more appropriate place for the javadoc, let me know!

PR #2367, opened 2018-03-03
[OkCoin] Improved get orders + fixed implementations throwing HTTP 429

cymp:


cymp:
commit 78b0c51 uses ratelimit, as the already existing implementation made multiple api calls without respecting public api calls limitations. I fixed this by introducing a custom limiter, not perfect but still better than the original non working implementation (HTTP 429). Any suggestion or further fix is appreciated.

alvins82:
Some improvements I can see -
- The ratelimiter looks to only affect tradeservice - would be better to affect all calls to the exchange.
- The implementation will block the current thread which might not be so good. Possibly better way to do it would be to use futures? Does make the implementation more complex with executors/threading etc.
- Following the above - a larger architectural question as to where this should logic should be handled. E.g. could leave current code as is and introduce some sort of wrapper which handles all the executors, futures etc. That way developer is able to work at the level where they can leave rate-limiting to the library or implement themselves.

cymp:
Thanks for your feedback.
I chose to remove the multi api calls implementations. I didn't want firstly to do that because some users might be using them, but as they don't work they have no reason to be still there. I replaced them with what the exchange can support:
- cancelOrder(CancelOrderParams orderParams), needing an id and a market name
- getOpenOrders(OpenOrdersParams params), needing exactly one market name 
- cancelOrder(String orderId) and getOpenOrders() throw a NotAvailableFromExchangeException

xchange-okcoin/pom.xml
```diff
@@ -8,7 +8,7 @@
         <version>4.3.4-SNAPSHOT</version>
     </parent>
 
-    <artifactId>xchange-okcoin</artifactId>
+    <artifactId>xchange-okcoin_custom</artifactId>
```
timmolter: is this on purpose?

cymp:
No it is not, sorry I should not have committed this file...

PR #2377, opened 2018-03-06
Added fields to Trade class

aleks-azen:
Added fields to Trade class


aleks-azen:
This is for expanding the XChange-stream repo in the future. Having only tradeIds with no orderIds streaming in is useless

timmolter:
Maybe xchange-stream should use `UserTrade`.

aleks-azen:
You're right, should have dug around more

PR #2378, opened 2018-03-06
[core] Added OrderBook sorted constructors

cymp:
we can now specify if we want to sort bids and asks at orderbook creation

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
@@ -33,6 +33,7 @@ public final class OrderBook implements Serializable {
    */
   private final List<LimitOrder> bids;
 
+
   /**
    * Constructor
    *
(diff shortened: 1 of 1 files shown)


rafalkrupinski:
I have a feeling it wont work ;)

cymp:
@timmolter Feel free to close this if you don't like it as I have other pull requests to make and I don't want to batch them into this one.

timmolter:
@rafalkrupinski why do you feel it won't work?
@cymp I'll address this ASAP. You could always create separate feature branches for parallel PRs.

npomfret:
Instead of this constructor I think a better approach would be to add a static factory method, something like `createWithSortedOrders(...)`

cymp:
@npomfret fair point, but I don't like too much mixing public constructors and static factory methods, and it is obviously out of question to privatize the OrderBook constructor.

npomfret:
Ok, but factory methods are a good alternative to having multiple (and easily confusable) constructors.   You can't name constructors so it's often not clear that they might be doing something other than constructing (assigning parameters to fields).  Personally, I would never make a constructor do anything other than check the validity of the parameters and then assign them to fields - I'd always use a factory method if those parameters needed some pre-processing before they can be assigned.  Plus the stack trace you get if something goes wrong is a tiny bit clearer in a factory method.

cymp:
Yeah sure you don't need to convince me here, I would have prefered having factory methods too, but only them.

I just think designing a class with public constructor + factory methods a bit odd, I'm not used to have both since it mixes two designs. I mean if anyone starts looking at how to instanciate an OrderBook, he will type new OrderBook(... in his IDE and since there is a public constructor, he won't instinctively look at the factory methods and fully understand the class.

Thanks for giving your opinion.

rafalkrupinski:
@timmolter I thought XChange is java7 compatible. My Bad, removing the comment

timmolter:
I think the best way for XChange to construct objects is the builder pattern, if I had to pick one out of the many ways possible. We have implemented builders for quite a lot of the generic DTOs, but some haven't yet been converted.

rafalkrupinski:
@timmolter why just one design pattern for the whole project?

timmolter:
personal preference, I guess.

PR #2379, opened 2018-03-06
Corrections + add candlestick for some exchanges

sebastienkoch:
The code is now more secure, and doesn't fail when values are null.

I've also added timeout or changed the rate limit when necessary.

I've also deleted unavailable currency pairs.

The diff is showing too many lines. Sorry for that. Could someone help me to avoid this in the future?


timmolter:
Any idea why the diff views show almost everything as being different? Are you using a different carriage return encoding or tabs perhaps?

sebastienkoch:
It may effectively be because of a carriage return encoding issue. But it's strange that some lines appear correctly, and some others don't, even within the same file (for exemple in [xchange-bitfinex/src/main/resources/bitfinex.json](https://github.com/timmolter/XChange/pull/2379/commits/b3baf919215f3ee782a42ace3331c5241dddac51#diff-679bf1abfe4316ec4a6880ee9f226d09 "xchange-bitfinex/src/main/resources/bitfinex.json")).  
  
I'll try to find a better solution in the future.

timmolter:
Are you on Windows? What IDE are you using? You can switch the return encoding like this: http://jonathonstaff.com/blog/issues-with-line-endings/

See our code style page: https://github.com/timmolter/XChange/wiki/Code-Style

sebastienkoch:
I'll restart this from scratch.

PR #2392, opened 2018-03-09
bitfinex] implemented XMR (monero) withdrawals

npomfret:


xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/service/BitfinexAccountService.java
```diff
@@ -82,17 +83,26 @@ public String withdrawFunds(Currency currency, BigDecimal amount, String address
...
   public String withdrawFunds(WithdrawFundsParams params) throws IOException {
     if (params instanceof RippleWithdrawFundsParams) {
-      RippleWithdrawFundsParams rippleWithdrawFundsParams = (RippleWithdrawFundsParams) params;
-      return withdrawFunds(rippleWithdrawFundsParams.currency, rippleWithdrawFundsParams.amount, rippleWithdrawFundsParams.address,
-          rippleWithdrawFundsParams.tag);
-    }
-    if (params instanceof DefaultWithdrawFundsParams) {
+      RippleWithdrawFundsParams xrpParams = (RippleWithdrawFundsParams) params;
+      return withdrawFunds(
+          xrpParams.currency,
```
timmolter: using getters and setters would be preferable to be consistent with the rest of the lib.

npomfret: Done

PR #2404, opened 2018-03-13
[Huobi] Add new exchange

SergeyLukashevich:
Add previously removed exchange using its new API.


xchange-huobi/src/main/resources/logback.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
timmolter: This file should be removed.

PR #2407, opened 2018-03-13
[OKCoin/OKEX] support for batch cancel

cymp:
Added support for cancelling up to 3 orders at the same time.


cymp:
@timmolter please accept this asap since it fixes a build error caused by OrderBook.java.

PR #2415, opened 2018-03-16
(Bug) QuoineAdapters.class

makarid:
Hello, i believe that i have corrected a bug that i found on QuoineAdapter class.The think is that with the previous code the Adapter could take only a currencyPair that has 3char on base currency and 3 char on counter currency.If there was a currencyPair like this FLIXX/ETH ,the Adapter returned FLIX/XXE as a currencyPair.If you find something wrong please tell me.Thanks


npomfret:
Does this work for all scenarios (trades, funding etc)?

PR #2419, opened 2018-03-16
I believe i have corrected a Bug on QuoineAdapter

makarid:
Please see the correction and tell me if something is wrong.I have check it and i works for me.Thanks


makarid:
I am sorry but what i have done wrong?

PR #2425, opened 2018-03-18
[dsx] metadata - renamed BCC to BCH (DO NOT MERGE YET)

npomfret:


npomfret:
@timmolter am waiting for the exchange to update one of the API endpoints (https://dsx.uk/mapi/info) that has the old names it it still

PR #2427, opened 2018-03-20
[dsx] deal with the metadata being out of sync with the market name for bitcoin cash

npomfret:


walec51:
do we solve this problem on the library level?
BitBay also uses BCC for Bitcoin Cash instead of BCH and I see not correction of this type in its code

npomfret:
No it's just a thing for DSX.  They are mid way through renaming their bitcoin cash symbol from bcc to bch

PR #2432, opened 2018-03-23
[xchange-core] make xchange-core less prone to accidental refactors p…

jnorthrup:
…er  #2431

when any module has /x/y.json file and it is of a value to refactor or clone the package and specialize the package names, intellij will cause great havoc refactoring without tedious measures taken

PR #2443, opened 2018-03-27
disambiguate the test resources into packages

jnorthrup:
continues  #2432 to disambiguate the test resources into packages toward "improved" practices which allow better merged assemblies, shading, and uberjar for deployments and commandline runners.


jnorthrup:
@timmolter this doesn't appear to be a hotly debated issue, and the outstanding pr's are at low point.  

if all the test resources are equally in their (typically dto) packages, and all the code is equally formatted before commits, and the new .gitattributes solves for mixed line-endings,  then the source tree should weather concurrent modification over time much better, and enable better assembly/shading, refactoring, and incremental compile/test options in IDE's

any objections?

timmolter:
What role does `.gitattributes` have in this exactly? I don't understand.

I like the concept of having consistent code formatters via Maven, Eclipse and intelliJ, and the [fmt-maven-plugin](https://github.com/coveo/fmt-maven-plugin) looks like a nice plugin that ties that all together. A couple questions:

1. Will having this plugin in the pom enforce that all PRs follow the correct style?
1. Does the plugin handle the line endings too?

What "quieter failsafe plugin" are you referring to? I don't see it.

jnorthrup:
re: formatter(S): 

 * the formatter (google style or aosp style) is left on "verbose", maybe that's what turned me off about it so i made it skipped by default.  
   * can be put in the "release" profile to run when upgrading version number all at once.  i didnt think of this in time.
   * maybe turning off verbose is enough to tolerate formatting on every build.

 * the formatter in the comments is better, it is driven by JDT configs, but JDT is broken upstream and it doesnt work is says "module closed".  the workarounds didnt fix that, so for another day.  it can use the configs you have now for eclipse.  except its broke.

re: lin-endings 
/.gitattributes/  changes the line-endings on commit - so this will minimize whitespace conflicts.  i noticed there isn't a single line-ending in the tree. I borrowed that .gittributes file from someone who spent more time studying it than me, defaults to LF except in intellij xml.  seems to me .net succeeds more often when xml is crlf also...

if one programmer just sets this on commandline, it doesn't help.  i think i have windows and ubuntu git competing for line-endings on my setup.  so .gitattributes is another project-wide global config like .gitignore.  

re: integration tests
there is a failsae plugin config in comments [here](https://github.com/jnorthrup/XChange/blob/75da408ad04b272c02dea08599f471950240c78f/pom.xml#L448).  files in src/main/test-integration are ignored in IDE's by default, so you can more easily do continuous junit (a good thing) as you modify and refactor code.  

i did not move the integration tests or change the plugin, but i have a script to do it all at once in polyglot branch.  so it's here in the comments in case that's of interest to anyone else.

pom.xml
```diff
@@ -86,6 +87,9 @@
     </ciManagement>
 
     <modules>
+<!--shouldn't this always be first?-->
```
timmolter: It doesn't matter. Maven figures out it needs to first compile core.

timmolter:
Thanks.

What's `etc/util/mass-checkin-refactor.sh`?

So this PR ONLY contains:

1. fully-qualified test package names
1. The `fmt-maven-plugin` plugin that can be called upon to format the code according to the `google` format
1. Reformatted code using the `fmt-maven-plugin` plugin.

This PR DOES NOT contain

1. line ending changes
1. integration tests moved

timmolter:
I'm going to accept the new code formatting changes after a couple of days leaving this PR open unless there is massive backlash against it here. I like using a standardized code format (google) which has synchronized IntelliJ and Eclipse formatters. I once tried making synced formatters, but always failed at it.

jnorthrup:
What's etc/util/mass-checkin-refactor.sh?

when you asked for the one big change to be broken out, i needed to write a script to perform the breakout of before and after per module.  its checked in for the value of transparency and if others have ideas about global changes, the tool is there for now to mitigate the scope of conflicts

pom.xml
jnorthrup: this is new information for me.  it's definitely been a source of pain in the past

jnorthrup:
i think 30/70 include line-ending changes, there were more java files using CRLF and more JSON using LF before.   no module was safe.  on github" ?w=1" hides ws but they have never made any UI hints for it

timmolter:
I just tried using google code formatter templates in IntelliJ and Eclipse and they both fail when trying to import them with invalid format errors. I also tried the google-formatter plugin in IntelliJ, and while it works, there were formatting differences between that and what the `fmt-maven-plugin` plugin produces. So in practice, this is unfortunately a failed concept. Too bad, really.

timmolter:
OK, the formatter files actually work. I was downloading them incorrectly: https://github.com/google/styleguide/issues/338#issuecomment-377208923

However, all 3 formatting methods produce different results.

jnorthrup:
the hope is that if the formatter runs prior to checkin, say in "package", then the deltas will be smaller and clearer to changes despite devs who are compulsive or reckless (guilty here) with reformat/reordering.  

the nascent benefits are 
  * code generation and scripts are more predictable with sed, grep, etc.
  * vim users are hard headed and dgaf about your standards...
  * choice of IDE goes way down on the list of entry barriers
  * line-endings stop hiding meaningful changes and rewriting whole files

sorry for the configs failure, i don't think i have invested time in editor configs for an age  given the number of linux installs and laptops i go through in a given year

timmolter:
After some experimentation, this maven plugin is compatible with IntelliJ and Eclipse if you setup both IDEs with the correct plugins as opposed to the formatter files. 

Here's my preferred plan in merging your changes.

1. Close this PR
1. Open a new issue regarding the new formatter proposal, seek consensus
1. format the code myself and push the changes
1. @jnorthrup shows me the method to fix the line endings and I push those changes myself after playing around with it
1. @jnorthrup creates a PR with the package name refactor in the test package (Do you have a script for that?)
1. @jnorthrup creates a PR with the integration test package

jnorthrup:
I'll follow your lead and fill in the blanks.

jnorthrup:
2 releases later, I'm still asking to disambiguate the test sources

timmolter:
What's the `reformat.skip` in pom.ml for again?

timmolter:
Regarding moving of the integration tests, can you list out the reasons this is necessary/advantageous? I don't think being ignored by an IDE is an advantage actually, and would cause more effort to run them from an IDE. Right now, I'm not convinced moving the integration tests brings any advantage.

jnorthrup:
there is still a posted bounty on refactoring the metadata.  this will involve some drastically propogated changes into the source tree.  as is typical, refactoring isn't incremental, but tends to make things collapse, and disappear, and remove assumptions and dogmatic practices.  

my personal opinion is that the XChange style guide promotes a unified build practice for all modules and therefore these are just packages and not differentiated tools -- this is more or less proven in my private branch, they behave like packages not modules, until you start to bring in code generation and other things.  as it stands now the sum of changes to all poms is maybe 3 or 4 total java deps above and beyond what xchange-core and rescu provide. 

for swagger API code generation and kotlin stuff, the maven provides adequate tool differentiation per module and the plugins are appropriate to break out into seperate modules.

the challenges or "special treatment" when bringing legit legal jaxrs bindings into the XChange/Rescu convention is currently made harder than it needs to be by the use of final classes and a lack of interface based software contracts all the way up the line.  if a binding throws Exception, rescu refuses to generate the proxy.  "features" like this are hard to work around and change the requirements of a supported API binding from a simple job of writing adapters and wrapper classes to a more brittle solution including generated code that is hand maintained.

I'd propose the use of okhttp and gson code that swagger uses if only because it looks equally reliable and well maintained on all fronts -- the swagger generator is ancient, okhttp is from square, and ancient, and gson is from google, and ancient.  for these tools stackoverflow has comprehensive solutions, swagger.io has closed down its wiki and referred all support reqs to SO and github at this point. 

i have personally had no regrets choosing gson over jackson whenever there is a fork in the road because of ease of prototyping, gson tends to accept the json object model as mostly maps, where jackson tends to enforce the java object model and may reject missing or added elements and halt working software if upstream changes.  swagger generates pure jaxrs with jackson json or generates full implementations with okhttp, or rx-java.  so far there is no eaasy path from generated bindings into XChange due to all the final classes and safeguards against whatever the safeguards are there for.

jnorthrup:
constant, neverending "not on my watch" responses. 

https://github.com/timmolter/XChange/wiki/Bulk-POJO-Generation shows something very similar to swagger, but swagger also generates the jaxrs, or the well maintained rest services proxy code outright, in almost a dozen flavors.

i have a swagger spec which embeds json-schema, and then the generator assigns pojos to request and response specifications.  in my case, i had about 5 working generated swagger modules in 2 weeks.

>I don't support a refactor using Swagger, gson, okhttp, etc. either because it's too major of a change and _we like things how they are_. Everything works fine as is.



however, as I try to find a middle ground from _we like things how they are_ and " i had about 5 working generated swagger modules in 2 weeks.", I have gone two routes to solve for forward progress:

a) (the defensive option)  built a project fork that allows me freedom from  _we like things how they are_ 
shared as the polglot brnach in my fork, and 

 b) (contributing back to the community, altruism) opening pr's and issues to have discussions about technical debt, and methods of achieving additions to XChange in ways that we can sign off and move on to more interesting labors.   

@mmazi has done a good job of cultivating the exchange modules at a certain point in time to a refactored module  

image

so maybe im tooting the horn of API's you are not familiar with and don't have energy to vet, but if the tests pass, and more importantly, the community can use the modules and fix them as they need to, why would you impose your personal viewpoint on what should and should not be allowed in the tree?  you seem to argue for footprint, and minimize gap from what exists, but these methods are baroque, manually cvultivating each dto? even if there's an IDL defining them all in one place? 

perhaps you can perform reviews and request documentation expansion instead of boolean edicts.
#####

I'm porting legal JAXRS to the rescu JAXRS provider and the jaxrs provider factory breaks on "throws Exception".  why should 60+ modules all depend on a single, aging, one-off module that has outlived anyone who cares about the rationales present?  that probably worked well for a handful of modules and developers, years ago, but well, I'm wasting my time to argue for the metrics of productivity here.  This is your commercialized product, and the github presence seems incidental to garner free additions, but in no way to foster a community.

cheers.

PR #2446, opened 2018-03-27
Add getTickers to MarketDataService and implement Kraken

caleb-allen:
This extends #2388 and implements `getTickers` for Kraken

### DIFF
--- a/xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenMarketDataService.java
+++ b/xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenMarketDataService.java
@@ -1,6 +1,9 @@
 package org.knowm.xchange.kraken.service;
 
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenMarketDataService.java
```diff
@@ -31,6 +37,16 @@ public Ticker getTicker(CurrencyPair currencyPair, Object... args) throws IOExce
     return KrakenAdapters.adaptTicker(getKrakenTicker(currencyPair), currencyPair);
   }
 
+  @Override
+  public List<Ticker> getTickers(Params params, Object... args) throws IOException {
```
timmolter: I don't think `args` is necessary now with the `Params` argument.

caleb-allen:
@timmolter agreed. Done

PR #2448, opened 2018-03-28
Unified MathContext

walec51:
I think we should unify our MathContext for all BigDecimal operations where it makes sens and you would not want to see more then 8 digits after the decimal point. There is no point in constructing a new MathContext object for each operation.


walec51:
if this change is of with you all then I'll expand it to make use of MathUtils.DEFAULT_CONTEXT in every BigDecimal operation which didn't specify it explicitly.

jnorthrup:
eth has 18 decimal places.

timmolter:
@walec51 Just create a reference `MathContext` in the class it is used. It will probably be different across the many exchanges, but consistent for a given exchange.

PR #2455, opened 2018-03-31
Merge pull request #3 from timmolter/develop

cacofishadct:
sync


timmolter:
can you explain this PR? Seems weird.

cacofishadct:
I suppose you are referring to the changing limit to String part?
from https://www.kraken.com/help/api. 
limit = maximum net amount that can be deposited right now, or false if no limit
The old code will throw error since it may return false and failed to parse as BigDecimal.

cacofishadct:
Was trying to sync originally in the PR, but later added more changes to the code.

xchange-kraken/src/main/java/org/knowm/xchange/kraken/dto/account/KrakenDepositAddress.java
```diff
@@ -5,13 +5,16 @@
...
   private final Long expiretm;
   private final boolean newAddress;
 
-  public KrakenDepositAddress(@JsonProperty("address") String address, @JsonProperty("expiretm") Long expiretm,
+  public KrakenDepositAddress(@JsonProperty("address") String address,
+      @JsonProperty("tag") String tag, @JsonProperty("expiretm") Long expiretm,
       @JsonProperty("new") boolean newAddress) {
     super();
     this.address = address;
+    this.tag = tag;
```
timmolter: I don't see the tag field in the API docs: https://www.kraken.com/help/api#private-user-funding

timmolter:
OK, I see that limit can be false now, according to their API docs.

xchange-kraken/src/main/java/org/knowm/xchange/kraken/dto/account/KrakenDepositAddress.java
cacofishadct: True. It's not in the doc.  I tested out....

timmolter: Are you saying `tag` is in the response based on your experience?

cacofishadct: Yes, I tried 2 coins which need tag/memo, and tag is always returned in that case.

PR #2461, opened 2018-03-31
Bug fix for KucoinAdapter.adaptUserTrade

michaelscheung:
Converting KucoinDealtOrder to UserTrade, orderOid is the orderId, oid is the id. Fixes https://github.com/timmolter/XChange/issues/2457


nikos:
Yep, this is an exact duplicate of #2459 :-)

PR #2463, opened 2018-04-03
[kucoin] Allow to specify (optional) timespan on trade history params.

nikos:
Actually a followup PR to support querying trade history with specific start- and end-time (see #2453).


nikos:
@timmolter Wow, this was amazingly fast! Thanks a lot, Tim!

PR #2469, opened 2018-04-05
update bitmex futures pairs

boppleton:
some expired contracts still remain, i don't want to remove them yet and risk breaking someones code.


boppleton:
@jnorthrup absolutely.  i would be happy to implement that!

PR #2473, opened 2018-04-05
[Binance] Add support for placing limit order with TimeInForce flags.

MoarDm:
Add TimeInForce FOK enum (in case of [official api](https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#enum-definitions))
Add posibility to place limit order with any TimeInForce flag


golart:
+

PR #2479, opened 2018-04-07
Removed param not supported by Bittrex any more

walec51:


timmolter:
Can you look into the class that failed compilation during the CI build?

walec51:
fixed

PR #2486, opened 2018-04-09
First endpoint from new Bitbay API

walec51:
Its still not officially published but its the only option if you want to download any users full trades history.
Current official API only gives you past 3-4 months of data.

### DIFF
--- a/xchange-bitbay/src/main/java/org/knowm/xchange/bitbay/v3/BitbayAdapters.java
+++ b/xchange-bitbay/src/main/java/org/knowm/xchange/bitbay/v3/BitbayAdapters.java
@@ -1,0 +1,50 @@
+package org.knowm.xchange.bitbay.v3;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import org.knowm.xchange.bitbay.v3.dto.trade.BitbayUserTrade;
+import org.knowm.xchange.bitbay.v3.dto.trade.BitbayUserTrades;
+import org.knowm.xchange.currency.Currency;
+import org.knowm.xchange.currency.CurrencyPair;
+import org.knowm.xchange.dto.Order.OrderType;
+import org.knowm.xchange.dto.marketdata.Trades;
+import org.knowm.xchange.dto.trade.UserTrade;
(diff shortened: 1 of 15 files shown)


npomfret:
@walec51 excellent work, thanks for this.

npomfret:
@walec51 did you consider using `TradeHistoryParamsPaging` instead of using `TradeHistoryParamNextPageCursor` together with `TradeHistoryParamLimit`?

walec51:
The pageCursor parameter is NOT a number, you cannot set it to zero or iterate it and expect it to work.
The cursor value for the first page is 'start', for the next page you always receive an string in UserTrades that somewhat looks like an UUID. Therefore this type of paggination system is totally different from the one user TradeHistoryParamLimit where your just iterate 0,1,2,3,... I've seen the pagination system used in BitBays new API in NoSQL solutions like Cassandra so it think they migrated to one for historical data. On a technical level its way more efficient for large datasets then a typical numeric limit / offset solution or page numbers. 

should I implement an example or an unit test on how to use it?

npomfret:
No that's ok I understand what you mean.

What is confusing though, is the response I'm getting if I supply no parameters:
```
{
  "status": "Ok",
  "totalRows": "25",
  "items": [
    ... 10 items here
  ],
  "query": {
    "markets": [
      []
    ],
    "limit": [],
    "offset": [],
    "fromTime": [],
    "toTime": [],
    "initializedBy": [],
    "rateFrom": [],
    "rateTo": [],
    "userAction": [],
    "nextPageCursor": []
  },
  "nextPageCursor": null
```

npomfret:
No that's ok, I understand.  I'm going to add a specific params object that makes it more obvious to use the 'start' parameter.

Where did you find this information?  I'm looking [here](https://docs.bitbay.net/v3.0.0/docs/historia-transakcji) but can't see the paging instructions.

walec51:
I've go it from here:
https://github.com/stasbar/tax-ledger-cli/blob/master/src/main/kotlin/com/stasbar/taxledger/exchanges/bitbaynew/BitBayApi.kt

the author of this project was talking with BitBays support that he could not download all trades using offset / limit or fromTime / toTime so they gave hit info about this undocumented param

Issue #2395, opened 2018-03-10
abucoins error "INVALID SIGN"


npomfret:
I'm seemingly randomly getting a "INVALID SIGN" error from abucoins. Anyone else seeing this?

cymp:
I don't get such error.

makarid:
Do you mean Invalid hash signature?

npomfret:
I guess that's what the error means.  The exact text of the error from the API is "INVALID SIGN".

npomfret:
Maybe you have a threading issue or something like that.  You can't make concurrent request with the same api key for a lot of exchanges. I only get this error on abucoins so I don't think that's the issue in my code (but it's the only idea i have right now).

makarid:
Thanks a lot for your reply @npomfret . I didn't know about concurrent issue.Thanks

npomfret:
I can reproduce the issue now.  I use multiple API keys.  When I make a request with one, and then another request with another I get this error.  No idea why though.

npomfret:
Fixed in https://github.com/knowm/XChange/pull/2488

PR #2488, opened 2018-04-10
[abucoins, bitbay]

npomfret:


npomfret:
Fixes https://github.com/knowm/XChange/issues/2395

PR #2489, opened 2018-04-10
[bitbay] v3 api work

npomfret:


timmolter:
Nice usage of the default keyword!

PR #2501, opened 2018-04-15
Fixed NPE and index out of bounds exceptions in Livecoin

walec51:
There ware two problems with unpopular markets in Livecoins module:

- if not trades happened recently an empty list was return which resulted in an index out of bound exception in the adapter
- if no bids or asks are on a given market live coin returns a nulled field instead of an empty list which resulted in a NPE in the adapter

Besides the above I removed currencies and pairs that are not supported on that exchange any more.

Why do we maintain such large JSON metadata in the library for exchanges that have proper remoteInit methods?

I think we should either limit then to just a few currencies/pairs that have no chance of being delisted or remove those JSON files entirely.


timmolter:
The metadata json files are there for the case when no one yet implemented a `remoteInit` and/or when the exchange doesn't have an endpoint for certain aspects of the metadata. If the `remoteInit` covers some aspect of the metadata, then that aspect can be deleted in the json file, IMHO.

Issue #2506, opened 2018-04-16
metadata for currencies/instruments doesn't scale in source code.


jnorthrup:
instead of pushing a new app update when currencies change or options expire, something a little cleaner should be found.  

case in point, 
![image](https://user-images.githubusercontent.com/73514/38829138-61a38c10-41e2-11e8-8371-85381426bed3.png)

180 currency listings are defined in code but never referenced.  this is because...  code smell..

Issue #2495, opened 2018-04-12
relax finals, introduce interfaces


jnorthrup:
certain aspects of the code could be easier to adapt to API's if they were interfaces.  

every single instance where there is a public final class could have a more functional interface alternative with no impact to the existing running code given refactoring tools that rename X with XImpl and serve the same interface extracted.  now that line endings and formatting are automatic, the delta should be clean and obvious.

I'd like to submit Accountinfo, Wallet, Currency, and CurrencyPair as interfaces with the existing Builders that provide the existing public final classes as they exist now (only where applicable-- some classes don't have builders, and some classes do have interfaces, this is confusing also.)

I'd even like to nominate someone else to perform the intellij refactorings, like @timmolter, who has a mind for making things more consistent.

jnorthrup:
artist's conception of an adapter that's no more stateful than absolutely necessary:

```java

  
    public interface AccountInfo {
        /**
         * Gets all wallets in this account
         */
        Map<String, Wallet> getWallets();

        /**
         * Gets wallet for accounts which don't use multiple wallets with ids
         */
        Wallet getWallet();

        /**
         * Gets the wallet with a specific id
         */
        Wallet getWallet(String id);

        /**
         * @return The user name
         */
        String getUsername();

        /**
         * Returns the current trading fee
         *
         * @return The trading fee
         */
        BigDecimal getTradingFee();
    }

    public interface Wallet {
        /** A unique identifier for this wallet 
         * @return The wallet id
         */
        String getId();

        /** A descriptive name for this wallet. Defaults to {@link #id} 
         * @return A descriptive name for the wallet
         */
        String getName();

        /**
         * Returns the balance for the specified currency.
         *
         * @param currency a {@link Currency}.
         * @return the balance of the specified currency, or a zero balance if currency not present
         */
        Balance getBalance(Currency currency);
    }

    public interface Balance extends Comparable<org.knowm.xchange.dto.account.Balance> {
        Currency getCurrency();

        /**
         * Returns the total amount of the <code>currency</code> in this balance.
         *
         * @return the total amount.
         */
        BigDecimal getTotal();

        /**
         * Returns the amount of the <code>currency</code> in this balance that is available to trade.
         *
         * @return the amount that is available to trade.
         */
        BigDecimal getAvailable();

        /**
         * Returns the amount of the <code>currency</code> in this balance that may be withdrawn. Equal to
         * <code>available - borrowed</code>.
         *
         * @return the amount that is available to withdraw.
[...]

Issue #2342, opened 2018-02-23
Let's reconsider lombok


brilliantnut:
I saw the old discussion on the #2008 and #2066 around removing the lombok dependency, and do understand your reluctance to include a number of 3rd party libraries, but I want to make the case for lombok again.

In my recent java development, lombok has become so common, that I almost consider it a part of the language. It ends up reducing a huge amount of boiler-plate code, and also reduces maintenance overhead - e.g. for toString(), equals, hashcode and constructors.

In most of my other projects, my DTOs look like this:

```java
import lombok.Data

@Data
public class DepositAddress {
  private final String address;
  private final String destinationTag;
}
```

and that's it. `@Data` gives you an RequiredArgs Constructor, Getters, Setters (for non-final fields), equals, hashcode, toString. If you end up adding, removing or moving around fields, the generated toString / equals / hashcode changes automatically.

Also, most IDEs (definitely Eclipse and Idea) now have first class support for lombok, so generated getters / setters & constructors show up correctly while coding in other classes.

bryantharris:
I'm a big fan of Lombok as well.  It also shouldn't be a run time requirement, just a build time requirement.  Also some one time fiddling with your IDE to recognize.

As with most approaches that remove boiler plate, it's tough to go back once you accept it as part of your development toolset.

jnorthrup:
I think boilerplate removal is important as well.  i have utilized kotlin with swagger-codegen to minimize the maintenance footprint in a particular PR, but this is the same basic thing.

rafalkrupinski:
I've only seen 1 abandoned project using lombok. Requires setting up IDE and java cmdline tools fo every developer and CI environment.

jnorthrup:
java as a language has peaked.  you can be grateful when a project is not overrun with Spring and dependency injection as those  tend to derail your ability to look at the code and understand what's happening at runtime because of the isolated implementation code bound from reflection and assemblies.  my personal favorite lombok-like thing was google autogen which does the exact same thing.  

https://www.youtube.com/watch?v=Q_Z_8HXt68k is an excellent take on the evolution of jvm options and what's emerging pragmatically

brilliantnut:
@rafalkrupinski wrote:
> I've only seen 1 abandoned project using lombok. Requires setting up IDE and java cmdline tools fo every developer and CI environment.

[...]

PR #2507, opened 2018-04-16
Balance,Currency,CurrencyPair,Wallet,AccountInfo as Interface

jnorthrup:
testing against conversation in #2495, also moves bitmex metadata into bitmex module since i was refactoring currency.   ( #2506 )

there was no code modified to speak of, this didn't require annotation hell like i expected, just need method called valueOf in the interface that matched the existing instanceOf method, so that was renamed.  now jackson reads the interface as a factory 

@see https://github.com/FasterXML/jackson-databind/wiki/Mapper-Features for ValueOf documentation to provide deserialization to Interfaces


jnorthrup:
@timmolter I just wanted to comment here, I haven't editted that much code here, but the previous formatter delta doesn't seem to come from the maven based google-formatter which is the source of all this commit noise.  

when intellij opened any files to refactor from Currency class to Currency interface it did prune imports, that can't be helped.  I have resolved the conflicts, yet again, thanks for your attention.

Jim

timmolter:
No worries. The 3 formatter tools (Eclipse, IntelliJ, Maven) are not 100% matched, related to the imports.

jnorthrup:
@walec51 please be specific
with this patch set no contracts have changed., 
no semantics have changed
no imports have changed
no serialization has changed
no sideffects have changed.
the changes are to constrictors and the option (no code was changed) to write code against pure interfaces.

jnorthrup:
@timmolter I think @walec51 brings up a good point, there is no roadmap and no major and minor designation to any changes.  there is no RFC for next version on the table, which seems like it creates the opportunity for FUD where refactoring is concerned

timmolter:
OK, I'm going to accept this unless anyone speaks up with objections. I'll bump the version to 4.4.0 afterwards. I know it's not ideal, but that's the way it is. @jnorthrup I'm going to merge some other PR in the mean time so you might have to fix some stuff on your branch - sorry about that!

cymp:
@timmolter @jnorthrup I have been away for a while and I see a lot of stuff refactored, are we going to have some release notes, or some backwards compatibility issues we would need to fix, or any guideline about what to do? Or maybe no issue at all?

cymp:
sorry I have just noticed that  @walec51 expressed the same concerns.

timmolter:
Oh, because the interface and impl are both called `Currency`....

walec51:
reviewing this PR is almost unbearable... improving formatting on all of the files should not be mixed with such a refactor

walec51:
Could the author explain to me how a interface with tons of code in default and static methods is better then a class? its not an interface in the classical sense - what can you do better with this that couldn't be done in the current implementation?

jnorthrup:
@walec51 there's no effort spent formatting, that belongs to the maven plugin.  

>Could the author explain to me how a interface with tons of code in default and static methods is better then a class? its not an interface in the classical sense - what can you do better with this that couldn't be done in the current implementation?

final classes make bad contracts.  the evidence is how the number of constructors could be 0 - pure interfaces with mostly simple functional code, but is in actuallty closer to one per usecase, and like Balance, has hidden bugs because the abstraction is doing things a DTO should never do

jnorthrup:
@timmolter the choice to leave those long package named Currency imports was just to spare the code any more touch than was necessary after intellij had gotten done doing 2-4 step permutations to arrive back at the same import graph.  I would have inflated the delta to files what were unrelated in order to reverse intellij's long form.   over time, like line-endings and automated formatting, these ought to normalize.  nothing formal there.

the implementation uses the new package but the code should still retain the same declaration and not exhibit the .impl package anywhere that im aware of, as a goal.

the intent was to spare the consumers any awareness of the new package by perserving creation in the same class and using a factory method.  also not to go with an interface closer to an SPI (e.g. impl package) instead of DTO's with CurrencyImpl classes.

i noticed you did not accept the conflict resolution and merge and bump the version.  other priotiries i guess.  for this reason I DID NOT invest any effort in making a pure functional example of modules using the new interfaces.

walec51:
I know maven did that but that does not change the fact that this should be split into two PR

classes ware not final and there is little difference between a simple factory method and a constructor

simple factory methods actually force you to do final classes because you have no usable constructors to use in classes that cloud inherit them

we could debate design philosophy for days but the question still remains - what functionality can you implement easier with those interface what you couldn't do in the current implementation?

is there an example of how beneficial this change is in those 439 changed files?
because in the first few files I only see that now we have to use full names `org.knowm.xchange.currency.CurrencyPair` which is a pain, not progress

jnorthrup:
@timmolter I do spend a lot of time trying to preserve existing functionality and minimize the alice in wonderland nature of old growth features here, please go ahead and fix whatever conflicts you create if this  work is a positive contribution, otherwise, just reiterate "it is what it is" and close the PR

walec51:
so personal, so angry... accusing me of being a slop and proclaiming yourself to be a design guru is not a technical argument

you didn't answer my questions and I'm not against consistency - I'm only evaluating if changing those classes to interfaces is the way to go

give me a good example of a CurrencyPair interface having more then one implementation and I'll agree that changing it to an interface was a good idea

you know, there is a reason why value objects like BigDecimal or OffsetDateTime ware NOT done as interfaces...

jnorthrup:
@walec51 
>you didn't answer my questions and I'm not against consistency - I'm only evaluating if changing those classes to interfaces is the way to go

https://github.com/knowm/XChange/issues/2495#issuecomment-380947706

I tried.   ... 9 days ago.  the common option to take an interface, and create a flyweight and on-stack implementation in pure functional code is what Interfaces enable.  

the amount of CurrencyPair unit test code is non-zero, and non-trivial. the amount of json serialization code is higher with Currencypair than the others i touched.   

if this PR does go in, I'm pretty sure my next new module will benefit from a much cheaper ExchangeMetadata instantiation where the loop variables are wrapped in an interface inside a stream().map() method call and forgotten ever a after.

walec51:
yes I did look it up 5 days ago

tell me, when creating an instance of an anonymous class that uses some objects don't those objects live in memory for as long as that instance you created?

so this will use up more memory compared to mapping what you need to a dto, because in this scenario all those dependent objects get deconstructed afterwards

ergo the example in getAccountInfo has nothing to do with the flyweight pattern
and is consuming more memory then mapping to simple value objects or DTO

jnorthrup:
https://www.beyondjava.net/blog/escape-analysis-java/ this one seems good.

jnorthrup:
" ergo the example in getAccountInfo has nothing to do with the flyweight
pattern
and is consuming more memory then mapping to simple value objects or DTO"

you're trolling now.  how's one vtable more expensive than another vtable?
particularly if it's on-stack?  whaat the hell is your point?  you appear
to be invoking the corncob antipattern...  look that one up too

walec51:
and what do you think escape analysis does to object instances used in an inner class?
deconstructs them before they are used?

I don't see how escape analysis changes what I've said. Lets say I have an app that stores order books from all exchanges in memory to do traversal trade analysis on them - I accrual do have such an app! DTOs will take less space in memory and be faster then objects from inner classes with references to some other objects which they have to map on the fly.

timmolter:
Thanks for the discussion. I'm happy someone else is taking a close look at this PR. 

>  instead of 5+ constructors it seems like a better defensive code practice to have one big constructor and use builders.

Many core DTOs have evolved over time to be inconsistent, and I think it would be great to have one constructor and a builder for all DTOs across the lib. That's one thing I like about this PR. 

The refactor with DTO interfaces seems a bit odd at first, but I can also admit that there are people out there that know much more about API design and the new functional programming API that came with Java 8. In order to keep XChange modern, it does seem like it should start taking advantage of those features and APIs. I've been content doing things in Java 6 style, but appreciate the need to keep this lib up to date using the latest or at least non-ancient JDK. But as a full disclosure, I'm not at all an expert in that area although I'm happy to learn. 

The PR format from @jnorthrup could be a lot better I must say - more atomic with better explanations as to why the changes are better, that everyone can understand. I'm not saying to break this PR up, but it's something to keep in mind for the future. I do appreciate your desire to improve things. Please do keep the discussion purely technical, otherwise any hope for cooperation and understanding will be lost.

walec51:
> the corncob antipattern

as in: Corncobs are difficult people who can be prevalent in the software development business

once again you are trying to insult me instead of addressing my point

look up the https://en.wikipedia.org/wiki/Flyweight_pattern pattern your self before u suggest your using it because your not using it in the getAccountInfo example

in your example the inner class you create will have a reference to `entry` object which will have a reference to the exchange API's response which may have much more data then one needs <- those objects will live in memory as long as the instance of the anonymous inner class you've created

walec51:
I fail to see how calling methods like this:

```
                                        @Override
                                        public BigDecimal getTotal() {
                                            return getAvailable().add(getFrozen());
                                        }
    
                                        @Override
                                        public BigDecimal getFrozen() {
                                            return entry.getValue().getOnOrders();
                                        }
```

is more performant in runtime then calling methods like this:

```
                                        @Override
                                        public BigDecimal getTotal() {
                                            return total;
                                        }
    
                                        @Override
                                        public BigDecimal getFrozen() {
                                            return frozen;
                                        }
```

I have an app that does in memory analysis on object from this library add things like this may hit it. Especially if you plan to do more complex mapping / adaptations in every method call...

walec51:
you can insult me all day long but that does not give you merit
I gave you many arguments against your orthodox 'everything should be an interface' mentality 

Java inlining doesn't make this:

```
                                        @Override
                                        public BigDecimal getTotal() {
                                            return getAvailable().add(getFrozen());
                                        }
```

equal to this

```
                                        @Override
                                        public BigDecimal getTotal() {
                                            return total;
                                        }
```

in bytecode, notice the add operation which you will do for every method call

A flyweight is an object that minimizes memory usage by sharing as much data as possible with other similar object. In your example if two calls to the exchanges API return the same data you will not save any space. You will have two instances of you anonymous inner class each with a reference to a independent response DTO instance. Therefore this is not a flyweight by any means.

jnorthrup:
@walec51 can you just stop and be pragmatic?  

I didn't write the Balance class, and I didn't change the Balance class methods except to break out and segregate the willywonka inference code into its own declared methods.  The DTO should never have been given method code but that wasn't on my watch.  Balances are a ledger.  if an Exchange tells you a given tuple of immutable amounts, why the hell are we inferring values in that code?  this is immaterial, I don't benefit from knowing the answer.

you should run "git blame Balance.java", if you want to know who made those decisions.

jnorthrup:
" A flyweight is an object that minimizes memory usage by sharing as much data as possible with other similar object. In your example"

my example shows an interface, something with no data, only a vtable, and stack objects pulled into the inner class methods.  i fail to see what's not flyweight about that.

walec51:
the point is:
- more complex mappings will not get inlined in your solution
- escape analysis doesn't eliminate the referenced objects from memory, they will live as long as the instance you consider 'flyweight' lives

ps. did you event try to read the Java flyweight example on the wiki?
I'll give you a clue: the cache attribute in the BMWSerieFlyWeightFactory is the core of this pattern

jnorthrup:
clearly willing to talk about subjects you have no experience or knowledge
with before today i see.  that makes you my first choice for authoring
exchange software.

>

walec51:
insults are not arguments...

jnorthrup:
@walec51 I wrote the damn PR? what arguments should I have with it?  Did you have any others?  WHen i cite facts, you make FUD where there are only facts.  are we done yet?

jnorthrup:
@walec51 if you want syntax and brevity improvements, see Let's reconsider lombok #2342

if you want performance metrics to expose my gross negligence, show some benchmarks to prove your point on this code.  if you want to argue about design patterns, I can only say that if we both belonged to a design patterns user group, I'd probably remove myself.  

the mass-format operation PR already happened, sorry you missed that one.  I run 3 alternate project structures for XChange and I have a vested interest in central consistency and stability.  

did you want to discuss 5.x featuresets? I took your sniping and constructively understood it to be a suggestion and opened an issue for that too. 

this project has a lot of room for improvement, and despite your best attempts, and anyone else's, I will probably continue to offer my most objective take on the matter, within the given constraints.  but the thing you don't see is any issues where i JUMP IN AND RELENTLESSLY FUD THE AUTHORS

For you, I'm almost tempted to give it a whirl

jnorthrup:
@timmolter I do truly spend a lot of my _time_, contributing.  we clearly have varying levels of participant experience in the 1200 forks to date, but so far noone's turned up a trojan horse or a breaking change with this PR.  

Interfaces make programming more reliable, and simpler.  I'd like to get to work beyond the conclusion of this PR, in order to submit new modules with or without the DTO constraints.  If my opinion of FUD is as valid as the opinions of risk that have been aired in here, then the _FUD_ is a regrettable sour note, but I'd like to move beyond fixing merge conflicts that seem to occur when high touch PR's get prioritized behind small PR's

walec51:
@jnorthrup I see you have removed a lot of your comments that contained insults towards me
that's good by now my comments look out of context

jnorthrup:
@walec51 your comments have been out of context with or without my responses.  The thing I won't be doing is validating any of your claims about the hotspot, you look stupid commenting far beyond your depth.

walec51:
I've raised concerns that the targeted approach to lazy mapping / adapters you want to pursue may hurt performance in scenarios which cache and frequently access object constructed by this library - your counterarguments end up being personal, not technical.

You have an orthodox bias that everything should be an interface - which I find to be childish and an anti-pattern, especially in the case of value objects:

https://martinfowler.com/bliki/ValueObject.html - judging from your comments your not familiar with the concept

https://martinfowler.com/bliki/InterfaceImplementationPair.html
https://www.symphonious.net/2011/06/18/the-single-implementation-fallacy/
https://mortoray.com/2014/08/01/the-false-abstraction-antipattern/

therefore I do not consider it as self-proving argument as you do.

jnorthrup:
@walec51 once again, FUD, and intellectual violence.

if you have some meaningful benchmarks, first you have to get some numbers, otherwise you've just established "I am familiar with A", and "B is not A"

jnorthrup:
at first you found 4 links, now its 6 links.  I have perused them all.

Your arguments make a lot of sweeping assumptions, lacking the context of the PR, and illuminating problems which this PR is not introducing.  

further, you're taking a declarative style and pruporting to have knowledge of performance hazards, backed up with ... technical debt articles. the technical debt from final classes manifests itself in the maintenance of Adapters in the modules, and my reluctance to go with the flow when I've got experience delivering simpler and more functional streaming code which is more fun to maintain.

the maintenance burdone in this codebase, and therefore also inflating this PR due to final classes 
is

```bash
jim@XChange$ find -iname *Adapter*java|wc -l
116
```

so.....
116 classes have the name "Adapter"

for how many DTO's?  

well we maintain packages named /dto/ 
```bash
jim@XChange$ find -iname *dto* -type d|wc -l
180
```

we have 180 source directories defining dto packages in this codebase among the 77 modules.

that's a lot of surface area to maintain.

```bash
jim@XChange$ wc -l `find -iname *Adapter*java`
   464 ./xchange-abucoins/src/main/java/org/knowm/xchange/abucoins/AbucoinsAdapters.java
    19 ./xchange-abucoins/src/test/java/org/knowm/xchange/abucoins/AbucoinsAdaptersAdaptCurrenyPairTest.java
[...]
   251 ./xchange-yobit/src/main/java/org/knowm/xchange/yobit/YoBitAdapters.java
    55 ./xchange-zaif/src/main/java/org/knowm/xchange/zaif/ZaifAdapters.java
    37 ./xchange-zaif/src/test/java/org/knowm/xchange/zaif/ZaifAdaptersTest.java
 21845 total
```

21845 lines of code 

what I'm proposing now is that I want to contribute modules that don't reflect the tradition of these maintenance trends, by enabling more functional interface based streaming in my modules, but to do so, it is more efficient and less maintenance to break away from final classes, and by the way 

what does Martin fowler say about 

some DTO's are finals
some DTO's have builders
some DTO's have factory methods
some DTO's have json usecases
NO classes use Serializable in the code but one test does to make sure that TWO SEPARATE CURRENCIES ARE EQUAL
SOME CLASSES LIKE EXCHANGE HAVE INTERFACE AND STILL REQUIRE AN ABSTRACT CLASS!!!
?

you keep coming at this with vitriolic attempts to open up a can of worms where none exist (e.g. performance degradation is not a hallmark of interfaces or inner classes or functional flyweight stream idioms, or even accessors.  vtables don't increase latencies either, sorry charley, your arguments are tedious and you dismiss the reference materials that back up hotspot vm optimizations specific to functional idioms ) 

In some cases, like Currency and CurrencyPair, there is just as much benefit to promoting a simple abstraction like a pair of strings to an interface as promoting the corresponding dependent abstractions to interfaces, this makes code generation a LOT easier, and serialization/deserialization more consistent, and lets me inherit my module DTO's from the xchange-core DTO's instead of being forced into builders and factory methods for which no parent abstraction is legal to use; thereby reducing the maintenance of yet another brittle pair of *Adapter*.java artifacts.  CurrencyPair has 6 constructors for the simplest possible Pair tuple abstraction.  if you got a beef with that then I share that with you.

i don't need to publish new packages full of adapters for xchange-core DTO's if those DTO's are interfaces, I can just instantiate them.  when I have an API client that want's to know what liberties I've taken with their published interfaces, they will see a consistent application of transforms based on software  constracts rather than a corpus of busywork adapter code that bloats billable objectives.

I'm writing this for the benefit of .. well, i doubt any benefit, there's no debate to your style of FUD here.  if i publish an article to validate inner workings you just spam links, many of which are also irrelevant.

cheers

walec51:
if any comments here can be viewed a intellectual violence then they would be those which you tried to delete

adapter code in your solution does not disappear - its just moved to anonymous inner classes

as for ditching simple DTOs i see little maintenance ease gained there

here is your 'interface and mapping in anonymous class' concept applied to OrderBooks:

```
package xchscripts;

import java.math.BigDecimal;
import java.time.Duration;
import java.time.Instant;
import java.util.Arrays;
import java.util.List;
import java.util.stream.Collectors;

public class Benchmark {
    
    static final int INTERATIONS_COUNT = 1_000_000;
    
    /**
     * Normal as in POJOs that ware prepared by an adapter
     */
    static class NormalOrder {
        
        private BigDecimal originalAmount;
        private BigDecimal price;

        public NormalOrder(BigDecimal originalAmount, BigDecimal price) {
            this.originalAmount = originalAmount;
            this.price = price;
        }

        public BigDecimal getOriginalAmount() {
            return originalAmount;
        }

        public BigDecimal getPrice() {
            return price;
        }
    }
    
    static class NormalOrderBook {
        
        private List<NormalOrder> bids;

        public NormalOrderBook(List<NormalOrder> bids) {
            this.bids = bids;
        }

        public List<NormalOrder> getBids() {
            return bids;
        }
    }
    
    interface IOrder {
        BigDecimal getPrice();
    }
    
    interface IOrderBook {
     
        List<IOrder> getBids();
    }
    
    static class ExchangeSpecificOrder {
        
        BigDecimal spend;
        BigDecimal received;

        public ExchangeSpecificOrder(BigDecimal spend, BigDecimal received) {
            this.spend = spend;
            this.received = received;
        }

        public BigDecimal getSpend() {
            return spend;
        }

        public BigDecimal getReceived() {
            return received;
        }
    }
    
    static class ExchangeSpecificOrderBook {
        
        private List<ExchangeSpecificOrder> buyOrders;

        public ExchangeSpecificOrderBook(List<ExchangeSpecificOrder> buyOrders) {
            this.buyOrders = buyOrders;
        }

        public List<ExchangeSpecificOrder> getBuyOrders() {
            return buyOrders;
        }
    }
    
    public static void main(String[] args) {
        Instant start = Instant.now();
        if (args[0].equals("normal")) {
            System.out.println("Starting normal test at " + start.toString());
            normalTest();
        } else {
            System.out.println("Starting interface test at " + start.toString());
            interfaceTest();
        }
        Instant end = Instant.now();
        System.out.println(
            "Ending test at " + end.toString() + ". Duration: "
                + Duration.between(start, end).toMillis() + " millis"
        );
    }
    
    static void normalTest() {
        NormalOrderBook orderBook = new NormalOrderBook(
            Arrays.asList(
                new NormalOrder(new BigDecimal("10"), new BigDecimal("1.4")),
                new NormalOrder(new BigDecimal("20"), new BigDecimal("1.6")),
                new NormalOrder(new BigDecimal("15"), new BigDecimal("1.5"))
            )
        );
        for (int i = 0; i < INTERATIONS_COUNT; ++i) {
            orderBook.getBids().get(1).getPrice();
        }
    }
    
    static void interfaceTest() {
        ExchangeSpecificOrderBook orderBookFromExchange = new ExchangeSpecificOrderBook(
            Arrays.asList(
                new ExchangeSpecificOrder(new BigDecimal("10"), new BigDecimal("14")),
                new ExchangeSpecificOrder(new BigDecimal("20"), new BigDecimal("32")),
                new ExchangeSpecificOrder(new BigDecimal("15"), new BigDecimal("22.5"))
            )
        );
        IOrderBook orderBook = new IOrderBook() {
            @Override
            public List<IOrder> getBids() {
                return orderBookFromExchange.getBuyOrders().stream()
                    .map(
                        order -> new IOrder() {
                            @Override
                            public BigDecimal getPrice() {
                                return order.getReceived().divide(order.getSpend());
                            }
                        }
                    )
                    .collect(Collectors.toList());
            }
        };
        for (int i = 0; i < INTERATIONS_COUNT; ++i) {
            orderBook.getBids().get(1).getPrice();
        }
    }
}
```

results:

Starting normal test at 2018-04-22T09:45:30.704Z
Ending test at 2018-04-22T09:45:30.902Z. Duration: 198 millis. Sum: 16000000.0

Starting interface test at 2018-04-22T09:45:50.142Z
Ending test at 2018-04-22T09:45:52.235Z. Duration: 2093 millis. Sum: 16000000.0

198 millis vs 2093 millis

the result would be even worse if instead of a ExchangeSpecific* DTO I would have a JSON object

walec51:
@jnorthrup wrote
> further, you're taking a declarative style and pruporting to have knowledge of performance hazards, backed up with ... technical debt articles

I was making two separate points there... don't confuse them

jnorthrup:
i was right, responding was pointless.  you hope you can persuade the readers that this PR should be rejected.

walec51:
you requested a benchmark from me and you got it...

jnorthrup:
what does Martin fowler say about

some DTO's are finals
some DTO's have builders
some DTO's have factory methods
some DTO's have json usecases
NO classes use Serializable in the code but one test does to make sure that TWO SEPARATE CURRENCIES ARE EQUAL
SOME CLASSES LIKE EXCHANGE HAVE INTERFACE AND STILL REQUIRE AN ABSTRACT CLASS!!!
?

walec51:
I agree that we should improve the consistency of this library

I agree that it shouldn't be nesesery to use an abstract class to implement an Exchange interface 

I don't agree that value objects like Currency or CurrencyPair should be interfaces
They should be flyweight, but not according the pattern in your head, but the pattern described in 
https://refactoring.guru/design-patterns/flyweight or https://en.wikipedia.org/wiki/Flyweight_pattern
with a factory method that would manage a pool of existing flyweights objects

also as my benchmark shows property access performance is bad in the target solution you are proposing

jnorthrup:
i noticed your test was pretty blatantly comparing apples to oranges so I made the two declaration styles perform the same work 
here 
https://gist.github.com/jnorthrup/e12acc4e0df968f1be2646758226c696

and the results are here
https://gist.github.com/jnorthrup/d1c5ee2c3c8fb6297da12ecb2e5b6404
 
the results show the interfaces make no difference to the implementation strategy performance at the scales you benchmark...

"C:\Program Files\Java\jdk1.8.0_161\bin\java.exe" -javaagent:G:\apps\apps\IDEA-C\ch-0\181.4445.78\lib\idea_rt.jar=60324:G:\apps\apps\IDEA-C\ch-0\181.4445.78\bin -Dfile.encoding=UTF-8 -classpath "C:\Program Files\Java\jdk1.8.0_161\jre\lib\charsets.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\deploy.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\access-bridge-64.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\cldrdata.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\dnsns.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\jaccess.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\jfxrt.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\localedata.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\nashorn.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\sunec.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\sunjce_provider.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\sunmscapi.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\sunpkcs11.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\ext\zipfs.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\javaws.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\jce.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\jfr.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\jfxswt.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\jsse.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\management-agent.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\plugin.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\resources.jar;C:\Program Files\Java\jdk1.8.0_161\jre\lib\rt.jar;I:\XChange\xchange-examples\target\test-classes;I:\XChange\xchange-examples\target\classes;C:\Users\jim\.m2\repository\ch\qos\logback\logback-classic\1.2.3\logback-classic-1.2.3.jar;C:\Users\jim\.m2\repository\ch\qos\logback\logback-core\1.2.3\logback-core-1.2.3.jar;C:\Users\jim\.m2\repository\org\knowm\xchart\xchart\3.5.1\xchart-3.5.1.jar;C:\Users\jim\.m2\repository\de\erichseifert\vectorgraphics2d\VectorGraphics2D\0.11\VectorGraphics2D-0.11.jar;C:\Users\jim\.m2\repository\org\reflections\reflections\0.9.11\reflections-0.9.11.jar;C:\Users\jim\.m2\repository\com\google\guava\guava\20.0\guava-20.0.jar;C:\Users\jim\.m2\repository\org\javassist\javassist\3.21.0-GA\javassist-3.21.0-GA.jar;I:\XChange\xchange-core\target\classes;C:\Users\jim\.m2\repository\com\github\mmazi\rescu\2.0.2\rescu-2.0.2.jar;C:\Users\jim\.m2\repository\com\fasterxml\jackson\core\jackson-databind\2.9.1\jackson-databind-2.9.1.jar;C:\Users\jim\.m2\repository\com\fasterxml\jackson\core\jackson-annotations\2.9.0\jackson-annotations-2.9.0.jar;C:\Users\jim\.m2\repository\com\fasterxml\jackson\core\jackson-core\2.9.1\jackson-core-2.9.1.jar;C:\Users\jim\.m2\repository\javax\ws\rs\javax.ws.rs-api\2.1\javax.ws.rs-api-2.1.jar;C:\Users\jim\.m2\repository\com\google\code\findbugs\jsr305\3.0.2\jsr305-3.0.2.jar;C:\Users\jim\.m2\repository\oauth\signpost\signpost-core\1.2.1.2\signpost-core-1.2.1.2.jar;C:\Users\jim\.m2\repository\commons-io\commons-io\2.6\commons-io-2.6.jar;I:\XChange\xchange-openexchangerates\target\classes;I:\XChange\xchange-wex\target\classes;C:\Users\jim\.m2\repository\org\apache\commons\commons-lang3\3.7\commons-lang3-3.7.jar;I:\XChange\xchange-bibox\target\classes;I:\XChange\xchange-binance\target\classes;I:\XChange\xchange-bitstamp\target\classes;I:\XChange\xchange-bittrex\target\classes;I:\XChange\xchange-campbx\target\classes;I:\XChange\xchange-bitcoincharts\target\classes;I:\XChange\xchange-bitcoinde\target\classes;I:\XChange\xchange-blockchain\target\classes;I:\XChange\xchange-kraken\target\classes;I:\XChange\xchange-kucoin\target\classes;I:\XChange\xchange-bitfinex\target\classes;I:\XChange\xchange-bitmex\target\classes;I:\XChange\xchange-bitcoinium\target\classes;I:\XChange\xchange-coinbase\target\classes;I:\XChange\xchange-coinegg\target\classes;I:\XChange\xchange-gdax\target\classes;I:\XChange\xchange-bitcoinaverage\target\classes;I:\XChange\xchange-cexio\target\classes;I:\XChange\xchange-anx\target\classes;I:\XChange\xchange-itbit\target\classes;I:\XChange\xchange-independentreserve\target\classes;I:\XChange\xchange-lakebtc\target\classes;I:\XChange\xchange-hitbtc\target\classes;I:\XChange\xchange-paymium\target\classes;I:\XChange\xchange-btctrade\target\classes;I:\XChange\xchange-cryptonit\target\classes;I:\XChange\xchange-poloniex\target\classes;I:\XChange\xchange-okcoin\target\classes;I:\XChange\xchange-bleutrade\target\classes;I:\XChange\xchange-mercadobitcoin\target\classes;I:\XChange\xchange-bitbay\target\classes;I:\XChange\xchange-bitso\target\classes;I:\XChange\xchange-taurus\target\classes;I:\XChange\xchange-quoine\target\classes;C:\Users\jim\.m2\repository\com\auth0\java-jwt\3.1.0\java-jwt-3.1.0.jar;C:\Users\jim\.m2\repository\commons-codec\commons-codec\1.10\commons-codec-1.10.jar;C:\Users\jim\.m2\repository\org\bouncycastle\bcprov-jdk15on\1.55\bcprov-jdk15on-1.55.jar;I:\XChange\xchange-ripple\target\classes;I:\XChange\xchange-therock\target\classes;I:\XChange\xchange-btcmarkets\target\classes;I:\XChange\xchange-gatecoin\target\classes;I:\XChange\xchange-cryptofacilities\target\classes;I:\XChange\xchange-bitmarket\target\classes;I:\XChange\xchange-coinmate\target\classes;I:\XChange\xchange-coinone\target\classes;I:\XChange\xchange-empoex\target\classes;I:\XChange\xchange-vircurex\target\classes;I:\XChange\xchange-ccex\target\classes;I:\XChange\xchange-livecoin\target\classes;I:\XChange\xchange-yobit\target\classes;I:\XChange\xchange-dsx\target\classes;I:\XChange\xchange-abucoins\target\classes;I:\XChange\xchange-cryptopia\target\classes;I:\XChange\xchange-btcturk\target\classes;I:\XChange\xchange-paribu\target\classes;I:\XChange\xchange-liqui\target\classes;I:\XChange\xchange-koineks\target\classes;I:\XChange\xchange-gateio\target\classes;I:\XChange\xchange-koinim\target\classes;I:\XChange\xchange-acx\target\classes;I:\XChange\xchange-bitflyer\target\classes;I:\XChange\xchange-bitz\target\classes;I:\XChange\xchange-zaif\target\classes;C:\Users\jim\.m2\repository\org\slf4j\slf4j-api\1.7.25\slf4j-api-1.7.25.jar;C:\Users\jim\.m2\repository\junit\junit\4.12\junit-4.12.jar;C:\Users\jim\.m2\repository\org\hamcrest\hamcrest-core\1.3\hamcrest-core-1.3.jar;C:\Users\jim\.m2\repository\org\assertj\assertj-core\3.9.1\assertj-core-3.9.1.jar" xchscripts.Benchmark
----- run 0
Starting normal test at 2018-04-22T10:47:39.926Z
Ending test at 2018-04-22T10:47:40.120Z. Duration: 194 millis. Sum: 16000000.0
Starting interface test at 2018-04-22T10:47:40.120Z
Ending test at 2018-04-22T10:47:40.231Z. Duration: 111 millis. Sum: 16000000.0
Starting normal test at 2018-04-22T10:47:40.231Z
Ending test at 2018-04-22T10:47:40.291Z. Duration: 60 millis. Sum: 16000000.0
Starting interface test at 2018-04-22T10:47:40.291Z
Ending test at 2018-04-22T10:47:40.434Z. Duration: 143 millis. Sum: 16000000.0
Starting normal test at 2018-04-22T10:47:40.434Z
Ending test at 2018-04-22T10:47:40.498Z. Duration: 64 millis. Sum: 16000000.0
Starting interface test at 2018-04-22T10:47:40.498Z
Ending test at 2018-04-22T10:47:40.554Z. Duration: 56 millis. Sum: 16000000.0


but as any one who has ever built a benchmark on java knows, you need to let the vm warm up before you can prove a point, so i added that too.  the end result looks like 


Ending test at 2018-04-22T10:47:51.262Z. Duration: 50 millis. Sum: 16000000.0
Starting normal test at 2018-04-22T10:47:51.262Z
Ending test at 2018-04-22T10:47:51.315Z. Duration: 53 millis. Sum: 16000000.0
Starting interface test at 2018-04-22T10:47:51.315Z
Ending test at 2018-04-22T10:47:51.369Z. Duration: 54 millis. Sum: 16000000.0
Starting normal test at 2018-04-22T10:47:51.370Z
Ending test at 2018-04-22T10:47:51.420Z. Duration: 50 millis. Sum: 16000000.0
Starting interface test at 2018-04-22T10:47:51.420Z
Ending test at 2018-04-22T10:47:51.472Z. Duration: 52 millis. Sum: 16000000.0
Starting normal test at 2018-04-22T10:47:51.472Z
Ending test at 2018-04-22T10:47:51.525Z. Duration: 53 millis. Sum: 16000000.0
Starting interface test at 2018-04-22T10:47:51.525Z
Ending test at 2018-04-22T10:47:51.580Z. Duration: 55 millis. Sum: 16000000.0

walec51:
you didn't run the actual code which illustrates how you want to eliminate DTOs (which you renamed to FUDNOWTest)

your interfaceTest does not do mapping of a response object returned by an API in the anonymous inner class so its no wonder it didn't show any performance differences

you removed the thing I say is bad in your design from the actual test...

walec51:
if you want to do mapping in the inner class property methods then they will always do more work then an already mapped DTO - that's the point I'm trying to show you and not comparing apples to oranges

jnorthrup:
i benchmarked the same DTO's, one which is an interface, one which is a class.  I don't know what you are implying I am attempting to do with this PR, but the benchmark sums it up. 

jdk8 supports this construct:
```java
    List<IOrder> collect =
        Arrays.asList((new BigDecimal("1.4")), (new BigDecimal("1.6")), (new BigDecimal("1.5")))
            .stream()
            .map(bigDecimal -> (IOrder) () -> bigDecimal)
            .collect(Collectors.toList());

    IOrderBook orderBook = () -> collect;
```
this is different from 
    List<NormalOrder> normalOrderStream =
        Arrays.asList((new BigDecimal("1.4")), (new BigDecimal("1.6")), (new BigDecimal("1.5")))
            .stream()
            .map(bigDecimal -> new NormalOrder(bigDecimal))
            .collect(Collectors.toList());
    NormalOrderBook orderBook = new NormalOrderBook(normalOrderStream);


you're not demonstrating the XChange source code with 6 years of unreafactored evolution, but it looks like you'd prefer to make it 7 years

walec51:
how does this have anything to do with the concept you presented earlier?!

```
  @Override
                public Wallet getWallet() {
                    Wallet wallet = null;
                    try {
                        wallet = new Wallet() {
                            Map<Currency, Balance> balances =
                                    RestProxyFactory.createProxy(ReturnCompleteBalancesApi.class, idexExchange.getExchangeSpecification().getSslUri()).completeBalances(new CompleteBalancesReq().address(apiKey)).entrySet().stream().map(entry -> new Balance() {
                                        @Override
                                        public int compareTo(@NotNull org.knowm.xchange.dto.account.Balance o) {
                                            return Comparator.comparing(Balance::getCurrency).compare(this, (Balance) o);
                                        }
    
                                        @Override
                                        public Currency getCurrency() {
                                            return Currency.getInstance(entry.getKey());
                                        }
    
                                        @Override
                                        public BigDecimal getTotal() {
                                            return getAvailable().add(getFrozen());
                                        }
    
                                        @Override
                                        public BigDecimal getAvailable() {
                                            return entry.getValue().getAvailable();
                                        }
    
                                        @Override
                                        public BigDecimal getAvailableForWithdrawal() {
                                            return getAvailable();;
                                        }
    
                                        @Override
                                        public BigDecimal getFrozen() {
                                            return entry.getValue().getOnOrders();
                                        }
                                    }).collect(Collectors.toMap(Balance::getCurrency, identity()));
    
                            public Balance getBalance(Currency currency) {
                                return balances.get(currency);
                            }
    
    
                            @Override
                            public String getId() {
                                return userName;
                            }
    
                            @Override
                            public String getName() {
                                return userName;
                            }
    
                        };
                    } catch (Exception e) {
                        e.printStackTrace();
                    }

                    return wallet;
                }
```

jnorthrup:
in the words of @timmolter 

/it is what it is/

if it is to be, it is up to @timmolter

walec51:
do you not see the obvious performance difference in:
a) getting a property already mapped one by an adapter
b) mapping a property for each getter call by a anonymous inner class

walec51:
in your test there was no "mapping in the inner class property methods", you just passed a BigDecimal to your anonymous class...

your previous comment is irrelevant but don't delete it like you did with the previous irrelevant ones

my results that actually tested your mapping concept stand
doing a JVM warm does not change the result in a significant way

Starting normal test at 2018-04-22T09:45:30.704Z
Ending test at 2018-04-22T09:45:30.902Z. Duration: 198 millis. Sum: 16000000.0

Starting interface test at 2018-04-22T09:45:50.142Z
Ending test at 2018-04-22T09:45:52.235Z. Duration: 2093 millis. Sum: 16000000.0

cymp:
@walec51 
Could you please run this again using JMH?

walec51:
yes, I'll post the JMH version tomorrow

jnorthrup:
this PR is about removing the restrictions on final dto's and using interfaces.  

@walec51 the jdk8 vm is a past release, it defines lambdas.  that's a lambda 
@cymp did you look at the two supposed benchmarks he published times for?   there appears to be two different usecases being benchmarked.  

i beleive @walec51 is under the impression i took liberty to insert computed results in place of stateful accessors.  since that's not in this PR, I really don't know what this benchmark proves, or why it matters.  if we are aruging whether interface declarations promoted to lambdas or inner classes cause meaningful slowdowns, I have tweaked the benchmark to reflect that, and i do think it is relevant to this discussion and appears in this PR.

cymp:
@jnorthrup No sorry I only got my phone atm, I just stumbled across the results of @walec51 's benchmark and was curious about the results.

walec51:
the only justification for objects like Curreny, CurrencyPair or OrderBook to be interfaces instead of value objects is the getWallet example you gave in the issue which brought us to this PR

if doing lazy mapping in anonymous inner classes instead of adapters hinders performance then we shouldn't do it ergo these is no need for value objects to be interfaces

jnorthrup:
@walec51 "my results that actually tested your mapping concept"

so far you haven't tested any of the concepts that would help me or that appear in the issue or the PR. 

 i think you are arguing for something you don't comprehend about variable scoping in my original example.  those inner variables in the Wallet interface are created only once.  

i did leave the proxy creation inline, because the proxies are cheap enough to worry about last.  it could be something in a higher scope or static, but that's not the point of the example.

im sorry if you misread the example and got the wrong impression of the intent of this PR.  I doubt that's true, I think you're more likely to keep arguing until noone has the time or energy to give this PR, and by attrition, you win.

jnorthrup:
@walec51 you seem to be most concerned about the Balance class

```java
         @Override
                                        public BigDecimal getTotal() {
                                            return getAvailable().add(getFrozen());
                                        }
```    

the idex json client API only provides available and frozen, not total.  there's a very complex, dangerous set of rules in the Balance "Class" about when null==total.  this is nothing new.  check it out for yourself.

jnorthrup:
more on method inlining -- 

i think you expressed objections to lazy value interpolation in my specific example, which bears profiling, its hard to know what the hotspot is doing in a given revision.

in a past incarnation, the hotspot did method caching which would inline methods called with immutable invariants.

since an inner class interface instantiation with no setters that uses local variables as super-scope references passed in, those values e.g. `entry` above become final and in a.add(b) a and b are both also final and so will be inlined.  and eventually, the inner class will be inlined and garbage collected, leaving only the value.

so there is an optimization tradeoff, of not declaring the "total" attribute to memoize this as explicit lazy result which has an up front cost of computing the value once or checking it for null the first time and then computing it -- the hotspot method inline profiler will actually set aside this particular loop to evaluate finals and invariants, and begin to decompose inlines as well as the method profile counters increase.  the outcome of a well managed immutable interface with getters is about as close to a static pojo as one can hope, if the parent scope values are also method parameters that appear only once.

the cost of this a.add(b) operation is insignificant, and XChange is outside the bounds of what high-frequency-transaction codebases would impose on concurrency and code standards. the benchmark certainly doesn't illustrate this edge case with the Balance class, but it's one of the things that concerns me most is that we have DTO's performing inference based on null checks.  I would trust an interface any day over the experience i once had of getting a NPE when i called total, in the past, with BCE proxy generated code.

walec51:
Yes, and an adapter would calculate this once, where your solution will do it for every call.

While this is not a big performance hit here I'm curious how a Trades or OrderBook implementation of your pattern would look like for exchanges which require non trivial mapping / adapters. If the only good that comes from treating value objects as interfaces is the ability to implement things like you did show in the getWallet example then I'm not in favor of it - period. The fact that there are a lot of inconsistencies in the library does not imply that every thing has to be an interface.

I have an app which downloads millions of trades and does intensive in memory analysis of order books. Your struggle to discredit my competence wasted enough of our time in the comments. Let the project owners decide.

jnorthrup:
@walec51 the code I have in the example is still preferable to the code that's in the balance class right now in this PR ported faithfully from the original single method:
```java
  /**
   * Returns the total amount of the <code>currency</code> in this balance.
   *
   * @return the total amount.
   */
  @Override
  public BigDecimal getTotal() {

    return total == null ? inferTotal() : total;
  }

  @Override
  public BigDecimal inferTotal() {
    return getAvailable()
        .add(getFrozen())
        .subtract(getBorrowed())
        .add(getLoaned())
        .add(getWithdrawing())
        .add(getDepositing());
  }
```

walec51:
as @jnorthrup correctly stated the code in the benchmark is not from the PR
it inspired by the example given in https://github.com/knowm/XChange/issues/2495
to with I understand this PR is a nesesery preparation with its 'everything should be an interface' mentality

ps. I'm perfectly fine with many changes in this PR - I just find the part that treats value object as interfaces a bad design choice and a starter to implement lazy property mapping which will be slower in use cases where data in one value object is frequently accessed.

jnorthrup:
@timmolter my goals with this PR and beyond are nothing too grandiose.  

per the tweaked benchmark, this 

```
    List<IOrder> collect =
        Arrays.asList((new BigDecimal("1.4")), (new BigDecimal("1.6")), (new BigDecimal("1.5")))
            .stream()
            .map(bigDecimal -> (IOrder) () -> bigDecimal)
            .collect(Collectors.toList());
    IOrderBook orderBook = () -> collect;
```
from the same class contract below 

```
    List<NormalOrder> normalOrderStream =
        Arrays.asList((new BigDecimal("1.4")), (new BigDecimal("1.6")), (new BigDecimal("1.5")))
            .stream()
            .map(bigDecimal -> new NormalOrder(bigDecimal))
            .collect(Collectors.toList());
    NormalOrderBook orderBook = new NormalOrderBook(normalOrderStream);
```

this PR is the stepping stone to arrive at the sample code in  #2495 

things motivating this PR:
 * interfaces are completely public, which means you never violate a final class inheritance contract by implementing one inline or utilizing it out of package scope downstream.  
 * you never run the risk of fudging up any mutable hidden data with an interface.
 * adapters to final classes are more expensive than throwaway inline interface instantiation  
 * for DTO's most of the time it I only need a simple stateless transform like upcasting a string to BigDecimal or renaming an attribute from inbound json.  memoization of data should be the exception rather than the rule.  the java optimizer even has ways and means to make it much cheaper to beat on an inline-able transform.  
 * the hotspot is very rigorous.  when you specify final, it's only a suggestion to the compiler.  the jit promotes everything to final and static anyways, even when there are child inheritors, there will be multiple optimized methods.  
 * neither "final" nor "private" stop synthetic proxies or bytecode engineered solutions from violating your contracts and touching private parts.  you can turn off reflection errors trivially to circumvent private access without either of the above.   this incentivizes design contracts with interfaces to keep the amount of exceptional code lower.
 * in terms of performance, you can write c code in java, and it performs like c code in the jvm.  OO is not oriented toward high performance like you can do with off-heap pointer code in java.  This PR is to enable more clarity and not geared to achieving a transactional thruput metric.

walec51:
Interfaces have benefits but 'everything should be an interface' ideology is just an ideology

I bet that if you had the power you would make classes like BigDecimal, Instant, Currency etc.  in the JVM also as interfaces. One of the many reasons why they are not interfaces is that you cannot enforce a proper equals function among all implementations. That's why they are implemented as value objects. In this regard I believe we should follow Java standards in our Currency, CurrencyPair, etc. value objects.

jnorthrup:
https://swagger.io/swagger-codegen/

jnorthrup:
why couldn't we just "reuse" Currency if it does what we need?

...

image

jnorthrup:
i think that the age of this project proves the point about final classes, you can make them final but that won't prevent patch submitters from pushing scope creep into a final class, or apparently wanting 200+ static instances of something where 170 are unused.

walec51:
Javas Currency class actually uses the flyweight pattern which you don't understand - it does not have to instantiate all the instances as statics but it assures that USD for example is never instantiated more then once. As I said before - the instances cache is the core of the flyweight pattern, not lambdas nor anonymous classes:

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/Currency.java?av=f#123

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/Currency.java?av=f#311

We don't use it because it views currencies according to the ISO 4217 standard which does not apply to cryptocurrencies.

walec51:
I think all the pros and cons of the 'everything should be an interface' ideology and the lazy mapping pattern which you want do accomplish with it have been laid out.

Now I think the have to give the project owners time to decide:
a) do they want to implement lazy mapping interface implementations from now on
b) or should we stick to value objects created by adapters and just fix them to make the library more consistent

jnorthrup:
right now the emphasis has been declared as minimizing the time toward the
release cycle and preserving the dep tree.

Interfaces don't change how adapters do something, and there may be
non-trivial usecases that make Adapter classes concise solutions.

I'm biased towards whatever makes me get done with my tasks in intellij
faster.   In the long term, that includes a code generation strategy to
match impedences with this jaxrs codebase convention, and avoiding anything
remotely like mental accounting about any specific feature of this codebase.

timmolter:
@walec51 I'm leaning towards option B. I still don't understand the advantages of option A especially in light of making everyone change the way they use and contribute to XChange. Is there a burning need for this refactor and is it worth the pain? If yes, I'm happy to consider it, but so far I don't get it. @jnorthrup gives a lot of explanations with a lot of fancy words, but I struggle to find the information I need. @jnorthrup, What are the advantages for OTHER users (not your personal desire to include a code generation strategy to match impedances with this jaxrs codebase convention)?

Option B is more intuitive and doesn't require an interface and concrete class with the same name (or the "I" or "Impl" naming convention). People have been using XChange in this way for a long time, and it seems to work well. People understand the API and the structure, which is very important to me, and the growth of this library, >95% from outside PRs, is a testament to that. "Don't change a running system" seems really relevant here. Of course the DTOs and other parts of the library should be made more consistent, but that's for another day.

jnorthrup:
@timmolter consistency and brevity would be good.  I find both with
interfaces.  I didn't take iberties with my ideologies in this PR, I would
just say that with my PR in place, I could minimize the time i spend
contributing new modules and they will be represent brevity and clarity

@walec has brought in
 * performance arguments
 * idiomatic righteousness
 * pedantic design patterns terminology distractions
 * an apples to oranges benchmark

BUT NO PR

i just want to get done, submit a new module against what i consider a
win-win solution for everyone, let keep your adapters and every one of your
object creation choices made in the existing codebase, and let me write
code without requiring those distractions.

jnorthrup:
what does "lazy mapping interface implementations" even mean?

if you think you have some memory grid analytic usecase that makes it
crucial to preallocate everything everywhere, i dont see how an interface
stops you from doing so. If you have some idea about efficiency of
allocation and premature optimizations to offer, be my guest.  @npomfret
wants public final attributes.  that said, it still wouldn't prevent an
interface for accessing the same object through methods.

jnorthrup:
@timmolter
i just caught "but that's for another day."

Tim why would another day be different than this one? you have  a user
complaining about the problem, and instead of trolling, he submits a patch
that solves the problem without breaking the codebase, at a significant
cost of time.  you said make in issue, and discuss, and so we have.  this
does offer some user a dimension that is consistent, public interfaces are
consistent.

what complex language or java mythology confuses you on this issue?

jnorthrup:
please be specific about "everyone change the way they use and contribute
to XChange "

if this were a clean slate, i would probably suggest lombok, or google
autovalues, but since there is so much old growth, interfaces are the
cheapest and cleanest way to make everyone happy.  with some effort and
attention, there's no reason to assume the functionality changes.  this
just opens up new implementation options.

as far as i know, you run maven clean install, the same classes get built
the same way.  can you explain where I'm changing something ?  if users
have code built against a master release, the new code will not jump out of
the maven repo and start causing them to experience exceptions.  last i
heard, this PR would be against a major revision update.

if there were multiple tracks of development, again, the API writer user
could choose, but I don't have a choice, I have put a module out that
represents to whatever lattitude I can get to deliver my level of
understanding of what's good for my customer.

sadly, this seems like a personal bias, arguments made tailoring toward FUD
seem to prevail

jnorthrup:
the debate that has preceded this post has been chock full of ideological assertions, none made by me, and philosophy of best practices, none of them receiving effort prior to this thread, apparently.

i would say what's important is not articles about design pattern dictionary definitions or martin fowler ramblings (he has the luxury to tell clients to replace thier code and hire him $220/hr per FTE) 

whats important is that the reader, and contributor examine refactoring and legacy recommendations when the the project outgrows best practices and consensus, if they ever really got fair representation in the first place.

i've been saying this whole time that final classes make me spend more time writing more code.  interfaces allow me to write exactly the code needed in the place that's appropriate.  

i don't tend to get technical and in the weeds quoting books, I am the pretty much the book on java refactoring, i have too many rationales make simple dogmatic explanations stick.  when I try, it's overly verbose.

this PR covers 5 classes, or thereabouts, and not more, because i'm already aware of the sensitive issues surrounding orders and trades, and I don't actually use those features in the code I maintain. I can't rightly justify the time to build the whole refactor and then submit a module with the repeated example of conflicted PR's courtesy of the commiters.  so no im repeatedly tasked to explain all aspects of this course of action, and have been asked to close it.

but also, even though it took hours to present this PR's changes, and minimize the delta, without anyone footing the bill for my time, you can inline all the interfaces in seconds and go back to your final classes between stand-up breaks. and check it all back in, and you probably won't even change the imports in so doing.

here's an article that discusses exactly the steps i undertook for this PR, because the circumstances seemed to match the ones described. 


https://dzone.com/articles/refactoring-in-a-legacy-code-jungle

if anyone else knows other reasonably low-noise articles on refactoring, those would be helpful to this context as well.
cheers.

PR #2523, opened 2018-04-23
DTO Refactorings as Interfaces 

jnorthrup:
@timmolter has requested a laser focussed debate and vote on this PR.  

the author sees it as a good idea to simplify several central abstractions to pure interfaces.

  1. some default methods were extracted from the refactored classes into the replacement interfaces due to momentum, these make sense to preserve and the costs were within reason to do so.
  1. the author preserves 100% unit test success and keeps the classes converted to interfaces of the same classpath and package imports.
  1. there are no changes to poms or build methods in this PR.
  1. @timmolter has expressed intent to nominate this PR against a higher version increment designation, to indicate that this involves major changes.
  1. the changes involved in arriving at interfaces are the minimum timewise effort to pass the unit tests, consisting of an inconsistent, and evolving practice of promoting numerous constructors per-class to a non-uniform mix of factory constructors and Builders which are easily made consistent to satisfy _n_ developers who need to change code that isn't broken.  
     * at no point was the functionality changed for any given object's construction.
  1. there are also notable whitespace/formatting impacts due to the number of touched call-sites by these DTO's
  1. knowing that strong opinions would be associate with changing any class names to IClass or ClassImpl, the author instead chose to create .impl.Class packages to preserve the existing final and non--final classes and avoid importing these directly.  
  1. the author's goal is to enable _1_ unsatisfied developer to take advantage of concise interfaces in the creation of API modules.
  1. using intellij, and having experience with which intellij refactorings are safe, harmless, and might be advisable, files that were touched during several of the promotion to interfaces  were granted some automated lambda tweaks and map iterator simplifications that were not directly related to converting to Interfaces.  these were partially instigated by the promotion of constructors to builders which occurred in some labda and streaming constructs and were decidedly less elegant. (and explains why it has been important for so many developers and PR's to add new constructors to final and xchange-core DTO's ) 
  1. the author, ceding functionality over stylistic perfection, left a few artifacts of full classnames and potentially other evidence that refactoring did occur, without knowingly altering any code that tests could detect.
 1. the author acknowledges that everyone is an expert on classes, interfaces, pojo's, DTO's, formatting, line-endings, loop constructs, memory leaks, design patterns, and yet declares the main purpose of this PR is so that the reader and the author can write module code against pure interfaces without further justifying any modifications to the overwhelmingly satisfying codebase that exists prior to this PR.  
[...]

PR #2524, opened 2018-04-23
Refactoring the DTO's , RFC from @Timmolter

jnorthrup:
@timmolter has requested a laser focussed debate and vote on this PR.  


walec51:
The author of this PR seems to follow an ideology that interfaces are the golden bullets to solve every design problem and advocates that every thing should be implemented using them, even value objects. The author of this PR will insult everyone who does not agree that EVERYTHING should be used through an interface and will accuse you of being an incompetent programmer, ideolog, sabotage agent or a FUD spreader if you think otherwise. He also needs those interfaces to peruse a totally different way of implementing exchange model mapping / adaptation in his new modules without an overall consensus that this is the way our community wants to go.

To any one wondering how the "debate" looked like: https://web.archive.org/web/20180423031617/https://github.com/knowm/XChange/pull/2507

I had to resort to archiving it as the author of this PR is known to delete his own comments to obstruct the conversation.

He is also creating multiple PRs with the same code to ditch any comments:
https://github.com/knowm/XChange/pull/2523
https://github.com/knowm/XChange/pull/2522
https://github.com/knowm/XChange/pull/2507

@timmolter last comment was:
https://github.com/knowm/XChange/pull/2507#issuecomment-383413590

PR #2528, opened 2018-04-23
I currency

jnorthrup:

by request in Email @timmolter has requested a laser focussed debate and vote on this PR.

image
thank you  @walec51 I can represent my own ideologies without your assistance.
 
Apoligies to the reader, and to Adam Walczak and in advance for whatever the reader may perceive as too much explaining, or too little explaining, or interpreting comments as FUD when they were something else, or attempting to minimize the amount of irrelevant comments in the PR. see .11 above.

what's important is to review the code and report back to Tim.


walec51:
could you just stop deleting and creating PR's with the same code for the 8 time...

@timmolter do you confirm that this off-the-github conversation took place?

because to me this just looks like an attempt to hide any comments that question the new direction that its author wants to pave for this library

Issue #2394, opened 2018-03-09
The BitcoinCash BCC / BCH problem


npomfret:
Does anyone have an elegant solution to the problem of different exchanges using different symbols for certain currencies (BitcoinCash being the obvious example)?

As the conversion isn't handled by xChange is the best (and only?) approach, to convert the symbol in the client code on it's way into, and out of, the exchange api?

michaelscheung:
I have a not-so-elegant solution where I convert every Currency to a SafeCurrency (also CurrencyPair to SafeCurrencyPair), a wrapper class that I defined. The conversion looks something like this: `SafeCurrency.new(Currency.BCC, binance)`. SafeCurrency then stores an identifier, which is the full name of the currency (BitcoinCash in this case). I have some brute force logic for every exchange I am using to convert a symbol like "BCH" along with its exchange into an identifier.

This is a many to many relationship (e.g. BCC can be BitcoinCash or BitConnect, and BCH can also be BitcoinCash). Before back converting from SafeCurrency to Currency, I also generate a mapping for each exchange, from Currency to SafeCurrency. That way, I can safely convert back from SafeCurrency to Currency with something like `safeCurrency.unsafe(exchange)`.

Another proposal that may be more elegant is to use CoinMarketCap, which does a pretty good job of showing which exchanges use what symbols, and what currency that symbol refers to. Not sure if this can be obtained via their API. While their coverage is quite impressive, it does still miss a few cases.

On a related note, I've noticed a number of other inaccuracies as well. For instance, Binance has a currency pair VIBE/BTC, but this gets incorrectly parsed as VIB/EBTC. I'd be happy to see if any of my wrapping code I've written to fix these issues can be integrated directly into XChange, though like I have said, it is not very elegant.

npomfret:
Hmm, thanks for the reply.  Food for thought.

I can't help but think the currency constants should represent the currency rather than the symbol.  So in the client java code you use something like `BITCOIN_CASH` instead of `BCH`, or `BITCOIN` instead of `BTC`.  Then each exchange converts that into their own symbol.  Most would use a standard mapping for almost every currency because almost all exchanges use the same names for almost all currencies.  But, when an exchange uses `BCC` for bitcoin cash, or `XBT` for bitcoin it would convert in it's own non-standard way, both on the way into the http call, and on the way out.

michaelscheung:
[...]

PR #2531, opened 2018-04-23
fix Hitbtc USDT balances

JBetser:
simply adds a mapping USD -> USDT in the hitbtc balance adapter

### DIFF
--- a/xchange-hitbtc/src/main/java/org/knowm/xchange/hitbtc/v2/HitbtcAdapters.java
+++ b/xchange-hitbtc/src/main/java/org/knowm/xchange/hitbtc/v2/HitbtcAdapters.java
@@ -236,7 +236,12 @@ public class HitbtcAdapters {
     List<Balance> balances = new ArrayList<>(hitbtcBalances.size());
 
     for (HitbtcBalance balanceRaw : hitbtcBalances) {
-      Currency currency = Currency.getInstance(balanceRaw.getCurrency());
+      String ccy = balanceRaw.getCurrency();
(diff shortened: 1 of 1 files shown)


walec51:
does this code mismatch only happen in wallet information?
other endpoint use USDT correctly?

timmolter:
My gut tells me this will break things. Can someone who uses Hitbtc chime in?

If an exchange uses USDT as a counter currency, then just use the USDT Currency. If it uses USD, then use the USD Currency.

fxmozart:
Hitbtc doesnt allow withdraw on usdt , so prices are skewed.



Olivier G.

walec51:
in their UI everywhere I see USDT instead of USD

now how does this look in their API?
do they use USD code instead od USDT everywhere?
or is it just in the wallet endpoint as this PR suggests?

fxmozart:
Its not buggy per say , they just dont allow users to withdraw in usdt for
a few months now .




Olivier Guiglionda


Mobile Phone : +33768461000

JBetser:
this problem is unrelated to the fact USDT withdrawals are currently disabled.
you can still read the balances and trade USDT in hitbtc.

walec51:
well then this is an extension of this problem https://github.com/knowm/XChange/issues/2394

and hacking it for one endpoints response object is not a solution - it would be introducing an inconsistency only

besides, they may already be applications that use this module with USD despite knowing that its actually USDT on this exchange

walec51:
I'd suggest that for now you should solve the problem on your applications level and join the disquisition on how to solve https://github.com/knowm/XChange/issues/2394

PR #2532, opened 2018-04-24
Add get order book in huobi

cacofishadct:
1. Add get order book in huobi
2. some formatting change.
3. Add deposit address, withdraw api in huobi
4. fix huobi place limit order issue


xchange-huobi/src/main/java/org/knowm/xchange/huobi/dto/marketdata/HuobiDepth.java
```diff
@@ -0,0 +1,73 @@
...
+    return asks;
+  }
+
+  public Date getTs() {
+    return ts;
+  }
+
+  public void setTs(Date ts) {
+    this.ts = ts;
+  }
```
walec51: is this attribute needed?
I find it odd that the whole object is immutable except for this property

xchange-huobi/src/main/java/org/knowm/xchange/huobi/dto/marketdata/results/HuobiDepthResult.java
```diff
@@ -0,0 +1,35 @@
...
+      @JsonProperty("status") String status,
+      @JsonProperty("ts") Date ts,
+      @JsonProperty("tick") HuobiDepth tick,
+      @JsonProperty("ch") String ch,
+      @JsonProperty("err-code") String errCode,
+      @JsonProperty("err-msg") String errMsg) {
+    super(status, errCode, errMsg, tick);
+    this.ts = ts;
+    this.ch = ch;
+    getResult().setTs(ts);
```
walec51: this is odd, these DTO object should represent the exchanges response and should not do any premature mapping nor have data coping logic

xchange-huobi/src/main/java/org/knowm/xchange/huobi/dto/marketdata/results/HuobiDepthResult.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  public Date getTs() {
+    return ts;
+  }
+
+  public String getCh() {
+    return ch;
+  }
```
walec51: pls you document what are those two attributes and do we intent on using them somehow

xchange-huobi/src/main/java/org/knowm/xchange/huobi/dto/marketdata/HuobiDepth.java
cacofishadct: Copied :). Removed now.

xchange-huobi/src/main/java/org/knowm/xchange/huobi/dto/marketdata/results/HuobiDepthResult.java
cacofishadct: Also copied. I am not very sure about the usage. Removed them in next change.

xchange-huobi/src/main/java/org/knowm/xchange/huobi/dto/marketdata/results/HuobiDepthResult.java
cacofishadct: Removed

PR #2534, opened 2018-04-24
Refactoring the DTO's , RFC from @Timmolter 

jnorthrup:

by request in Email @timmolter has requested a laser focussed debate and vote on this PR.

please find the review button to keep the context related to the contents of this pull request  
image

Thanks!


timmolter:
I not interested in off-topic non-technical discussion so please keep that to a minimal as it only slows down the process for me, as I'm very time-limited already. 

Filtering out everything that has been "discussed", what I see is that this PR boils down to this: "the author's goal is to enable 1 unsatisfied developer [the author] to take advantage of concise interfaces in the creation of API modules."

OK, it doesn't seem nefarious, but if someone wants to change a core concept that affects all users, I'd like to see some more voices weigh in and a general consensus occur, like on past topics. If people don't speak up, I assume there is no interest, and that's that.

@walec51 put in the effort to show a benchmark, and while @jnorthrup says it's comparing apples to oranges, from what I could tell I think he exactly benchmarked what @jnorthrup intends. Here's my comment addressing that to which I'm still waiting a direct response:

> @walec51 Your benchmark shows that mapping of a response object returned by an API in the anonymous inner class is much slower than getting a property already mapped one by an adapter. Thanks for doing the benchmark BTW! Where is the anonymous inner class method being used though in the PR? Having 439 files changed (which is expected for a major refactor) makes it hard for me to find the case we are debating about.

Note that while this PR doesn't have anonymous inner class mapping (as far as I can tell), @jnorthrup commented that "this PR is the stepping stone to arrive at the sample code in #2495", which does show anonymous inner class mapping.

> @jnorthrup Maybe you can make clear your ideas/intentions (perhaps plans for a followup PR) with using or not using anonymous inner classes to adapt raw to generic DTOs. It's not clear to me where you stand or what your intentions are.

So with no general interest in this change, and a benchmark showing really bad performance for a case that @jnorthrup this PR is the stepping stone for, I'm at this point leaning towards not allowing this to be merged. 

@jnorthrup All I need from you is to answer my question: How is this useful to other people?

timmolter:
@walec51 Yes, he emailed me and I responded that I'm currently traveling, sorry for my slow response times because of that, and that he should post a new issue about the interfaces to get a targeted discussion and a vote like we've done in the past.

jnorthrup:
> @jnorthrup All I need from you is to answer my question: How is this useful to other people?

I think it's useful to the people I'm billing, to minimize the amount of trust they require in "Class" x versus "Interface" x  for starters.  there's less trust in "idealogies" of "it is what it is" this way when the code shows interface A from one API mapped in the fewest places with the most direct declaration of casting, signatures, and transforms to interface B in a foreign API.

> @walec51 put in the effort to show a benchmark, and while @jnorthrup says it's comparing apples to oranges, from what I could tell I think he exactly benchmarked what @jnorthrup intends. Here's my comment addressing that to which I'm still waiting a direct response:

At the end of the day, if I'm maintaining a module and there is in fact a 20x performance gap in using computed methods, how does this affect @walec51 ?  honestly the chances of that scenario are remote, but in decentralized solutions it is possible that this is the fastest way to exonerate myself from accusations of hidden intentions in undisclosed class code.

walec51:
I on the other hand one day would like to trust that currencyA.equals(currencyB) or currencyB.equals(currencyA) will return true if both currencies are BTC even if those object ware instantiated by different modules. If value objects like Currency, CurrencyPair will be interfaces then it will never be possible to enforce a proper implementation.

Instead of doing so we could make them final classes, make their constructors private to ensure that the factory method manages all instances as it should according to the flyweight pattern - I'm talking about the pattern as it is described in literature, not the one imagined by @jnorthrup. The Currency class is already trying to do just that https://github.com/knowm/XChange/blob/develop/xchange-core/src/main/java/org/knowm/xchange/currency/Currency.java#L320 - we need to deal with https://github.com/knowm/XChange/issues/2394 to assure that it can do it for every instance and by doing so assure currency identification consistency.

Consistent object identification and equals symmetry are two of many reasons why value objects are usually implemented as value objects and not represented as interfaces.

jnorthrup:
@walec51 this looks like an attempt to drag this PR into the weeds and capitalize in the maintainer's limitted time.  issue comparators have not been modified in the codebase.  nor has the cardinality of instances.

walec51:
your accusing me of that since my first comments on https://github.com/knowm/XChange/pull/2507

ok, I gave my arguments why I see the 'all should be an interface' ideology as weak both from a performance and OOP design stand point

over and out - just don't delete and recreate this PR for the 10th time to hide my comments and the link to most of the disquisition

jnorthrup:
@timmolter @walec51 because this codebase is in fact heavily correlated with crypto, and sometimes the crucial all-encompassing "my life savings is in this" sentiment, I would specifically avoid any intersection with code maintained by participant who works toward character assassination and filibustering or who wasn't vetted personally.  this PR is in fact a stepping stone to enabling the freedom from trusting trolls and depending on the good will of future maintainers after the original author has moved on and can give no time or attention to scrutinize momentum.

jnorthrup:
Should i ask the "general consensus" of true=="CRY"=="CRY" token on Idex  if i wanted to specify 100% of the boundaries of my exchange's potentially decentralized, potentially trustless interfaces? that is a very real, very expensive mistake on eth.

timmolter:
I'm not sure how you would argue it. Where in the PR is that addresses specifically? 

Why do you keep avoiding my questions? I'm trying to give you the benefit of the doubt and understand what you are trying to accomplish. I can tell you are a smart person and an experienced developer, however maybe not the best communicator???

jnorthrup:
>How so? Technically.

if i am granted the lattitude to operate Xchange modules by interfface contracts, then I can speciufy what equality criterion exist for instruments, and I can control every facet of which singletons are created in a given module, and minimize any namespace attacks on client code interacting with XChange that have yet to be discovered in the cited currency usecase

jnorthrup:
> Why do you keep avoiding my questions? 

for completeness, which questions have been avoided?

timmolter:
Please take a look at all my past questions in your PR threads.

Would your "solution" help solve the issue in #2394? Maybe this would be a chance to highlight the benefits of your interface design refactor, but in a smaller easier pill to swallow. If you can show it's technically superior and is more secure and solves an outstanding problem along with an explanation besides "it's just better", I'd sure be interested in seeing it. 

>  the amount of trust they require in "Class" x versus "Interface" x for starters. there's less trust in "idealogies" of "it is what it is" this way when the code shows interface A from one API mapped in the fewest places with the most direct declaration of casting, signatures, and transforms to interface B in a foreign API.

Can you show specific code that addresses those issues as a way to illustrate how you made things more "trusting"?

jnorthrup:
a currency class where not only  Currency("BTC)==Currency("BTC") but also Currency("XBT")==Currency("BTC") is the current state of the codebase.

im providing a module interface for traders where there are two tokens:

https://ethplorer.io/address/0xf41e5fbc2f6aac200dd8619e121ce1f05d150077
image


https://ethplorer.io/address/0x47808c1ea94ede42cfa2fa429140d097f177d6b3#tab=tab-issuances
image

...

I have known of this namespace collision for "CRC" tokens already from personal experience, but this discussino has unearthed exactly the unseen consequences of "the road to hell is paved with good intentions"   ..

I need to caution the users of my new module to take extra precautions not to FOMO on a token by symbol, ever! this could be the domain of scammers.  I'd like the lattitude to handle the Currency comparators explicitly, now that the subject has been given ten minutes of thought.

jnorthrup:
im sorry, im thinking out loud, im not trying to take this discussion into the weeds, but the original premise was simple, and more importantly, this discussion has uncovered a nascent unplanned benefit of the PR request.

the main goal was to make the adapter code more terse and localized.

jnorthrup:
>Would your "solution" help solve the issue in #2394? Maybe this would be a chance to highlight the benefits of your interface design refactor, but in a smaller easier pill to swallow. If you can show it's technically superior and is more secure and solves an outstanding problem along with an explanation besides "it's just better", I'd sure be interested in seeing it.

i went over the past conversations and trtied to distill all the concerns in the PR description, and they are related to #2394 sample code, but that sample code was a 5 minute draft.

it was pointed out the convention creates computed methods for Balance.available using two stack locals, but as it turns out, all Balance.available and Balance.total accessors already compute against 6 Balance attributes anyways.  

is this superior? unknown.  this is defensive, and this allows my client to read that adapter solution on one page, and see the entire lifecycle of the entire object graph from the parameters, to the request, to the return value in one shot.

i consider this option to be the one that respects everyone's time who is invested in this module's solution.

jnorthrup:
i would equate https://github.com/knowm/XChange/pull/2534#issuecomment-383983600 with an answer to address potential security vectors.

i would equate https://github.com/knowm/XChange/pull/2534#issuecomment-383987564 to address whether this adds value 

but most importantly, this is simply a new option, and not imposed on working modules, and there is some possible beginning of a "consistency" direction that you'd like to table. this pr addresses AccountInfoService ExchangeMetadata, and may or may not set a precedent for work on MarketDataService and TradeService refactorings.

does this satisfy https://github.com/knowm/XChange/pull/2534#issuecomment-383983091 ?

jnorthrup:
edit, #2495  was my my code sample issue.

im not yet familiar with #2394

jnorthrup:
ok i have read The BitcoinCash BCC / BCH problem #2394 and #2535 and honestly there appears to be a lot of workarounds and bikeshedding that appears to come from all modules depending on the "Class" Currency without the options enabled by the "Interface" Currency in this PR.

It looks like there's a Gordian Knot to cut here.  I just pushed Currency and CurrencyPair to interface unknowning of any mounting measures of desperation in other threads.  

while I'd like to remain ego-less, I'd still prefer Interfaces declared in this PR to any other special treatment I' glanced at briefly, and with this PR, the existing debate can proceed on the trajectory it already has without changing the arity and cardinality of instances created from the factory methods.

jnorthrup:
@timmolter It seems like you are unfamiliar with the terminology and motivations behind  defensive coding.  when I refer to common defensive coding idioms you appear to be asking language about a silver bullet.

if I were to say this is a silver bullet, i fear I might get trolled.  what I can say is that for motivations of publishing more concise code, my choice would be to publish against DTO's that have interfaces.

this is not a silver bullet, but it allows my clients to read the minimum amount of code to understand the entire sequence diagram of any API call where the parameters are interfaces and the return values are interfaces.

i attempted to illustrate that for the module I'd like to publish, there's a crowded symbol namespace of *Currency* designations.

Interfaces can allow me to make guarantees that there are no synonyms for any Currency instantiated in the module, regardless of method performance, or memory footprint. I can publish an end-to-end solution inline using interfaces and can ask the reader of the code to make the fewest possible assumptions about what happens in those interfaces by looking at what's inline.   typically, this is also consistent and readable, and easily maintained loops.

_please_ let me know any specifics I seem to be missing wrt to this PR.  Thanks

jnorthrup:
>Can you show specific code that addresses those issues as a way to illustrate how you made things more "trusting"?

if the time invested in this PR leads to a pull, I will invest time in those changes before submitting a module that's waiting for resolution on this PR.  if this PR is rejected, I will submit a module that lacks this PR. 

I have already done everything in my power to trim long explanations, and to be thorough, and to submit code samples, and to fix ongoing conflicts.  I wish you'd ask someone else that you do trust to explain whether they think that Interfaces could help them avoid problem X.  this pr is intended to enable a better quality of submission on behalf of a third party.

jnorthrup:
>Can you show specific code that addresses those issues as a way to illustrate how you made things more "trusting"?

can you trust this Currency Object?
 
https://github.com/knowm/XChange/pull/2534/commits/516c4ffc506e3eb7a922892578ecc00ff87b89df

walec51:
without looking at its implementation, no I cannot
even with knowing its implementation I cannot because from another module I can get crazy troll implementation

```
    String someStackVariable = "BTC"l
    Currency btcFromUltraProJnorthrupModule =
         new Currency() {
 
           @Override
           public String getCode() {
             return someStackVariable;
           }
 
           @Override
           public CurrencyAttributes getAttributes() {
             return null;
           }
 
           @Override
           public boolean equals(Object obj) {
             boolean ret = false;
             if (obj instanceof Currency) {
               Currency currency = (Currency) obj;
               ret = Objects.equals(currency.getCode(), someStackVariable);
               //
             }
           return ret;}
         };
    String otherStackVariable = "BTC";
    Currency btcFromTrollsModule =
         new Currency() {
 
           @Override
           public String getCode() {
             return otherStackVariable;
           }
 
           @Override
           public CurrencyAttributes getAttributes() {
             return null;
           }
 
           @Override
           public boolean equals(Object obj) {
             boolean ret = false;
             if (obj != null && obj.getClass() == getClass) {
               Currency currency = (Currency) obj;
               ret = Objects.equals(currency.getCode(), otherStackVariable);
               //
             }
           return ret;}
         };

     assertEquals(btcFromUltraProJnorthrupModule, btcFromTrollsModule); // yee interface defence works

     assertNotEquals(btcFromTrollsModule, btcFromUltraProJnorthrupModule); // ups... it actually does not...
```

walec51:
so we should debug our apps to see what implementations fly under the interface and check if they have a trustworthy impelmentation? you do realize that one of the main features of this library is to provide an unified model so that we can write code that is independent from exchange APIs and process data from many exchanges simultaneously?

in the example you gave interfaces gained us nothing

I wanted to stay silent but I'll give you a hint

@timmolter expects you to give a example of a feature or a defensive technique that would be considered useful to developers other then yourself and is not easily implementable if Currency or CurrencyPair ware not interfaces

no such feature was presented in this PR nor in the related issues so there is no justification to merge this other than ideological views on OO design

jnorthrup:
>Would your "solution" help solve the issue in #2394? 
this unit test is demonstrating the issue.  this PR is demonstrating an indifferent observer's trivial shim to demonstrate an option unavailable before the PR.

jnorthrup:
me >this PR doesn't promise safety from other modules, but I did just demonstrate a solution to undesired singleton comparisons.

this PR makes it easier to avoid commingling code with other modules.  it also makes it easier to minimize the code allocated to seperate adapters.

jnorthrup:
@timmolter are you satisfied that your inclination to reject this PR, or do you need additional examples and debate, character aspersions are unpleasant to endure in the name of patience with your process.

jnorthrup:
> @timmolter expects you to give a example of a feature or a defensive technique that would be considered useful to developers other then yourself and is not easily implementable if Currency or CurrencyPair ware not interfaces

the unit test shows a singleton.  the issue  #2394 apparently limits the accessibility to certain object references in certain conditions.  

the potential for injecting singleton mappings with malicious synonyms is undesirable.  

the code doesn't require review to use the xchange-core.  just use write a gui and import the core thru maven.   

you can at whatever level poison one or more currencies in the singleton mappings for an unknown number of consumer clients to execute a trade.   all you need is a jar file and an xchange gui.  new exchanges are opening up all the time and always looking for client interface help.

"useful to developers" would be

1. you generate interfaces from a  swagger doc or some other jaxrs interfaces.  specifically interfaces.  don't pass the buck. but if you have to pass the buck, code generators. 
1. that the objects you consume from XChange are also interfaces
1. verify your intent is to avoid maintaining every line of code you can, and minimize every possible opportunity for poisoned static variables, maps, and whatever cat and mouse objects come from attempting to make things final or immutable.  the simplest thing is don't import code. deon't seperate functions into adapters. don't even write comments you're not willing to maintain, just prove entry and exit are interfaces. 
1. that when your client's customers start spamming the message boards about a lack of support, about dropped transactions, etc. for misappropriating value through symbol mappings or value calculations, your code is without external influences to point-in-time class files that you didn't vet, and that they may in fact have a jar file that corresponds with  a maven release md5 that traces back to a release with your commit.

this list is not exhaustive, but this debate is also conflated.

PR #2537, opened 2018-04-25
Interfaces for dto entities

jnorthrup:
By Tims request, this was applied to potential currency challenges and appears to offer a pragmatic alternative to the behaviors in xchange core dto.


jnorthrup:
>he best chance you have of convincing me and others of using interfaces would be to show what it would offer, preferably in a short and concise PR demonstrating said feature.

@timmolter i thought i had addressed this with one question:
>can you trust this currency class?

but instead of answering that question you closed the PR. 

I took the PR, and demonstrated a Unit test that shows the undesired behavior and tried, in vain, to illustrate to you, @timmolter that I was satisfied , that this conclusively resolves #2394 for all foreseeable usecases i have with Currency class, and that there is no unseen code in this solution.


```
String someStackVariable="BTC";
    Currency avoidXBT =
        new Currency() {

          @Override
          public String getCode() {
            return someStackVariable;
          }

          @Override
          public CurrencyAttributes getAttributes() {
            return null;
          }

          @Override
          public boolean equals(Object obj) {
            boolean ret = false;
            if (obj instanceof Currency) {
              Currency currency = (Currency) obj;
              ret = Objects.equals(currency.getCode(), someStackVariable);
              //
            }
          return ret;}
        };
    assertEquals(avoidXBT,Currency.BTC );
    assertNotEquals(Currency.XBT, avoidXBT);
    assertNotEquals(Currency.LTC, avoidXBT);
```

Issue #2512, opened 2018-04-17
web3j and/or bouncycastle


jnorthrup:
Ethereum uses a specific array of sha3 implementations and this drives ethereum wire protocols for at least one DAP I maintain.  

the checksums are backed by bouncycastle but the web3j layer is impl specific, and tries to be relevant to the web3js api as well.

can i include this in an upcoming PR?  there's no reasonable substitute for Idex

timmolter:
What's the scope of the upcoming PR? I never saw an exchange requiring raw ethereum wire protocols.

makarid:
It is need for decentralized exchanges, in order to sign orders and cancel them.

jnorthrup:
the requirement is here and below it https://github.com/AuroraDAO/idex-api-docs/blob/master/README.md#order 

web3j and contracts in java lack examples, this took a week to bitbang with the architect

timmolter:
OK, go ahead, just exclude the dependencies not needed under web3j, as it seems to pull in a lot of stuff. Ideally you'd just depend on bouncycastle.

jnorthrup:
![image](https://user-images.githubusercontent.com/73514/39330316-6b9dd824-49cb-11e8-87e2-d6b9d7f72189.png)

PR #2544, opened 2018-04-26
idex decentralized exchange

jnorthrup:


jnorthrup:
https://github.com/knowm/XChange/issues/2512#issuecomment-384779342 the deps look clean

walec51:
as for some strange reason I cannot leave comments in the code then I'll have to do it here:

proxies should not be created in each method call: https://github.com/knowm/XChange/pull/2544/files#diff-0168def634f27171c78aef20ac7ed214R44

silent errors are pure evil - those exceptions should be just wrapped in a ExchangeException: https://github.com/knowm/XChange/pull/2544/files#diff-0168def634f27171c78aef20ac7ed214R70

you should try to follow this in new modules: https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices

jnorthrup:
agreed on all points.

xchange-idex/src/main/java/org/knowm/xchange/idex/IdexSignature.java
```diff
@@ -28,13 +29,13 @@
     ECKeyPair ecKeyPair;
     try (ByteArrayOutputStream sig_arr = new ByteArrayOutputStream()) {
 
-      if (IdexSignature.debugMe) {
+      if (debugMe) {
```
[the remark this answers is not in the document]
jnorthrup: this data should not appear in any logs.  removed.

xchange-idex/src/main/java/org/knowm/xchange/idex/IdexTradeService.java
```diff
@@ -78,48 +94,34 @@ public OpenOrders getOpenOrders() {
...
+    return getOpenOrders();
   }
 
   @Override
   public boolean cancelOrder(String s) {
-    CancelApi proxy =
-        RestProxyFactory.createProxy(
-            CancelApi.class, getIdexExchange().getExchangeSpecification().getSslUri());
-    return false;
+    throw new UnsupportedOperationException("Orders Api Unnavailable at this time");
```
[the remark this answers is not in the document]
jnorthrup: this change globally affects this set of files.

image

xchange-idex/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    </organization>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>org.knowm.xchange</groupId>
+            <artifactId>xchange-core</artifactId>
+            <version>4.3.6-SNAPSHOT</version>
+        </dependency>
+        <dependency>
```
timmolter: This should go in the root pom.xml under dependencyManagement. Here, just remove the version.

xchange-idex/src/main/java/org/knowm/xchange/idex/IdexExchange.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    public static CurrencyPair getCurrencyPair(String market) {
+      CurrencyPair currencyPair;
+      Iterator<String> syms = asList(market.split("_")).iterator();
+      currencyPair = new CurrencyPair(syms.next(), syms.next());
+      return currencyPair;
+    }
+
+    public static void main(String... args) throws IOException {
```
timmolter: This doesn't belong here, perhaps move it to examples.

xchange-idex/pom.xml
jnorthrup: this was from huobi, the one before in the order, for consistency

xchange-idex/src/main/java/org/knowm/xchange/idex/IdexExchange.java
jnorthrup: will revisit this with a testnet example after some discussion.

xchange-idex/pom.xml
timmolter: I meant the org.web3j one :)

pom.xml
```diff
@@ -255,6 +256,13 @@
                 <artifactId>logback-classic</artifactId>
                 <version>${version.qos.logback}</version>
             </dependency>
+
+            <dependency>
+                <groupId>org.web3j</groupId>
+                <artifactId>crypto</artifactId>
+                <version>3.2.0</version>
```
timmolter: Ant reason you chose version 3.2.0 when 3.4.0 is available?

PR #2545, opened 2018-04-26
Support for Abucoins pagination in user trades history

walec51:


xchange-abucoins/src/main/java/org/knowm/xchange/abucoins/service/AbucoinsDigest.java
```diff
@@ -23,6 +23,9 @@ public String digestParams(RestInvocation restInvocation) {
     String method = restInvocation.getHttpMethod();
     String path = restInvocation.getPath();
     String queryParameters = restInvocation.getQueryString();
+    if (!queryParameters.isEmpty()) {
+      queryParameters = '?' + queryParameters;
+    }
```
walec51: without adding this I actually got an INVALID_HASH error each time query params ware used

PR #2548, opened 2018-04-27
[BitFinex] Added exception mapping

traviscollins:
Adding exception mapping with a few exceptions to all BitFinex API calls.


timmolter:
nice!

PR #2556, opened 2018-05-02
[exmo] new exchange implementation 

npomfret:


npomfret:
@michaelscheung it's pretty much done.  Would you like to test?

timmolter:
Why Maps and not exchange-specific DTOs?

npomfret:
Partially because it's a work in progress and I wanted to get something working to the person who requested it.

The DTO's aren't required for a working implementation and can be added in time.

npomfret:
@michaelscheung the only bit of the public api that's not implemented is the withdrawFunds method (waiting on some api auth from the exchange).  the rest is ready to go.

michaelscheung:
I noticed that `placeMarketOrder` throws a `NotAvailableFromExchangeException`, but it looks like they do support market orders - it would be a call to `order_create` (same endpoint as limit orders) with type `market_buy` or `market_sell`. Also, I think we can add this to `TradeService`:
```
@Override
  public TradeHistoryParams createTradeHistoryParams() {
    return new ExmoTradeHistoryParams();
  }
```

Otherwise, everything else looks good from my testing.

npomfret:
@michaelscheung good spot, thanks

I also noticed that certain user trades aren't coming though.  It seems some markets (KICK_ETH for example) are coming back empty.  Not sure why.  I've raised a support ticket.

timmolter:
you need to update the exmo pom to core version 4.3.7-SNAPSHOT

npomfret:
@michaelscheung i've implemented and tested the last method (withdraw funds) so i think this is done.  If you're happy please transfer the bounty to 19vHtAbDvvWzpctFsGBNMogQVuHB2kudqy

@timmolter if the build goes green please feel free to merge - it might be a good idea to squash the commits.  I'll add the strongly type classes to the API in time but (IMO) they're not required in order to have a working implementation of this exchange.

michaelscheung:
I haven't had a chance to test it in depth yet, but I've sent over the bounty of 0.075 BTC and you should be receiving it shortly. Thanks again!

Issue #2558, opened 2018-05-03
hitbtc v2 issue


npomfret:
I just noticed another problem with the hitbtc v2 implementation.  It's using the clientOrderId in the order id fields for limit orders and user trades.  This is a change from the v1 implementation and inconsistent with the rest of the implementations.

Any objection to me using the _correct_ (exchange assigned) order id?

We could add the `clientOrderId` to xchange as a separate field in the future.

npomfret:
https://github.com/knowm/XChange/pull/2560

PR #2560, opened 2018-05-05
[various] mostly tradehistory improvements

npomfret:


npomfret:
Also addresses https://github.com/knowm/XChange/issues/2558

PR #2563, opened 2018-05-07
[core, hitbtc] introduced concept of `clientOrderId` and implemented in Hitbtc

npomfret:


npomfret:
@timmolter good question, but one I can't answer I'm afraid.  It's actually a nice feature, I wish more would support it.

walec51:
clientOrderId - Unique identifier for Order as assigned by trader. Uniqueness must be guaranteed within a single trading day, including all active orders.

this looks like an unique feature for HitBTC - I personally find its week uniqueness constraint dangerous

are the any features which you cannot implement using the unique orderId generated by the exchange?

I personaly don't see any reason why this hitbtc specific feature should be present in a library which tries to unify access to many exchanges using a common API

timmolter:
@walec51 That's exactly why I asked. I cannot really justify adding that if only a very few or in this case one exchange has it.

walec51:
I would propose to remove changes to the core module from this PR an do the following:

in HitbtcTradeService expose methods:

```
    @Override
    public String placeMarketOrder(MarketOrder marketOrder) throws IOException {
        return placeMarketOrder(marketOrder, null);
    }

    public String placeMarketOrder(MarketOrder marketOrder, String clientOrderId) throws IOException {
        //...
    }
```

this will allow you pass this id for this exchange without the need to use exchange DTOs for the order object

npomfret:
Ok, this change seems unpopular.  I'll try something else - more focused to the HitBTC exchange.

npomfret:
Thanks for the feedback everyone.  I've had another go, a less invasive Hitbtc-specific implementation https://github.com/knowm/XChange/pull/2566

PR #2624, opened 2018-06-01
don't use port 80 if it's https

unlimitedsola:
this PR corrected the port setting when using HTTPS connection. It also set the port to 443 automatically when `sslUrl` is being set.


timmolter:
Please fix merge issue and push. Thanks.

pom.xml
```diff
@@ -210,13 +207,6 @@
                 <version>${version.commons.lang3}</version>
             </dependency>
 
-            <!-- Charting -->
```
timmolter: why remove this? It's needed.

timmolter:
There's too much lumped into this PR and there's a lot more changed than described. Smaller incremental PRs are better. I will go ahead and remove taurus and empoex and you can resubmit a PR with your other changes. Please leave XChart in the root pom.

PR #2626, opened 2018-06-01
Add ZenCash currency pairs

vherus:
ZenCash is becoming increasingly popular and is available on an already impressive and growing number of exchanges. We want to use this XChange package in our own projects, and this PR gives us that option.


walec51:
not that you need those statics, you can just do `new Currency("ZEN")` and `new CurrencyPair("ZEN", "BTC")` instead

Issue #2533, opened 2018-04-24
Introduction of a SecurityException


walec51:
I would like to introduce a `SecurityException extends ExchangeException` which would be thrown in the following situations:

- authentications failed - credentials unrecognized by the exchange
- authorization failed - no permission to access some endpoint
- wrong request signature - in exchanges where this type of error is thrown if the API keys are unrecognized

The purpose of this exception is to separate situations where an applications user could have passed improper credentials from technical errors. In consequence this will allow us to provide a better UI experience and error messaging.

Is every one ok with this?

If you are than I'll add this functionality to the following exchanges in the next two weeks:
Binance, BitBay, Poloniex, Bitfinex, Bittrex, Cryptopia, BitMarket, Abucoins

npomfret:
Sounds good to me.

timmolter:
Yes, please do.

walec51:
because a SecurityException is already in java.lang I'm perusing the name ExchangeSecurityExchange for this case

walec51:
If someone has 5 min please check if the code is going in the right direction:
https://github.com/knowm/XChange/pull/2628

timmolter:
Looks good except that you should put the error handling method in the BaseService, rather than in the adapter. See `GDAXBaseService`.

PR #2628, opened 2018-06-01
More informative exceptions

walec51:
For now the first commit showing the direction I want to take.

Adding a *ErrorAdapter to every module which tries to throw more informative ExchangeException child class if it recognizes a common error situation like: passing a not supported currency pair or failing to authenticate.

The CryptopiaErrorAdapter depends on the  CryptopiaBaseResponse. Other exchanges with more HTTP compliant APIs will require passing the HTTP status and error response object to an error adapter.


walec51:
expanded concept to Poloniex and started attacking #2533

please give me some feedback if this is going in the right direction

timmolter:
Looks goo except that you should put the error handling method in the BaseService, rather than in the adapter. See `GDAXBaseService`.

walec51:
This will make it harder to test:

```
/** @author walec51 */
public class CryptopiaBaseServiceTest {

  private Exchange exchange;
  private CryptopiaBaseService service;

  @Before
  public void setUp() {
    exchange = mock(Exchange.class);
    // ... exchange specification mocking ...
    service = new CryptopiaBaseService(exchange);
  }

  @Test(expected = CurrencyPairNotValidException.class)
  public void throwIfErrorResponse_wrongMarket() {
    // Cryptopia actualy sets successful = true in this situation
    CryptopiaBaseResponse response =
        new CryptopiaBaseResponse(true, null, null, "Market STRG_BTC not found");
    service.throwIfErrorResponse(response);
  }

  @Test(expected = ExchangeException.class)
  public void throwIfErrorResponse_unrecognizedErrorMessage() {
    CryptopiaBaseResponse response =
        new CryptopiaBaseResponse(false, null, null, "Some other error message");
    service.throwIfErrorResponse(response);
  }

  @Test(expected = ExchangeException.class)
  public void throwIfErrorResponse_unsuccessfulWithoutErrorMessage() {
    CryptopiaBaseResponse response = new CryptopiaBaseResponse(false, null, null, null);
    service.throwIfErrorResponse(response);
  }

  @Test
  public void throwIfErrorResponse_successfulWithoutErrorMessage() {
    CryptopiaBaseResponse response = new CryptopiaBaseResponse(true, null, null, null);
    service.throwIfErrorResponse(response);
  }
}
```

as I have to add a lot of mocking just to test error handling.

Do we really want this?
Do many other modules implement error handling like GDAXs one does ?

walec51:
@timmolter please decide if you agree with my argumentation or should I just stick with the *BaseService.handleError method pattern as shown in GDAX

walec51:
I've pushed the refactored code and also made exception / DTO handling more consistent in the two modules. Please take a look.

My approach contradicts one rule from from step 7 of https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices

> Also catch all exchange-specific Exceptions in the *Raw class and re-throw them as ExchangeExceptions

To me exchange specific exceptions are just a specific case of DTO and therefore I see the above rule internally inconsistent with another rule in the same paragraph:

> The *Raw class should fetch the JSON and deserialize it into exchange-specific DTOs.

If *Raw classes should stick to the Exchanges representation of data then I think this should also apply to exceptions / error data. In one case (like in this PR) the exchanges API will return an error message but in another it can return an error code which may have to be mapped to a message and some context data. I assume that we have *Raw classes for rare cases when XChanges generic data model is too generic and one has to call those *Raw methods directly to get access to exchange-specific fields. I think that's a good approach and throwing exchange specific exceptions would be an extension of it.

timmolter:
looks good

walec51:
thank you, I modified step 7 in https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices in accordance to this change and I'm proceeding with enhancing error handling in more modules

PR #2631, opened 2018-06-05
change api-nonce to api-expires for bitmex restful api

declan94:
When running multiple programs, the original 'api-nonce' signature mode has significant problems.

This PR changes the signature mode for Bitmex to the new version, which specifies 'api-expires' instead of 'api-nonce'.

Here's the relative api document: [https://www.bitmex.com/app/apiKeysUsage](https://www.bitmex.com/app/apiKeysUsage)


timmolter:
what is `lag` for?

declan94:
@timmolter It means how long before the signature expires. The variable name may be confusing, do you have any suggestions?

timmolter:
According to their docs: api-expires: A UNIX timestamp after which the request is no longer valid. This is to prevent replay attacks.

It's not even a nonce anymore, really. I'd change `CurrentTimeSecNonceFactory` to `ExpirationTimeFactory` and call `lag` something like `expirationTime` and comment it as being in seconds.

declan94:
Have modified the code according to your suggestions.

PR #2650, opened 2018-06-13
Foundations for better error handling in Bittrex and Livecoin

walec51:
Applying the new error handling scheme and improving consistency in accordance to https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices 

These are the foundations needed to implement https://github.com/knowm/XChange/issues/2533 and other more precise exceptions types


walec51:
I'm still writing some additional integration tests to make sure this does not break anything

could use some help in testing - if some one has an app that uses the Livecoin and/or Bittrex trade and account API please test it against this PR

walec51:
@timmolter would you be ok if I started to add wiremock based integrations tests like this one for authorized API calls:
https://github.com/knowm/XChange/pull/2650/files#diff-8c5490aeb5f76c54cea25b81f6769818

The advantages of them are:
- we can test our code using json's copied from exchanges documents which out setting up test account and storing api keys in our code base
- one can easily modify them to use the an actual account and connect to the real API if the need arises

timmolter:
Yeah, sure

pom.xml
```diff
@@ -271,6 +271,13 @@
 
 	<!-- All children will have these dependencies -->
 	<dependencies>
+            
+                <!--For String Joining, etc. -->
+                <dependency>
+                        <groupId>org.apache.commons</groupId>
```
timmolter: if not ALL modules need this, put it under dependencyManagement please.

xchange-bittrex/pom.xml
```diff
@@ -29,6 +29,13 @@
             <artifactId>xchange-core</artifactId>
             <version>4.3.8-SNAPSHOT</version>
         </dependency>
+        
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock</artifactId>
```
timmolter: put this under dependencyManagement too please.

pom.xml
walec51: its already there
removed the version from below

xchange-livecoin/src/main/java/org/knowm/xchange/livecoin/LivecoinAdapters.java
```diff
@@ -258,65 +266,45 @@ public static FundingRecord adaptFundingRecord(Map map) {
         null);
   }
 
-  public static List<Wallet> adaptWallets(List<Map> data) {
-    Map<Currency, WalletBuilder> wallets = new HashMap<>();
-    for (Map balance : data) {
-      String type = balance.get("type").toString();
-      String ccy = balance.get("currency").toString();
-      String value = balance.get("value").toString();
+  public static Wallet adaptWallet(List<LivecoinBalance> livecoinBalances) {
```
walec51: the previous implementation created a separate wallet for each currency with one balance
changed it to work as the generic API intended - Livecoin have no separate wallets

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/BittrexAdapters.java
```diff
@@ -245,7 +245,9 @@ public static Balance adaptBalance(BittrexBalance balance) {
     List<UserTrade> trades = new ArrayList<>();
 
     for (BittrexUserTrade bittrexTrade : bittrexUserTrades) {
-      trades.add(adaptUserTrade(bittrexTrade));
+      if (!isOrderWithoutTrade(bittrexTrade)) {
```
walec51: by writing a test with the example JSON from Bittrexes documentation I found out that our implementation has a bug and generates a trade in getTransactionHistory with price = 0 if there was an order which didn't buy / sell anything

xchange-livecoin/src/main/java/org/knowm/xchange/livecoin/LivecoinAdapters.java
```diff
@@ -235,7 +243,7 @@ public static UserTrade adaptUserTrade(Map map) {
         price,
         DateUtils.fromMillisUtc(Long.valueOf(map.get("date").toString())),
         id,
-        map.get("externalKey").toString(),
+        Optional.ofNullable(map.get("externalKey")).map(Object::toString).orElse(null),
```
walec51: in the documentation I see a "external" property, maybe in production its different
either way it can be nulled

pom.xml
timmolter: then it shouldn't be in `dependencies` as it makes it a dependency on all modules.

PR #2658, opened 2018-06-18
[Bitfinex] cancel all orders implementation

cymp:


cymp:
@timmolter some v2 calls use some java.org.knowm.xchange.bitfinex.v1 classes (like BitfinexException for example). I will work on creating a java.org.knowm.xchange.bitfinex.all package for common classes for both v1 and v2 versions if nobody does this before me.

PR #2665, opened 2018-06-21
Add maker taker fees API

BenjaminTrapani:
Adds API for fetching volume-based maker taker fees and updates many exchanges' metadata for worst case fees in the meantime. The only implementation of the make-taker fee API is in Bitfinex, but I'll add many more soon if the API is okay.


timmolter:
please sync up with current develop branch

timmolter:
How about using this endpoint: https://docs.gdax.com/#user-account

BenjaminTrapani:
I have some partially complete code that uses the linked endpoint as an initial step. Once we get the 30 day trailing volume, we need to compute the equivalent in USD at the current market rate. I think it is the case that for each currency pair, the base exists in at least one product that relates it to USD directly, so no extended conversion-rate calculations should be required. Is it okay if the caller to this function provides the most recent order book? We could overload this API with one version accepting an order book and the other with no parameters. The exchange implementations could provide a definition for the one they require (hopefully no order book), and the unimplemented one would throw an exception as expected. This would let developers automate the process of determining when to pass an order book into the API.

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/dto/account/BitfinexTradingFeeResponse.java
```diff
@@ -0,0 +1,60 @@
+package org.knowm.xchange.bitfinex.v1.dto.account;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.math.BigDecimal;
+
+@JsonIgnoreProperties({"maker_fees", "taker_fees"})
```
walec51: this annotations seams pointless

xchange-core/src/main/java/org/knowm/xchange/service/account/AccountService.java
```diff
@@ -125,4 +128,21 @@ default TradeHistoryParams createFundingHistoryParams() {
...
+   *     request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the
+   *     requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the
+   *     requested function or data, but it has not yet been implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  public default Map<CurrencyPair, Fee> getDynamicTradingFees() throws IOException {
+    throw new NotYetImplementedForExchangeException();
+  }
```
walec51: how many exchanges provide this type of endpoint?
I don't think this is a generic feature for now

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/dto/account/BitfinexTradingFeeResponse.java
BenjaminTrapani: It is necessary to prevent jackson from throwing errors, not to mention the fact that it improves performance.

xchange-core/src/main/java/org/knowm/xchange/service/account/AccountService.java
BenjaminTrapani: Virtually all exchanges asses fees as a function of trailing 30 day volume. The brackets and the way that the trailing volume is computed vary by exchange, which is why the API exists. Some exchanges provide an explicit way of querying this information, while for others it is required to derive it from trailing 30 day volume and published fee brackets.

timmolter: I think this is a needed addition.

timmolter:
please fix the conflicts, and I'll merge this. Thanks!

xchange-core/src/main/java/org/knowm/xchange/service/account/AccountService.java
walec51: ok

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v1/dto/account/BitfinexTradingFeeResponse.java
walec51: but the BitfinexTradingFeeResponse has no "maker_fees", "taker_fees" properties
the nested BitfinexTradingFeeResponseRow class has them and they are not ignored there

BenjaminTrapani:
Bummer, looks like some exchanges like Kraken have fee tiers per currency pair. I will update the changes to reflect this, since the current version associates a fee tier with an exchange.

PR #2675, opened 2018-06-27
Add missing currencies + rate limits + allow creation of orders from JSON

sebastienkoch:
Add missing currencies for some exchanges
Add rate limits for some exchanges
Allow creation of orders from JSON

### DIFF
--- a/xchange-binance/src/main/resources/binance.json
+++ b/xchange-binance/src/main/resources/binance.json
@@ -517,5 +517,12 @@
       "scale": 8,
       "withdrawal_fee": 0.02
     }
-  }
+  },
+  "public_rate_limits": [
+    {
+      "calls": 120,
+      "time_span": 60,
+      "time_unit": "seconds"
+    }
+  ]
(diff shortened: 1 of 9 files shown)


timmolter:
I don't want to make the DTOs mutable, hence the `final` keywords need to stay. Also, I don't want to allow for annotations like `@JsonIgnore`. These topics have been discussed before, not that I expect you to have seen them, and I'll refrain from writing out all the reasons here for now.

sebastienkoch:
It would be pratical to have DTOs mutable in order to be able to store them in JSON format, and then deserialise them when necessary (for exemple for testing purpose).

If you don't want to approve this modification, could you please at least commit the others?

timmolter:
You can serialize an immutable object to JSON. That should be no issue. 

No, I am not able to select out the changes that we'd like to keep. Can you do it from your end?

walec51:
@sebastienkoch will you be working on this PR?

PR #2681, opened 2018-07-02
[Upbit] Add

water0126:
add upbit exchanges.


xchange-upbit/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+	<!-- Parent provides default configuration for dependencies -->
+	<dependencies>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>xchange-core</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>com.auth0</groupId>
+			<artifactId>java-jwt</artifactId>
```
timmolter: I'd rather not have to introduce another dependency to XChnage. Can you use one that's already included in the parent pom?

water0126: Ah. Yes, I edited

FedrBodr:
@water0126 hello is you do support of trades services wrappers for make orders?

PR #2684, opened 2018-07-04
[dsx] fixed mapping of DSX currency pairs

andre77:
mapping is done using the meta information provided by DSX: https://dsx.uk/mapi/info


andre77:
the old implementation converted the pair "usdtusd" to "USD/TUS" :-)
now it works properly

PR #2694, opened 2018-07-09
[Bittrex] Market Order Support

traviscollins:
Bittrex no longer supports market orders. This is a work around
that submits Limit orders that are very likely to be taken up immediately.


cymp:
Please don't accept this PR:

1. Very dirty
2. Constructor change ruins backwards compatibility

Please find another way.

traviscollins:
I’ll address #2 by just adding a single argument backwards compatible constructor. 

What do you mean by #1?

cymp:
@traviscollins please see comments on PR https://github.com/knowm/XChange/pull/2695

traviscollins:
Thanks @cymp. It's important to note that the current code in the Bittrex XChange implementation for placeMarketOrder needs to be changed no matter what - as Bittrex no longer supports market orders (current code will perform a request and be rejected 100% of the time).

PR #2695, opened 2018-07-09
[Gemini] Market Order

traviscollins:
Gemini does not support market orders directly. This is a work around that places effective market orders by placing out priced limit orders that are very likely to be taken up immediately.

### DIFF
--- a/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/GeminiExchange.java
+++ b/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/GeminiExchange.java
@@ -23,7 +23,7 @@ public class GeminiExchange extends BaseExchange {
     concludeHostParams(exchangeSpecification);
     this.marketDataService = new GeminiMarketDataService(this);
     this.accountService = new GeminiAccountService(this);
-    this.tradeService = new GeminiTradeService(this);
+    this.tradeService = new GeminiTradeService(this, this.marketDataService);
   }
 
   @Override
--- a/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/service/GeminiTradeService.java
+++ b/xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/service/GeminiTradeService.java
@@ -1,41 +1,41 @@
 package org.knowm.xchange.gemini.v1.service;
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Date;
 import org.knowm.xchange.Exchange;
 import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.dto.Order;
+import org.knowm.xchange.dto.marketdata.Ticker;
 import org.knowm.xchange.dto.trade.LimitOrder;
 import org.knowm.xchange.dto.trade.MarketOrder;
 import org.knowm.xchange.dto.trade.OpenOrders;
 import org.knowm.xchange.dto.trade.UserTrades;
 import org.knowm.xchange.exceptions.ExchangeException;
-import org.knowm.xchange.exceptions.NotAvailableFromExchangeException;
 import org.knowm.xchange.gemini.v1.GeminiAdapters;
 import org.knowm.xchange.gemini.v1.GeminiOrderType;
 import org.knowm.xchange.gemini.v1.dto.trade.GeminiLimitOrder;
 import org.knowm.xchange.gemini.v1.dto.trade.GeminiOrderStatusResponse;
 import org.knowm.xchange.gemini.v1.dto.trade.GeminiTradeResponse;
+import org.knowm.xchange.service.marketdata.MarketDataService;
 import org.knowm.xchange.service.trade.TradeService;
-import org.knowm.xchange.service.trade.params.CancelOrderByIdParams;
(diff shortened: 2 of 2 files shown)


cymp:
Please don't accept this PR:

1. Very dirty
2. Constructor change ruins backwards compatibility (constructor is public...)

Please find another way.

traviscollins:
@cymp What do you mean by your first bullet, "very dirty"?

cymp:
Because:

1. It is a hack and not guaranteed to be allowed by the exchange (and doesn't work in many)
2. It is a hack and the logic is shady for maintenance
2. It is not guaranteed to work, especially in low volume and large spread markets (although it should most of the time)
3. Maybe the most important one: there is an embedded call to getTicker, making the placeMarketOrder make two calls. This is not what we want in this project. One of the practical reason is that 95% of the exchanges impose api calls rate limits, and it is not possible to control this rate with embedded calls.
5. Less important, but you should use BigDecimal.TEN, LimitOrder builder.

If you want to use this logic, I suggest that you should create your own placeMarketOrder method in your client code, which would call XChange's getTicker and placeLimitOrder methods.

traviscollins:
Thanks Cymp for your opinions and input. I generally disagree with most of 1-3, slightly agree with 4 (we already have to count calls outside the API anyway), and will check on 5. But I'll propose an alternative and ask @timmolter (or whoever is the review board) for input. 

The root of the problem is that there's no method through the XChange API to know which exchanges support Market Orders and those that don't. If a "supportsMarketOrders" flag was present on ExchangeMeta data, or perhaps CurrencyPairMetaData (but havn't seen a need for that personally) - logic like the above could be extracted into upstream code, and implemented separately for every user.  Otherwise, upstream code must delve deeper and create their own meta data systems. 

I'll note that we've been using code similar to this locally in an upstream project, and it's very effective (literally thousands of trades with no failures due to the logic intended in this pull request).

cymp:
Regarding your disagreement with 1-3: 
When I said this code could fail, here is a scenario: let's say we have a market with a tiny volume, manipulated by a bot. Someone buys at 1sat and then the bot is kicking in and create liquidity on ask and bid books, making the price rising although no trade happen. The price could go up to 20 sats with no trade still not happening. With your code, if you place a market buy, it will place a buy order at 10 sats, which is below the actual book price, and won't market buy at all. It would just place an order in the book sitting there at 10 sats: we really don't want this. This scenario looks stupid, but it is not the only possible one and remember that the crypto markets have 0 regulation, a lot of things can happen. 

I have noticed that you also applied this hack earlier in https://github.com/knowm/XChange/commit/4379846dbccde20461b031acffd097f425fa96c6#diff-99fce3ee282f23829d5b7f62f5684a59 as well. While you are looking for an alternative as you said, could you also please revert any place in the library where you committed this?
@timmolter maybe you want to give your opinion here

The way for XChange to notify that a method is not supported is to override it in the specific exchange implementation and write a one-liner throwing a NotAvailableFromExchangeException. This way, you just have to try catch on the placeMarketOrder call and your client code will know and call your custom placeMarketOrder.

traviscollins:
@cymp Thanks for the idea on the catching of an exchange exception - but we try to avoid those (very slow compared to a boolean check, etc). 

At the very least I'd advocate for a utility class that can be used to place "Market like orders as Limit Orders" - as it's a very useful thing to do (not a hack), and one that not everyone should have to recreate.

cymp:
If you don't want to catch an exception, you can catch it just once and store the result in a boolean for the following calls. This way you will suffer from the 0.001ms caused by the catch overhead just once.

timmolter:
XChange is intended to be an interface to an exchange, matching the exchange's API as directly as possible. So if the exchange doesn't have market orders, then the XChange module should throw `NotAvailableFromExchangeException` for `placeMarketOrder`. Any "simulated" market order using limit orders can and should be handled by your client code.

cymp:
@traviscollins FYI, following this confirmation from @timmolter, I created a PR reverted this same code in PoloniexTradeService.

traviscollins:
I think this is just a difference between view points - theoretically it’s great to that the library is bent to a slim posture. But in practice this is a feature that every serious trader needs - and what people actually need should be the focus of this library.

timmolter:
@traviscollins I totally understand where you're coming from. And it's tough for me to decide one way or another. In the end, those exchanges do not have market orders and your solution is what **you** need, but it's not safe to assume that that's what **everyone** needs. @cymp is one such person to give a concrete example. 

If all exchanges follwed the same play book and made their APIs exactly the same, then XChange would be the perfect library, but it's not the case, and client code is just going to have to handle the inconsistencies.

traviscollins:
The concept of “sell/buy X number of a coin right now, irrespective of cost” is absolutely a feature needed by the vast majority of traders. (Note, I’m speaking of people who trade coins as a job, not developers). And that is a necessity irrespective of whether the exchange supports traditional Market orders. 

The implementation details of that feature can be up for debate. But as @cymp noted, the details of implementing that feature vary on a per exchange basis. 

Our upstream code now has to extend every Xchange library exchange implementation with a custom interface for issuing market orders. This is code that every upstream project that services traders (again, not developers) will have to do.

traviscollins:
Deleted my previous comment because I doubt it will change minds. 

Tim thanks for what you’re doing here. It is important work, and a project from which I’ve received great benefit.

PR #2698, opened 2018-07-10
[binance] Fix setting priceScale from exchange metadata

mgrzaslewicz:


roeiba:
The suggested fix is not good enough, and it uses the precision of the tick as the price scale.

The correct way is to iterate all filters of a given symbol, and set price scale that holds ALL the filters. This is done by choosing the minimum price scale among all filters:

	Filter[] filters = symbol.getFilters(); 

	for (Filter filter : filters) {
	  switch (filter.getFilterType()) {
	  	case "PRICE_FILTER":
	 	  priceScale = Math.min(priceScale, numberOfDecimals(filter.getMinPrice()));
	  	  break;
	  	case "LOT_SIZE":
	  	  // In Binance, minimum amount is also minimum step size
		  // Remove all trailing zeros in order to get the real scale the amount  
	  	  minAmount = new BigDecimal(filter.getMinQty()).stripTrailingZeros();
	  	  maxAmount = new BigDecimal(filter.getMaxQty()).stripTrailingZeros();
		  break;
	  	case "MIN_NOTIONAL":
	  	  minNotional = new BigDecimal(filter.getMinNotional());
	  	  break;
	    }
	  }

Where number of decimals is:

	private int numberOfDecimals(String value) {
	    return new BigDecimal(value).stripTrailingZeros().scale();
	  }

badgerwithagun:
Aha, I've been suffering from this. Is anyone following up on this? If not, I will.

walec51:
@sebastienkoch @mgrzaslewicz will you be working on this PR?

mdvx:
> Aha, I've been suffering from this. Is anyone following up on this? If not, I will.

Yeah, I had to patch this too in my code

PR #2699, opened 2018-07-11
Fix Bitflyer last traded price

mlerner:
The last traded price for the Bitflyer ticker is null, even though it is present in the returned ticker data.


xchange-bitflyer/src/main/java/org/knowm/xchange/bitflyer/BitflyerAdapters.java
```diff
@@ -78,13 +78,15 @@ public static Ticker adaptTicker(BitflyerTicker ticker, CurrencyPair currencyPai
...
     BigDecimal volume = ticker.getVolume();
+    BigDecimal last = ticker.getLtp();
     Date timestamp =
         ticker.getTimestamp() != null ? BitflyerUtils.parseDate(ticker.getTimestamp()) : null;
 
     return new Ticker.Builder()
         .currencyPair(currencyPair)
         .bid(bid)
         .ask(ask)
+        .last(ask)
```
pefi9: Shouldn`t be there `.last(last)` instead of `.last(ask)`?

PR #2700, opened 2018-07-11
[BitFinex.v2] Add getTicker

Panchen:
- Add generic and raw v2 getTicker methods as well as getTickers.
- Bug fix MDS v2 vs v1 member instantiation in v2.remoteInit.


xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/BitfinexExchange.java
```diff
@@ -52,15 +48,15 @@ public ExchangeSpecification getDefaultExchangeSpecification() {
 
   @Override
   public void remoteInit() throws IOException, ExchangeException {
+    /* Exchange meta data not available in v2 calls
```
mlerner: Is there a reason not to delete this code?

Panchen: @mlerner - fair point. The initial implementation of v2 remoteInit() was cloned from v1. It uses v1 service objects rather than v2 so was hiding the new v2 raw market data methods being added. Unlike v1, BF v2 api doesn't provide exchange meta data - as far as I'm aware. Didn't like either of 2 obvious options of completely removing all trace of market data initiation or mixing v1 and v2 implementations (of the same MarketDataService object). v2 needs more work. My pull request was to get basic market data working in v2 for myself and others to finish implementation.

Issue #2703, opened 2018-07-12
BItstamp Funding history request Error


yurivin:
library creates request with double slash in the end: https://dev-publicapi.cryptonit.net/api/v2/user_transactions//

So Bitstamp understands this request like try to get trading pair related transactions but not funding history and throws error: {"status":"error","reason":{"MARKET_UNKNOWN":[]}}"; line: 1, column: 1]

PR #2704, opened 2018-07-12
Issue #2703. Bitstamp Funding request bug fix

yurivin:
library creates request with double slash in the end: https://dev-publicapi.cryptonit.net/api/v2/user_transactions//

So Bitstamp understands this request like try to get trading pair related transactions but not funding history and throws error: {"status":"error","reason":{"MARKET_UNKNOWN":[]}}"; line: 1, column: 1]

Solution is to use implementation of web service call without currencyPair parameter,


timmolter:
It looks like the currency pair is part of the API.

https://www.bitstamp.net/api/

Maybe there's some other issue?

yurivin:
I working with this API. So if I need to have history of deposit and withdrawal transactions I should not send currecyPair to the API call. And it works ok and I get funding history in response. In other case? if I send currencyPair I get only trading transactions history for specified currencyPair and NO funding history at all. And we have related error which I described above.

Also, I see that in BItstampAuthenticatedV2 class line 52 already exists call for this url without currencyPair and without second slash. Only BItstampAccountService.getFundingHistory uses incorrect api call. So I fixed it and tested.

yurivin:
https://www.bitstamp.net/api/

In API documentation also specified two variants of this API function call with and without currencyPair. Search by https://www.bitstamp.net/api/v2/user_transactions/

PR #2711, opened 2018-07-19
Update BitfinexPayloadDigest.java

tiagoferraro:
suggested change according to the bitfinex documentation


cymp:
Hello, could you please detail?

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/common/service/BitfinexPayloadDigest.java
```diff
@@ -1,6 +1,6 @@
 package org.knowm.xchange.bitfinex.common.service;
 
-import java.util.Base64;
+import android.util.Base64;
```
walec51: -1
this library is not only for android

PR #2719, opened 2018-07-24
[independentreserve] Currency mapping 

andre77:
there are additional currencies to Xbt and Eth like: Bch, Xrp, Ltc.
the mapping should be done outside of xchange code, otherwise every new currency will require the code to be modified...


nielsdraaisma:
Removing this mapping would require all existing users of this library to add this code, which might not be trivial. I expect many users to work with various exchanges, if I have to start making different calls to various implementations that drastically reduces the value of this library.

It's painful to have to change the code when a new currency is added and the currency code doesn't match, but it's a price worth paying for having a consistent API.

PR #2723, opened 2018-07-26
[quadrigacx] added withdrawal support for additional currencies: BCH, BTG, LTC

andre77:
due to the quadriga API design, one needs to add extra method for every currency, unfortunatelly...

### DIFF
--- a/xchange-quadrigacx/src/main/java/org/knowm/xchange/quadrigacx/QuadrigaCxAuthenticated.java
+++ b/xchange-quadrigacx/src/main/java/org/knowm/xchange/quadrigacx/QuadrigaCxAuthenticated.java
@@ -100,6 +100,30 @@ public interface QuadrigaCxAuthenticated {
       throws QuadrigaCxException, IOException;
 
   @POST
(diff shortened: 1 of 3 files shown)


walec51:
use PathParam and implement come `currency to api name` mapping instead of adding a method for each crypto

andre77:
@walec51 thanks for your comment. yes it could be an option
for now it was faster for me to keep the structure of methods 1:1
if quadriga adds 10 more currencies with 2x10 more methods, one should use your approach

walec51:
to me this is just lazy copy-paste style coding and a bad practice that will be followed in the code
please improve the implementation, the point of code review is not to accept anything that works

andre77:
@walec51 thanks for your comment.
i'm fine with the current implementation, it works und the way to provide methods in the interface 1:1 mapped to those from the API description is not the worst option, even though i do not like the c&p style. its not me, who copy&pasted the methods, its the devs by quadriga in this particular case.
feel free to improve it though and implement your suggestion

Issue #2337, opened 2018-02-21
Port to java9


rafalkrupinski:
1. [x] make RestProxyFactory mockable
2. [ ] change tests so they don't require PowerMockito
    - [ ] acx
    - [x] anx
    - [x] bitmarket
    - [ ] bleutrade
    - [ ] btcmarkets
3. [ ] upgrade maven in travis
4. [ ] add java9 in travis

@timmolter please assign to me

walec51:
are we planning to move to java.time to?

rafalkrupinski:
@walec51 as much as I'd love to ditch that awful class from all java, what would be the reason to make that effort and break compatibility?

timmolter:
@rafalkrupinski Any more progress here? I feel this is the most pressing change needed at the moment. Can I help in some way?

rafalkrupinski:
I've no work in progress at the moment, feel free to handle any of the above exchanges

walec51:
successfully complied all modules on jdk10 with these changes: https://github.com/knowm/XChange/pull/2731

PR #2731, opened 2018-07-29
jdk 10 support, livecoin NPE fix, mavens formatting fix

walec51:
* this PR makes XChange compile under jdk 10
* also fixed one NPE caused by an anomaly in Livecoins API
* maven forces some formatting changes on many files I did not edit during the build


walec51:
related issue: https://github.com/knowm/XChange/issues/2337

PR #2734, opened 2018-08-02
Let account service to get all coin adresses as API allows

baboono:


walec51:
thanks for the PR, looks ok, *ServiceRaw classes can expose exchange specific features

what worries me is that despite the fact that maven corrects code formatting during build you managed to push badly formatted code - did you build and test it?

PR #2735, opened 2018-08-02
Will let developer bypass encapsulation and use more functionalities …

baboono:
…of Poloniex APi


walec51:
if you want access to exchange specific features then cast a *Service to *RawService

Issue #2737, opened 2018-08-06
Official Cryptonit Public API Updated On


bitcoinmaniac:
Cryptonit.net changed ralisation of it's official public API and provides library Update. Old API is no longer official supported and depricated. 

Pull request: https://github.com/knowm/XChange/pull/2738

bitcoinmaniac:
This is official cryptonit.net FINOM team pull requset. 
If you have any doubts, please write here or at yurij.vinogradov@finom.tech

PR #2738, opened 2018-08-06
Cryptonit Public API Updated On. Issue #2737

bitcoinmaniac:


bitcoinmaniac:
java-websocket mentioning excluded from pom

PR #2744, opened 2018-08-07
[gateio] Added support for funding history

andre77:
added support for funding history (deposits / withdrawals)
additionally thrown away the nonce factory stuff for gate.io, the API does not require any nonce parameter


andre77:
additionally added support for withdrawal execution

PR #2765, opened 2018-08-20
[independentreserve] fixed funding history

andre77:
see conversation https://github.com/knowm/XChange/pull/2739


andre77:
additionally added the support for currency parameter, which was not passed before, see https://github.com/knowm/XChange/pull/2765/commits/0ce9ea88cb1e76a7bbdf9fb0acfece8137e42f1f

actually a funny thing, BTC and ETH withdrawals were supported only, since the currency parameter is optional, see the doc:
[primaryCurrencyCode: Optional primary currency code. If not passed API will interpret an address as Bitcoin or Ether depending on address passed.](https://www.independentreserve.com/API#WithdrawDigitalCurrency)

PR #2773, opened 2018-08-24
[Bittrex] fixing of classes

theo1985:
the JSON conversion wasn't happening, so cleaned up the code


xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/trade/BittrexOrder.java
```diff
@@ -0,0 +1,224 @@
...
+  public String getCondition() {
+    return condition;
+  }
+
+  public String getConditionTarget() {
+    return conditionTarget;
+  }
+
+  @Override
+  public String toString() {
```
walec51: toString seams no needed - may be lets remove them in those dtos?
such long ones don't help in debuging in any way and there not used in any other context

xchange-itbit/src/main/java/org/knowm/xchange/itbit/v1/ItBitAdapters.java
```diff
@@ -69,7 +68,7 @@ private static DecimalFormat getCryptoFormat() {
     cryptoFormat.setDecimalFormatSymbols(CUSTOM_SYMBOLS);
     cryptoFormat.setMaximumFractionDigits(4);
     cryptoFormat.setGroupingUsed(false);
-    cryptoFormat.setRoundingMode(RoundingMode.HALF_UP);
+    cryptoFormat.setRoundingMode(RoundingMode.UNNECESSARY);
```
walec51: -1
this will cause ArithmeticException's
https://docs.oracle.com/javase/7/docs/api/java/math/RoundingMode.html

walec51:
rest looks ok

xchange-itbit/src/main/java/org/knowm/xchange/itbit/v1/ItBitAdapters.java
walec51: ok, but this will cause 0.001000034 to also throw an exception
don't you think that a `HALF_DOWN` and a `if(someDecimal.compareTo(MIN_AMOUNT) < 0) throw new ItBitException` would be better then relaying on the formatter for validation?

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/trade/BittrexOrder.java
theo1985: I'm happy to remove it, if that's the design for dtos

theo1985: Removed

xchange-itbit/src/main/java/org/knowm/xchange/itbit/v1/ItBitAdapters.java
theo1985: Truncating extra precision

PR #2776, opened 2018-08-27
[btcmarkets] - Changed type of BTCMarketsPlaceOrderResponse.id from Int to Long matching API return values

nielsdraaisma:
Increasing the size of the Id field on the placeOrder return, matching the current API returned values.


walec51:
+1 looks ok

nielsdraaisma:
Would be great to get this in, i suspect if affects all users of btcmarkets and it's a small change.

PR #2783, opened 2018-08-29
 fix order amount precision error for huobi 

desinerlee:


xchange-huobi/src/main/java/org/knowm/xchange/huobi/service/HuobiTradeServiceRaw.java
```diff
@@ -78,7 +78,7 @@ String placeHuobiLimitOrder(LimitOrder limitOrder) throws IOException {
                 String.valueOf(
                     ((HuobiAccountServiceRaw) exchange.getAccountService())
                         .getAccounts()[0].getId()),
-                limitOrder.getOriginalAmount().setScale(4, BigDecimal.ROUND_DOWN).toString(),
+                limitOrder.getOriginalAmount().toString(),
```
walec51: toString can return scientific notation if an exponent is needed
I doubt that  Huobis API can handle that

timmolter:
what is unit amount exactly?

xchange-huobi/src/main/java/org/knowm/xchange/huobi/service/HuobiTradeServiceRaw.java
desinerlee: The originalAmount parameter in the huobis API has different precision requirements for different currency pairs, so the precision of the originalAmount parameter should be provided by the caller.

desinerlee: If you set the wrong precision, you cannot place an order.

desinerlee:
@timmolter In some futures trading scenario, the originalAmount parameter represents the number of contracts, and the unitAmount parameter represents the par value of each contract, so that the transaction cost and profitability can be calculated through the originalAmount parameter and the unitAmount parameter.
For an example, See the OkCoin documentation and search for "unit_amount": https://github.com/okcoin-okex/API-docs-OKEx.com/blob/master/API-For-Futures-EN/REST%20API%20for%20FUTURES.md

timmolter:
You don't use `getUnitAmount()` anywhere for fixing the precision error. Please remove that from the high level API (`Order`) for now. We need to look closely at that addition and there's maybe a better solution. Maybe we need place/cancel Future methods.

Issue #2766, opened 2018-08-21
gdax pagination implemented?


IanWorthington:
According to https://docs.pro.coinbase.com/#get-trades the request is paginated but I'm not seeing how to use this in xchange.

Is this actually implemented?

madwicket:
Yes, it is implemented.
If you look in GDAX.java it is here:

  @GET
  @Path("products/{baseCurrency}-{targetCurrency}/trades")
  GDAXTrade[] getTrades(
          @PathParam("baseCurrency") String baseCurrency,
          @PathParam("targetCurrency") String targetCurrency)
          throws GDAXException, IOException;

IanWorthington:
OK, I'm pretty sure, though totally surprised, that this has never been implemented.  I'll take a look at it.

PR #2784, opened 2018-08-30
Add support for GDAX paginated getTrades

IanWorthington:


IanWorthington:
This is associated with https://github.com/knowm/XChange/issues/2766

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
```diff
@@ -90,7 +97,74 @@ public GDAXProductBook getGDAXProductOrderBook(CurrencyPair currencyPair, int le
...
+		    	log.debug("Rate limit exceeded, sleeping for 1000ms");
+		    	
+		    	rateLimited = true;
+		    	
+		    	try {
+					Thread.sleep(1000);
+				} catch (InterruptedException e1) {
+					// TODO Auto-generated catch block
+					e1.printStackTrace();
+				}
```
walec51: was this PR build using maven before pushing?
seams that it was not because maven would fix the formatting

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
```diff
@@ -90,7 +97,74 @@ public GDAXProductBook getGDAXProductOrderBook(CurrencyPair currencyPair, int le
...
+	      return gdaxTrades;
+	    	
+	    } catch (GDAXException e) {
+	    	if (e.getHttpStatusCode() == 429) {
+		    	log.debug("Rate limit exceeded, sleeping for 1000ms");
+		    	
+		    	rateLimited = true;
+		    	
+		    	try {
+					Thread.sleep(1000);
```
walec51: this should not be done at this level, remove the sleep calls and do call rate limiting in the client code

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataService.java
```diff
@@ -51,10 +56,67 @@ public OrderBook getOrderBook(CurrencyPair currencyPair, Object... args)
...
+			  latestTradeId = gdaxTrades.getEarliestTradeId();
+			  
+			  if (gdaxTradesNew.getEarliestTradeId() == null) {break;}
+			  if (gdaxTrades.getEarliestTradeId() <= fromTradeId) {break;}
+		  }
+		  
+		  log.debug("earliest-latest: {}-{}", gdaxTrades.getEarliestTradeId(), gdaxTrades.getLatestTradeId());
+		  
+		  if (log.isDebugEnabled()) {
+		    gdaxTrades.stream().forEach(System.out::println);
```
walec51: why do you use `System.out` and `e.printStackTrace' in the code if you have logging available?

xchange-gdax/pom.xml
```diff
@@ -29,7 +29,14 @@
...
             <version>${project.version}</version>
         </dependency>
-
+        
+        <dependency>
+        	<groupId>org.projectlombok</groupId>
+        	<artifactId>lombok</artifactId>
+        	<version>1.18.2</version>
+        	<scope>provided</scope>
+        </dependency>
```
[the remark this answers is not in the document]
IanWorthington: Thanks for your comments.  We established a little while back that lombok was to be permitted but not required.  You might want to check out the issue for it.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
IanWorthington: Not, because the 'format' keyword doesn't appear to be recognised under my eclipse.  If there's a way to resolve that I'm happy to do so.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
IanWorthington: Afraid I disagree with this.  Rate limiting is a lower level function, ideally lower than this, but until we can hook it into the rescu package this is the best we can do.  Client code should NOT have to be concerned with the gdax rate limits I think.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
walec51: ok

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataService.java
IanWorthington: I'm torn here.... In this case just because it improves the readability of the log and is only for debugging to the console.  Though it prevents redirection of course.  I'm not fussed if you want me to change it.

I'll have to check out if there's a better way of handling the printstacktrace.

walec51: I don't understand whats your point
SLF4J prints stracktraces when you do
`logger.debug("Some message about the context", exception);

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
IanWorthington: All that is certainly true. But, rather surprisingly, this sketchy solution actually works in testing rather well.

I'm not aware of any other method that specifically needs to fire large numbers of calls at gdax.

walec51: I'm not claiming it does not work, if it didn't I would not call it a 'solution' :o)
getGDAXProductOrderBook, getGDAXTrades, getGDAXHistoricalCandles
and similar methods in every other module 
we need a global solution, not case by case boiler code

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
IanWorthington: You're quite right, it did.  Fixed.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataService.java
IanWorthington: Fixed.

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
IanWorthington: Adam,  if I change this then the client has to be responsible for not only all retries but also chaining the results of all the different calls together.  This really does seem an unnecessary load to place on the client code.

I appreciate your point of view about needing a global solution, but until we get one then case-by-case boilerplate is the best we can do.

I really am very reluctant to remove this.

IanWorthington:
@walec51 

Adam: the problem with the build error has swapped over from 10 to 8!  Would you mind giving it another poke?  Thanks.

xchange-gdax/pom.xml
timmolter: The dependency should be added to the parent pom under dependencyManagement

xchange-gdax/src/main/java/org/knowm/xchange/gdax/service/GDAXMarketDataServiceRaw.java
timmolter: I agree that client code should be responsible to retries. We've had conversations about this in other threads. Mainly, the scope of XChange is to provide gateways to all the REST endpoints. One sorta global solution that clients can use is the `Retries` class, but I haven't personally used it yet. Can you try it out?

xchange-gdax/pom.xml
IanWorthington: @timmolter 

Tim -- I'm unsure how to do this correctly.  I've moved the dependency to the parent pom but I'm unable to build it either under eclipse or pushing it, maybe because the gdax pom points to the parent 4.3.10-SNAPSHOT, not to my new parent.

But if I change that, presumably I break something else?  What's the correct way to proceed here please?

Thanks.

timmolter:
Take a look at any of the other module's poms with a dependency. You need to define it in both parent and module, but the module one should have no version defined.

IanWorthington:
@timmolter -- Thanks Tim.

pom.xml
```diff
@@ -287,7 +288,13 @@
 				<artifactId>logback-classic</artifactId>
 				<version>${version.qos.logback}</version>
 			</dependency>
-
+			
+			<dependency>
+	        	<groupId>org.projectlombok</groupId>
+	        	<artifactId>lombok</artifactId>
+	        	<version>${version.lombok}</version>
+	        	<scope>provided</scope>
```
timmolter: Why provided?

IanWorthington: lombok is compile-time only.  See https://projectlombok.org/mavenrepo/

Issue #2668, opened 2018-06-21
Change GDAX endpoint to api.pro.coinbase.com


malkusch:
Coinbase just sent an email that api.gdax.com became deprecated and should be changed to api.pro.coinbase.com until end of this year:

> GDAX APIs will continue to function until December 31, 2018. We encourage Coinbase Pro users to begin migrating their API usage away from api.gdax.com to api.pro.coinbase.com.

They didn't mention any API change nor does their documentation even point to that new endpoint.

Edit: It seems that they want to rebrand GDAX into Coinbase Pro. Maybe it's also appropriate to rename the complete package as well, as GDAX won't have any meaning in the future.

EricAnderson1000:
I was thinking about taking this on. I believe the creation of a cloned gdax project named xchange-coinbasepro maintained separately is the way to go. Coinbase is treating this as a different API (even though it is really just cloned at this point). But we should maintain them separately. This also allows users of XChange to choose when they use the new API rather than being forced with new XChange updates.

Thoughts? Shall I proceed?

malkusch:
I don't think a clone is necessary. From what I understand it's a rebranding. API changes do happen on both hosts now. The old host won't work beginning with 2019, so there's really no need to keep that after that point. Currently both hosts do work, which makes now a good point to migrate to the new host.

EricAnderson1000:
I would like to see the Classes and module renamed at the very least. But yes good point, if it's just a rebranding, everything should remain backwards compatible. If that is wrong, we will be potentially breaking users of GDAX API vs CoinbasePro's

Changing the API on the class is a very easy change. Why hasn't this already been done?

Are you in agreement about the naming (packages, classes, etc)?

malkusch:
I'm not the owner here, so what I say has no more weight than anybody else's opinion. But my opinion is first just change the host name of the endpoint. This shouldn't take to much ceremony and is required for a release before the end of this year. The earlier the better.

Changing class names and packages is a cosmetic change and can happen at any point later, preferably with duplicating them first and deprecating the old names to give clients not too much headaches.

PR #2793, opened 2018-09-11
Adapting GDAX code to rebranding effort --> CoinbasePro

EricAnderson1000:
See https://github.com/knowm/XChange/issues/2668

This change renames the GDAX project to CoinbasePro, deprecates the old code (backwards compatible).


IanWorthington:
@EricAnderson1000 

I mean maybe just change gdax to inherit it so it's not a breaking change down the road?  I like that both will continue to continue to coexist for now though.  Maybe that's enough if we mark gdax as deprecated?

EricAnderson1000:
@IanWorthington 

If there is enough support for it, I could make all the GDAX classes extend from the CoinbasePro ones so that any updates would automatically get updated to both. 

However it seemed like quiet a bit of more work only to delete the whole module in a few months. They close the GDAX api URI and the end of this year.

timmolter:
I'm for deprecating the gdax module and moving all updates to the CoinbasePro module, and deleting the GDAX module ASAP.

PR #2799, opened 2018-09-13
[CoinbasePro] Merging - Add support for GDAX paginated getTrades #2784 - from gdax

EricAnderson1000:
@timmolter @IanWorthington

Merging changes from GDAX Pull Request https://github.com/knowm/XChange/pull/2784 into CoinbasePro

After this change we could consider GDAX branch dead and merge new features into CoinbasePro

### DIFF
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProMarketDataServiceRaw.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProMarketDataServiceRaw.java
@@ -1,8 +1,12 @@
 package org.knowm.xchange.coinbasepro.service;
 
 import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+import lombok.extern.slf4j.Slf4j;
 import org.knowm.xchange.Exchange;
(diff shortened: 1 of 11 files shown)


xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProMarketDataServiceRaw.java
```diff
@@ -93,6 +98,60 @@ public CoinbaseProProductBook getCoinbaseProProductOrderBook(CurrencyPair curren
     }
   }
 
+  static Instant lastCall = null;
```
EricAnderson1000: I'm not sure this is a thread safe way of accomplishing this. But maybe we solve this in another Merge Request since this is how we did this on GDAX

IanWorthington:
Thanks Eric/Tim.

I totally agree we need a better way of handling retries. But until we do this is better than what we did have. 

I might rework it into an aspect, which has proven useful to me in handling database retries.  But I'm busy with some other stuff right now so that will have to wait for a little bit.

Comment about not being thread-safe noted.  I think that will make it perform non-deterministically but it will still catch the errors and back off, so it's safe.  Will fix though, thanks.

PR #2801, opened 2018-09-14
[Bitmex] Add missing order parameters

SaschaZ:
This adds all missing order parameters with a more simpler API.
I also modified the error handling a little bit to get more precise error messages from Bitmex.


timmolter:
Please fix conflicts and move your gdax changes over to the coinbasepro module, since the gdax module is deprecated.

PR #2802, opened 2018-09-16
[bitstamp] Fix: order ids have exceeded integer max value. Changed to longs

joakjo:
Bitstamp orders have exceeded [Integer.MAX_VALUE](https://docs.oracle.com/javase/7/docs/api/constant-values.html#java.lang.Integer.MAX_VALUE) leading to 

``` java
Caused by: org.knowm.xchange.bitstamp.dto.BitstampException: null (HTTP status code: 200)
at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[na:1.8.0_151]
at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[na:1.8.0_151]
at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[na:1.8.0_151]
at java.lang.reflect.Constructor.newInstance(Constructor.java:423) ~[na:1.8.0_151]
at com.fasterxml.jackson.databind.introspect.AnnotatedConstructor.call(AnnotatedConstructor.java:124) ~[jackson-databind-2.9.2.jar!/:2.9.2]
at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.createFromObjectWith(StdValueInstantiator.java:274) ~[jackson-databind-2.9.2.jar!/:2.9.2]
at com.fasterxml.jackson.databind.deser.ValueInstantiator.createFromObjectWith(ValueInstantiator.java:228) ~[jackson-databind-2.9.2.jar!/:2.9.2]
at com.fasterxml.jackson.databind.deser.impl.PropertyBasedCreator.build(PropertyBasedCreator.java:189) ~[jackson-databind-2.9.2.jar!/:2.9.2]
at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeUsingPropertyBased(BeanDeserializer.java:487) ~[jackson-databind-2.9.2.jar!/:2.9.2]
at com.fasterxml.jackson.databind.deser.std.ThrowableDeserializer.deserializeFromObject(ThrowableDeserializer.java:65) ~[jackson-databind-2.9.2.jar!/:2.9.2]
at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:159) ~[jackson-databind-2.9.2.jar!/:2.9.2]
[...]

PR #2804, opened 2018-09-17
[bitstamp] order ids do not fit into 32 bit integers

andre77:


andre77:
oh, there is already a PR for that https://github.com/knowm/XChange/pull/2802
little difference though, i have changed the withdrawal and transaction ids to long as well

PR #2805, opened 2018-09-18
Unused logback.xml removal

SergeyLukashevich:
Remove Logback configuration files from a project, since they are not needed
in a library and raise conflicts in the projects where it's used.


walec51:
+1

timmolter:
I already moved the bit-z one. The examples one should stay.

PR #2806, opened 2018-09-18
[Bitmex] Add order modify

SergeyLukashevich:
Add `modifyOrder` to org.knowm.xchange.service.trade.TradeService, because many exchanges are able either modify existing order or cancel/replace it;
Add order modify for Bitmex.


SergeyLukashevich:
> cancel/replace is not the equivalent of modifying an order because the order id changes then you cancel/replace
> what exchanges support this besides Bitmex?

As far I know, order modification as is, exists only on Bitmex, but other exchanges allow cancel/replace by one request, for instance: 
Cex.io: https://cex.io/rest-api#cancel-replace-order]
ANX: https://anxv3.docs.apiary.io/#reference/trading/ordernew
Bitfinex: https://docs.bitfinex.com/v1/reference#rest-auth-replace-order

Since a new method can be implemented for modifying (when order ID doesn't change) or cancel/replace (ID is changing), it returns the respective ID, that allows use it for both operations.

As an option, the method can be renamed to `orderReplace` or `orderCancelReplace`, and in case of Bitmex it acts as modify.

walec51:
canceling and creating a new order can already by done using the existing API and is the most generic way of achieving what you want - doing it in one req is a neglectable optimization if one wants to use generic code

adding methods to generic interfaces that only work for single exchanges is not the way to go

just use the method BitmexTradeServiceRaw.replaceOrder in your app if you need it

SergeyLukashevich:
I see few benefits in the atomic cancel/replace:

* Cancel/replace performed by one request has minimal time between these operations. Of course, all exchanges are blackboxes for me, but I can suppose that in "good" exchanges cancel/replace is performed by one transaction or at least in minimal time.
* IMHO order modify or cancel/replace is the most used operation in trading. In many strategies I saw, 95% of all operations was limit order's price change. Eventually, when you make 1 request instead of 2, you seriously reduce traffic, and won't exceed exchange request rate limit, that most likely is a reason why many exchanges have this functionality.

walec51:
generic interfaces should have features that are generic -at least 3/4 of exchanges have them

there is no sense in having generic interfaces with implementations that mostly throw unsupported operation exception

if you want to optimize then use methods from *ServiceRaw classes

SergeyLukashevich:
Absolutely agree, don't make sense to add generic function for only one exchange.

After long and boring googling there is a list of such exchanges:

Cancel/Replace

Cex.io: https://cex.io/rest-api#cancel-replace-order]
ANX: https://anxv3.docs.apiary.io/#reference/trading/ordernew
Bitfinex: https://docs.bitfinex.com/v1/reference#rest-auth-replace-order
Poloniex: https://poloniex.com/support/api/#rest_trading (see "moveOrder")

Modify:
Bitmex: https://www.bitmex.com/api/explorer/#!/Order/Order_amend
Cobinhood: https://cobinhood.github.io/api-public/#modify-order
Quoine: https://developers.quoine.com/#edit-a-live-order

I expected more but seems there is a sufficient amount.

walec51:
I've added a issue describing your feature request https://github.com/knowm/XChange/issues/2818 and a page where we can list generic feature candidates like this one to research which exchange APIs support them: https://github.com/knowm/XChange/wiki/Generic-feature-candidates

walec51:
+1 PR is in accordance with our conclusion at https://github.com/knowm/XChange/issues/2818

build error is not related to changes in this PR, java 11 build is broken on the develop branch and requires this fix: https://github.com/knowm/XChange/pull/2916

PR #2813, opened 2018-09-25
[Binance] Do not silence Exception on remoteInit

theo1985:


walec51:
+1 but
remove the commented code and pass the exception as the cause like

```
throw new ExchangeException("Failed to initialize", e);
```

PR #2820, opened 2018-09-28
Using lombok getters and builders

EricAnderson1000:
I noticed that project lombok was introduced to this project some time ago. I believe using lombok annotations could greatly reduce boiler plate code. Also make it much easier to create builders which create immutability but also allow for easier use than just a constructors.

How the the XChange community feel about using these?

### DIFF
--- a/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/trade/BittrexOpenOrder.java
+++ b/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/trade/BittrexOpenOrder.java
@@ -2,7 +2,13 @@ package org.knowm.xchange.bittrex.dto.trade;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigDecimal;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.ToString;
 
+@Getter
+@Builder
+@ToString
 public class BittrexOpenOrder {
 
   private final String uuid;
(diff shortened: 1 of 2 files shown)


IanWorthington:
@EricAnderson1000 

Hi Eric --

I use, and really like parts of lombok but I believe that this passion is not universally shared.  I would thus be nervous about rolling out changes which only add in lombok in case it upsets major contributors/maintainers of packages.  What you chose to do with your own is another matter of course. Likewise in the event of major work.

Thoughts?

EricAnderson1000:
@IanWorthington 

Thanks for responding. Yes this PR is an attempt to gauge the opinion of the group in the matter.
IMHO

Pros:
To the end user of the library it should be be seem less right? Other than when investigating source code, they wouldn't know. 
Code is more concise and easier to read without magic that gets in the way.

Cons:
IntelliJ and some other IDEs require plugins for the annotation processing to occur correctly. 

I would be curious to hear any other pros and cons.

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/trade/BittrexOpenOrder.java
```diff
@@ -2,7 +2,13 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigDecimal;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.ToString;
 
+@Getter
+@Builder
+@ToString
```
walec51: Builder
ToString

seam not needed here

we only use the constructor for initialization (actually the Jackson library does) in DTOs
toString is only used for debuging and printing the entire object is no better then just browsing it in the debuggers UI
therefore I think we should avoid it in DTO's all together unless you can think of a sort string representation of the object that would be more helpful then the debuggers UI

IanWorthington:
> @IanWorthington
> 
> Thanks for responding. Yes this PR is an attempt to gauge the opinion of the group in the matter.
> IMHO
> 
> Pros:
> To the end user of the library it should be be seem less right? Other than when investigating source code, they wouldn't know.
> Code is more concise and easier to read without magic that gets in the way.
> Few chances for errors since getters and/or setters and Builders are generated.
> 
> Cons:
> IntelliJ and some other IDEs require plugins for the annotation processing to occur correctly.
> 
> I would be curious to hear any other pros and cons.

There's a big thread going over the pros and cons in Issues.  Given that after months we finally reached a decision on this do we really want to start debating it again?

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/trade/BittrexOpenOrder.java
IanWorthington: What's "seam"?

I don't think Builder is used in this example.

ToString is definitely worth having though.  I certainly use it a printed representation in trace and debug statements all over the place.

And, as a principle, lomboking a module should not change anything that previously existed.

EricAnderson1000: Yes I added Builder. My frustration came when needing to create this object for unit testing, I had to provide every property to the constructor, when functionally it wasn't required for the test.

It seems that this holds true for this object. Every property is required to construct, when it is just a POJO. This leads to putting in null values. Which should be avoided as well.

EricAnderson1000:
>There's a big thread going over the pros and cons in Issues. Given that after months we finally reached a decision on this do we really want to start debating it again?

Ah thanks. I wasn't even aware of that discussion. However it looks like the end result of that discussion was the following correct? 

>timmolter 
>Well after leaving this open for discussion for some time now, it looks like the majority vote for having lombok as an option, but not as a requirement. Feel free to add it and start using it then...

https://github.com/knowm/XChange/issues/2342

EricAnderson1000:
Closing this issue until there is more general acceptance of project lombok.

timmolter:
I accepted a PR a few weeks ago that used lombok for slf4j. It was the first time someone submitted a PR with Lombok since we decided it would be allowed. I was then inconvenienced with errors in Eclipse and had to figure out how to install Lombok to make Eclipse happy again. I feel stuff like this will turn off new-comers to Xchange who get frustrated just getting the dev environment setup, since even Maven is hard enough to wrap your head around in the beginning. That was why I originally hesitated and was against Lombok and my recent experience just confirmed my fear.

Having said all that, we did agree that Lombok is to be accepted into this project based on a issue #2342, so don't hold back at this point. I hope that the proponents of Lombok can eventually show how beneficial it is.

PR #2825, opened 2018-10-10
Updates and fixes for Cryptonit exchange.

bitcoinmaniac:


timmolter:
do you really want to make public those API keys? You should use a json file as described here: https://github.com/knowm/XChange/wiki/Frequently-Asked-Questions#json-file-on-classpath
, and don't check in the json file to git.

bitcoinmaniac:
API keys issue fixed

PR #2839, opened 2018-10-19
[Liqui] Unplugged Liqui getAccountInfo

cymp:
The balances cannot be retrieved in one API call. It is better to return nothing instead of returning wrong balances.


cymp:
@timmolter Looks like there is a Travis issue, would you please have a look?

PR #2853, opened 2018-10-30
Bit z

belerweb:


xchange-bitz/src/main/resources/logback.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
timmolter: remove this file or put it in `test`

PR #2863, opened 2018-11-14
[Coinbene] add symbols to market data

ivelichkevich:


xchange-coinbene/src/main/java/org/knowm/xchange/coinbene/CoinbeneExchange.java
```diff
@@ -39,10 +42,16 @@ public ExchangeSpecification getDefaultExchangeSpecification() {
...
   }
 
   @Override
-  public void remoteInit() throws IOException, ExchangeException {}
+  public void remoteInit() {
+
+    try {
+      exchangeMetaData = ((CoinbeneMarketDataService) marketDataService).getMetadata();
+    } catch (IOException e) {
+      LOG.error("Retrieving MetaData for Coinbene failed", e);
```
timmolter: This shouldn't be just logged. It should let the IOException through.

xchange-coinbene/src/main/java/org/knowm/xchange/coinbene/CoinbeneExchange.java
```diff
@@ -46,12 +46,12 @@ public ExchangeSpecification getDefaultExchangeSpecification() {
...
 
   @Override
-  public void remoteInit() {
+  public void remoteInit() throws IOException, ExchangeException {
 
     try {
       exchangeMetaData = ((CoinbeneMarketDataService) marketDataService).getMetadata();
     } catch (IOException e) {
-      LOG.error("Retrieving MetaData for Coinbene failed", e);
+      throw new IOException("Retrieving MetaData for Coinbene failed", e);
```
timmolter: remoteInit() throws IOException, so you don't have to catch it and rethrow it. Just take the try catch away completely.

PR #2875, opened 2018-11-29
[openexchangerates] Fixed invalid use of base vs counter

nielsdraaisma:
- Added test for OERMarketDataService and OERMarketDataServiceRaw
- Fixed invalid use of base vs counter

### DIFF
--- a/xchange-openexchangerates/pom.xml
+++ b/xchange-openexchangerates/pom.xml
@@ -29,6 +29,18 @@
             <version>${project.version}</version>
         </dependency>
 
+        <!-- PowerMockito for unit tests -->
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito2</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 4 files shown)


xchange-openexchangerates/pom.xml
```diff
@@ -29,6 +29,18 @@
...
         </dependency>
 
+        <!-- PowerMockito for unit tests -->
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito2</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-reflect</artifactId>
```
timmolter: Please define this in the parent POM and use the same version as the other powermock dependency.

nielsdraaisma: Fixed, i upgraded powermock to 2.0.0-RC.1 ( was 2.0.0-beta.5) in order to make it work using the same version.

PR #2876, opened 2018-11-29
[binance] better error handling

walec51:


xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceMarketDataService.java
```diff
@@ -86,44 +99,47 @@ public Ticker getTicker(CurrencyPair pair, Object... args) throws IOException {
...
   public Trades getTrades(CurrencyPair pair, Object... args) throws IOException {
-
-    Long fromId = null;
-    Long startTime = null;
-    Long endTime = null;
-    Integer limit = null;
-
     try {
-      if (args[0] != null) fromId = Long.valueOf(args[0].toString());
-    } catch (Throwable ignored) {
```
walec51: besides exception mapping I also fixed this because I consider silent errors as the more pure form of evil ;)

PR #2892, opened 2018-12-20
[xchange] added SystemOrverloadedException

titusthefox04:
I added SystemOverloadException to common xchange module to manage http error 502.
I added support to SystemOverloadException to bitmex

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/exceptions/SystemOverloadException.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/exceptions/SystemOverloadException.java
@@ -1,6 +1,9 @@
 package org.knowm.xchange.exceptions;
 
-/** An exception indicating there the rate limit for making requests has been exceeded */
+/**
+ * An exception indicating that the server is overloaded and the service is temporally unavailable
+ * i.e. when server return an http error 502
+ */
 public class SystemOverloadException extends ExchangeException {
 
   public SystemOverloadException(String message) {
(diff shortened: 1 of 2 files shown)


xchange-core/src/main/java/org/knowm/xchange/exceptions/SystemOverloadException.java
```diff
@@ -0,0 +1,21 @@
+package org.knowm.xchange.exceptions;
+
+/** An exception indicating there the rate limit for making requests has been exceeded */
```
walec51: then is the system overloaded in general (as the name of the exception suggests) or is the call rate for your account exceeded (as the comment suggests)

for the former we already have an exception type: RateLimitExceededException

titusthefox04: My idea is to manage overload in general i.e. when server return an http error 502.
I added the particular exception because with the latest changes Bitmex client generates a generic exception when the server returns the error in attached image; then my software cannot understand what is really happened. Before introduction of "updateRateLimit" function, Bitmex client didn't trap the exception and I could manage 502 error;  if you have a better idea to return the specific error, I can modify my code and create a pull request.

image

walec51: still the name of your class says one thing and the comment above it says another, fix the comment then

walec51: look ok to me now

PR #2894, opened 2018-12-27
Multiple changes, see comments per commit

roeiba:

PR #2898, opened 2019-01-07
[core] - Reverted changes to currency class and added test to validate toString behaviour

nielsdraaisma:
See attached testcase , Currencies.XBT.toString() should not return BTC, since various implementations use toString to determine the currency code that gets send to the exchange.


walec51:
+1

this will probably break other changes introduced in

https://github.com/knowm/XChange/pull/2894/files by @roeiba 

reverting the entire PR might be required to have a working branch

PR #2905, opened 2019-01-13
BTCTurk Authenticated APIs

mertguner:
I added to authenticated APIs for BTCTurk.
I added examples for new APIs.


mertguner:
i will try again

PR #2907, opened 2019-01-13
[bitfinex] add placeStopOrder support, consistent getOrders behaviour with other exchanges

badgerwithagun:
The implementation of `placeStopOrder` is harmless enough; I've just aliased the previous very bitfinex-specific flag-based method of submitting stop orders with the newer API based on `StopOrder`.

However, the change to `getOrders` is a breaking change. It brings Bitfinex in line with Coinbase Pro and Binance in that stop orders are returned in the "hidden orders" side of `OpenOrders` rather than the limit orders side, thus existing code will have to change to pick up stop orders.  My intention is to also do this for some other exchanges and standardise on it.

I have considered the possibility of a feature toggle allowing the behaviour to be reverted to the old behaviour for existing users.  Please let me know if that would be useful, and if so, the most consistent way to implement it.


badgerwithagun:
Any thoughts on this, @timmolter?  If you're worried about the backwards compatibility issue, I can come up with something.

PR #2908, opened 2019-01-13
Extend xchange-kucoin to be able to retrieve k-lines and wallet balance.

israelcolomer:


israelcolomer:
Not intended to create a PR

PR #2920, opened 2019-01-29
[Abucoins] Remove whole module because exchange is no longer operating

jakubkalicki:
Exchange abucoin is not working anymore. 
See: [https://abucoins.com/en](url)


walec51:
API is still working and is used for downloading historical transactions

jakubkalicki:
Ok, i'm closing in that case.

PR #2946, opened 2019-02-16
Kucoin 2.0 API

### LINKED ISSUE #2914: Kucoin 2.0 API
Is anyone working on implementing this? I don't want to duplicate effort.

Happy to assist if anyone wants to spread the load.

badgerwithagun:
Fixes #2914.

Also removes the 1.0 implementation which is now completely non-functional.

This implementation is unusual in that it is a thin wrapper around the official SDK. The official SDK hasn't been published to Maven Central (yet?) so I've forked it, fixed it and republished to Maven Central from my own org for the time being.


xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/v2/KucoinV2ExceptionClassifier.java
```diff
@@ -0,0 +1,39 @@
...
+public final class KucoinV2ExceptionClassifier {
+
+  KucoinV2ExceptionClassifier() {}
+
+  public static RuntimeException classify(KucoinApiException e) {
+    if (e.getMessage().contains("check environment variables")) {
+      return new ExchangeSecurityException(e.getMessage(), e);
+    }
+
+    // TODO need to handle HTTP error codes which aren't covered
```
badgerwithagun: See https://github.com/knowm/XChange/issues/2914#issuecomment-464319161 for query on this one.

badgerwithagun:
Added removal of the existing implementation which doesn't work as of this morning.

badgerwithagun:
Broken since removal of Kucoin v1.  I'm also hitting some problems with the Kucoin SDK.  Hold off on using/merging for the moment.

badgerwithagun:
Added balances and open orders.

Remaining:

* Create limit order
* Create stop order
* Create market order
* Cancel order
* Trade history

akzolt:
When will it be merged? (approximately)

badgerwithagun:
1-2 weeks to completion at this rate if I don't get any help, then usually a few days for review and merge.

timmolter:
It's preferable if you do not use an SDK dependency and interact directly via REST requests like the rest of the modules in XChange. This will allow more people to understand and fix code in this module and not be dependent on the SDK getting updated. If that ever gets abandoned or has slow update turnaround time, the the XChange module is broken and we can't fix it easily. That fact that the SDK isn't even om Maven Central is also bad.

badgerwithagun:
@timmolter: darnit.

The SDK will be on Maven Central shortly. I forked it to my own organisation and have published it from there to the OSSRH snapshots repo, heavily modified (with all my changes contributed upstream).  I was going to do a full release and pin to a release version before declaring this ready for merge, so it will be a _bona fide_ dependency.

The code in the SDK is MIT, so someone could port most of it into XChange unmodified, but this is work I wasn't planning on when I declared my intentions [ten days ago](https://github.com/knowm/XChange/issues/2914#issuecomment-464170809).  I don't have time for the extra work right now, much as I'd love to.

Otherwise, it's pretty much done.  Everything now works.

Would you rather wait, leave this PR open waiting for this, or merge (once I've done the Maven Central publish) and let me (or someone else) follow up with the porting work as a separate PR?

There are a few other TODOs in the code which I was going to follow up on anyway:

- Some flags and options which I think should become part of the core API but are currently exchange-specific, such as `TimeInForce` or client-supplied ids for orders. I want to commonise these.
- Apply parameter filtering to both limit and stop orders in `OpenOrdersParams` in a cleaner (but less compatible) way
- Kucoin talks about rate limiting in the API docs but the API doesn't seem to enforce these yet, so I haven't been able to make sure these errors are propagated correctly

I could just post up four issues here and bite them off in turn.

badgerwithagun:
Either way, forked SDK is now in Maven Central and I've pinned xchange-kucoin to use it.

timmolter:
If you have time, it be better to get rid of that dependency. But I know the world isn't an ideal place. ;)

straumat:
@badgerwithagun hello, is it normal that the open field is always null ?

badgerwithagun:
> @badgerwithagun hello, is it normal that the open field is always null ?

The open field on what?

straumat:
Ticker

badgerwithagun:
Looks like at the time this was written, the Kucoin API wasn't returning an open price so it wasn't available to be mapped to XChange:

https://github.com/knowm/XChange/blob/develop/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAdapters.java#L57

That may have changed since.

PR #2947, opened 2019-02-18
[cryptofacilities] consistent date parsing using new instance of SDF

andre77:
consistent date parsing using new instance of SDF to be thread safe

### DIFF
--- a/xchange-cryptofacilities/src/main/java/org/knowm/xchange/cryptofacilities/CryptoFacilitiesAuthenticated.java
+++ b/xchange-cryptofacilities/src/main/java/org/knowm/xchange/cryptofacilities/CryptoFacilitiesAuthenticated.java
@@ -78,7 +78,8 @@ public interface CryptoFacilitiesAuthenticated extends CryptoFacilities {
   CryptoFacilitiesFills fills(
       @HeaderParam("APIKey") String apiKey,
       @HeaderParam("Authent") ParamsDigest signer,
-      @HeaderParam("Nonce") SynchronizedValueFactory<Long> nonce)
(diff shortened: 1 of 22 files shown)


timmolter:
Would it be possible to use `DateUtils` instead? Maybe the following method would work:

```java
  /**
   * Converts an ISO formatted Date String to a Java Date ISO format: yyyy-MM-dd'T'HH:mm:ss.SSS'Z'
   *
   * @param isoFormattedDate
   * @return Date
   * @throws com.fasterxml.jackson.databind.exc.InvalidFormatException
   */
  public static Date fromISODateString(String isoFormattedDate)
      throws com.fasterxml.jackson.databind.exc.InvalidFormatException {

    SimpleDateFormat isoDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
    // set UTC time zone - 'Z' indicates it
    isoDateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
    try {
      return isoDateFormat.parse(isoFormattedDate);
    } catch (ParseException e) {
      throw new InvalidFormatException("Error parsing as date", isoFormattedDate, Date.class);
    }
  }
```

Otherwise you could add a new method in that class so it can be reused by other modules.

andre77:
@timmolter 
i used the own method `String format(Date date)` since the method `org.knowm.xchange.utils.DateUtils.toUTCString(Date)` produces not exactly the desired result.
Adding another additional method to the core date utils is not the best option imho, since every exchange can differ slightly in terms of parsing/formatting dates...

PS: why don't they use the unix time with millis ?...

Issue #2886, opened 2018-12-15
Bitmex exchange metadata and use of currency pairs is inconsistent


badgerwithagun:
If I do the following:
```
 Exchange exchange = ExchangeFactory.INSTANCE.createExchange(BitmexExchange.class.getName());
 System.out.println(exchange.getExchangeMetaData().getCurrencyPairs().keySet());
```
I get:
```
[BTC/BTC, EOS/BTC, BCH/BTC, ADA/BTC, ETH/USD, ZEC/BTC, TRX/BTC, XLM/BTC, LTC/BTC, \
XRP/BTC, BTC/USD, ETH/BTC, DASH/BTC, XMR/BTC, ETC/BTC]
```
There a a few problems here

- Several of these pairs aren't actually supported by Bitmex (looks like `bitmex.json` is just a copy & paste. I can fix that easily)
- `BTC/BTC`?
- The fact that `BTC` is used instead of `XBT`.

This last bit is interesting.  The code in `org.knowm.xchange.bitmex.BitmexAdapters.adaptToExchangeMetaData` seems to do this deliberately, but...
```
System.out.println(exchange.getMarketDataService().getTicker(CurrencyPair.BTC_USD));
```
... returns null, whereas...
```
System.out.println(exchange.getMarketDataService().getTicker(new CurrencyPair("XBT", "USD")));
```
... works (as I would expect, to be honest!)

How _should_ this work?  Should Bitmex accept XBT as one might expect (in which case it's the exchange metadata implementation that's wrong), or should it translate BTC to XBT internally (in which case the first of those two examples should work, but not the second)?

My inclination is to remove all the cleverness in `org.knowm.xchange.bitmex.BitmexAdapters.adaptToExchangeMetaData` and make it use Bitmex-native currency codes throughout.

badgerwithagun:
It seems to be a bit worse than I thought.  Here's what I have to do to get the ticker and trades for ADA:
```
System.out.println(exchange.getMarketDataService().getTicker(new CurrencyPair("ADA", "Z18")).getLast());
System.out.println(exchange.getMarketDataService().getTrades(new CurrencyPair("ADA", "BTC"), BitmexPrompt.QUARTERLY));

So I can't use the same currently pair to get trades as I do for tickers.  Anyone know why?  Why not just make it `ADA/Z18`, `XBT/Z18`, `XBT/USD` across the board, in line with Bitmex' own representation, and get rid of `BitmexPrompt`?

[...]

PR #2950, opened 2019-02-24
[Bitmex] Fix meta data and currency inconsistence

jakubkalicki:
See #2886


jakubkalicki:
@badgerwithagun, i have finished. You can review it, if you like.

xchange-bitmex/src/main/resources/bitmex.json
```diff
@@ -1,111 +1,123 @@
 {
   "currency_pairs": {
-    "BCH/BTC": {
+    "XRP/H19": {
```
badgerwithagun: I wonder if these should be here at all. In a few months time, the H19 contracts will expire and this will be out of date.

jakubkalicki: That's good point. I have removed everything that will expire some day.

xchange-bitmex/src/main/resources/bitmex.json
```diff
@@ -1,111 +1,63 @@
...
-      "scale": 5,
-      "withdrawal_fee": 0.00002
+      "scale": "",
+      "withdrawal_fee": ""
     },
-    "LTC": {
-      "scale": 5,
-      "withdrawal_fee": 0.00100
+    "BCH": {
+      "scale": "",
```
badgerwithagun: Why are these all blanked?

xchange-bitmex/src/test/java/org/knowm/xchange/bitmex/service/marketdata/BitmexOrderBookFetchIntegration.java
```diff
@@ -0,0 +1,43 @@
...
+  @Before
+  public void setUp() {
+    bitmexExchange =
+        (BitmexExchange) ExchangeFactory.INSTANCE.createExchange(BitmexExchange.class.getName());
+    marketDataService = bitmexExchange.getMarketDataService();
+  }
+
+  @Test
+  public void getOrderBookTest() throws IOException {
+    CurrencyPair pair = new CurrencyPair("XBT", "M19");
```
badgerwithagun: Another expiring contract reference. Can we update this test to use `determineActiveContract`?

xchange-bitmex/src/test/java/org/knowm/xchange/bitmex/service/marketdata/BitmexTickerFetchIntegration.java
```diff
@@ -0,0 +1,66 @@
...
+  @Test
+  public void determineActiveContractTest() {
+    CurrencyPair perpetualXBT =
+        bitmexExchange.determineActiveContract("BTC", "USD", BitmexPrompt.PERPETUAL);
+    CurrencyPair quarterlyLTC =
+        bitmexExchange.determineActiveContract("LTC", "BTC", BitmexPrompt.QUARTERLY);
+    CurrencyPair quarterlyETH =
+        bitmexExchange.determineActiveContract("ETH", "BTC", BitmexPrompt.QUARTERLY);
+
+    assertThat(new CurrencyPair("XBT", "USD")).isEqualTo(perpetualXBT);
```
badgerwithagun: Again, this will break in a few months.

xchange-bitmex/src/test/java/org/knowm/xchange/bitmex/service/marketdata/BitmexTradesFetchIntegration.java
```diff
@@ -0,0 +1,42 @@
...
+  @Before
+  public void setUp() {
+    bitmexExchange =
+        (BitmexExchange) ExchangeFactory.INSTANCE.createExchange(BitmexExchange.class.getName());
+    marketDataService = bitmexExchange.getMarketDataService();
+  }
+
+  @Test
+  public void getTradesTest() throws IOException {
+    CurrencyPair pair = new CurrencyPair("XBT", "M19");
```
badgerwithagun: And another

xchange-bitmex/src/main/resources/bitmex.json
jakubkalicki: At Bitmex we have following currencies: XBT, USD and contracts. In this json file we define currencies such as BCH, LTC etc, but in fact we don't use them as currencies. They are just symbols defined to keep consistency with service interfaces. 

Scales should be defined for contracts (so for our pairs like XTB/H19) themselves, which is done dynamically in updateExchangeMetaData method.

xchange-bitmex/src/test/java/org/knowm/xchange/bitmex/service/marketdata/BitmexTradesFetchIntegration.java
jakubkalicki: I will change M19 to USD

PR #2953, opened 2019-02-27
Changes to itBit API

chrisrico:
As of 2019-02-25 they now support 8 decimal places of precision
On 2019-02-28 they will support market orders

Both of these changes are documented in their [staging API documentation](https://staging-api.itbit.com/docs).

### DIFF
--- a/xchange-itbit/api-specification.txt
+++ b/xchange-itbit/api-specification.txt
@@ -2,6 +2,5 @@ IT Bit Exchange API specification
 ======================================
 
 
-http://api-portal.anypoint.mulesoft.com/itbit/api/itbit-exchange/docs/raml
-http://api-portal.anypoint.mulesoft.com/itbit/api/itbit-exchange/docs/code
-
+https://api.itbit.com/docs
+https://staging-api.itbit.com/docs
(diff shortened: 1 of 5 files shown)


chrisrico:
Hmm, this isn't going to work. Their API is weird and needs a fiat amount in `funds` when placing a BUY market order, but a crypto amount in `amount` when placing a SELL. I don't think that will work with the MarketOrder API unless we calculate the fiat amount based on the order book, which seems highly sub-optimal.

timmolter:
I'm not really sure. You'll have to do some trial and error testing.

chrisrico:
I don't think I explained my question very well. What I mean is this:

itBit expects to receive a fiat amount when issuing a buy order, and a crypto amount when issuing a sell order. Is it safe *from an API perspective*, to expect the same value (`MarketOrder.getOriginalAmount()`) to refer to either a fiat amount or a crypto amount depending on the `MarketOrder.getType()`?

timmolter:
From the high level XChange API, it should behave like all the other modules so that there are no surprises. If you need to do some conversions at the *Raw level before sending the requests off to itBit then that's how it should be done.

walec51:
@chrisrico will you be working on this PR?

PR #2958, opened 2019-02-28
[Lykke,Coinex,Coindeal,Blockchain,CoinbasePro,Quoinex] Adding 3 exchanges and fix/update some other parts

makarid:
Hello, this is a big commit and i am sorry for that, i don't know to chop it and make it smaller ones.If you can help me i will do it smaller. I have add 3 new exchanges and update some issues on Quoinex, CoinbasePro and some other minor commits. Please see it and tell me if anything is wrong in order to change it.Thank you very much


makarid:
I get an error on java 11 build only. I have HTTPException class on LykkeExchange which is from javax.xml.ws.http. Does this library has been removed on java11?

timmolter:
yeah, see here. What exactly do you use that dependency for? Can you paste a link showing the Java code? Perhaps I can recommend a workaround without adding the 3rd party dependency.

makarid:
Hello @timmolter . I use it in this code: 

    public boolean cancelAllLykkeOrders(String assetPairId) throws IOException,HTTPException {
        try{
            lykke.cancelAllOrders(assetPairId,apiKey);
            return true;
        }catch (LykkeException e){
            throw new ExchangeException(e.getMessage());
        }
    }

Well actually i don't remember why i put it. :)

timmolter:
I'm pretty sure you don't need the `HTTPException`. No other exchange uses that. See the ANX implementation in the project and read step 7 here: <https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices>. If anything use the rescu http exceptions:<https://github.com/mmazi/rescu/tree/develop/src/main/java/si/mazi/rescu>

makarid:
OK then,i will delete it and resubmit the commit

timmolter:
instead of always closing the PR you can just keep pushing changes.

PR #2960, opened 2019-02-28
[Lykke,Coinex,Coindeal,Blockchain,CoinbasePro,Quoinex] Adding 3 exchanges and fix/update some other parts

makarid:
Hello, this is a big commit and i am sorry for that, i don't know to chop it and make it smaller ones.If you can help me i will do it smaller. I have add 3 new exchanges and update some issues on Quoinex, CoinbasePro and some other minor commits. Please see it and tell me if anything is wrong in order to change it.Thank you very much


pom.xml
```diff
@@ -185,7 +185,10 @@
 		<module>xchange-wex</module>
 		<module>xchange-yobit</module>
 		<module>xchange-zaif</module>
-	</modules>
+        <module>xchange-coindeal</module>
+        <!--<module>xchange-coinex</module>-->
```
timmolter: you really want this commented out?

pom.xml
```diff
@@ -354,6 +357,12 @@
 			<version>${version.assertj}</version>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
```
timmolter: What is this needed for exactly?

pom.xml
makarid: Coinex isn't completed like the rest exchanges, that is why i have comment it. I am adding it in order more people can update it but i don't want to included it on the project yet. Maybe i am wrong here, feel fee to tell me your opinion,about this.Thanks

pom.xml
timmolter: Can you say why `jackson-annotations` is necessary?

PR #2961, opened 2019-02-28
[Bittrex-Kraken] various fixes

cymp:
Bittrex: fixed BigDecimal calculation with now possible null values, due to Bittrex 02/27/2019 backend update
Kraken: fixed null pointer exception cases on not supported currency pairs


cymp:
I am not so happy to replace the nulls with zeros, but that's how it seems to work on their side now.

PR #2986, opened 2019-03-08
[core] Overrode the 'hashCode' method.

Enfernuz:
The class is lacking a hashCode implementation.


Enfernuz:
I'm not sure that the build failed because of the introduced change. The job logs lead to a failed integration test with `btcTurkMarketDataService.getBTCTurkOHLC(...)` that is no way related to the `Wallet` class.

timmolter:
it's because an integration test ran and failed. Not your fault and I pushed a correction.

PR #2993, opened 2019-03-13
Simulated exchange

### LINKED ISSUE #2964: [simulated] Create simulated/dummy/mock exchange
I have a lot of code built up around XChange (and xchange-stream) which needs end-to-end testing.

To facilitate this, I am in the progress of creating a "native" in-process exchange implementation, `xchange-simulated`, which implements all the main XChange interfaces but using a simple, low performance but thread-safe in-process "exchange".  This has its own L3 orderbook for each coin, a matching engine, and allows multiple `Exchange` instances with different api-keys in their `ExchangeSpecification` to trade against each other.

It will support:

* All market data (ticker, order book, trades)
* Trading (limit order and market order only for now, no stops, but that could easily be added)
* Trade history
* Open orders
* Account balances

I will eventually extend it with an equivalent xchange-stream implementation.

In all cases, I am intending to ensure that it throws the correct core exchange exceptions (e.g. `FundsExceededException`) at the right times, but will also _randomly_ throw exceptions that a designer should expect to handle gracefully, such as `NonceException`, `SocketTimeoutException` or `RateLimitExceededException`.  The intention is to flush out reliability issues in developer code and encourage adherence to xchange-core standards.

Should be ready soon.

badgerwithagun:
Fixes #2964.


badgerwithagun:
This is working well in my test cases now and is ready for merge.

Current obvious limitations:

- No stop support yet
- Only one pair currently (that can easily be extended just by updating `simulated.json`, I just haven't bothered)
- No support for any non-standard features which are implemented differently across different exchanges, such as time-in-force, self-trade prevention, etc.  I think these are ripe for converting into generic APIs and then supporting here too.
- It's highlighted how patchy and inconsistent the exceptions are that are raised by the different exchange implementations. I've tried to establish a reasonable best-practice superset but again, perhaps this is a decent catalyst for more standardisation.

timmolter:
nice work!

pom.xml
```diff
@@ -299,6 +301,14 @@
                 <version>${version.powermock}</version>
                 <scope>test</scope>
             </dependency>
+      
+            <!-- Hamcrest for testing -->
+            <dependency>
```
timmolter: we already have `org.assertj` for testing. Can you use that instead?

badgerwithagun:
Not familiar with it, but I'll have a look.

timmolter:
https://github.com/mkordas/hamcrest-to-assertj
maybe useful...

badgerwithagun:
Unfortunately it contains slightly more hardcore Hamcrest (custom matchers). If not, I'd have just used raw asserts rather than adding the dep.

I'm sure it's workaroundable.

badgerwithagun:
OK, done. Turns out AssertJ's lacking a bunch of things I take for granted in Hamcrest, but I was able to work around them.

PR #3017, opened 2019-03-28
[livecoin] orderbooks parsing fix

walec51:
Livecoin added a third value to their order book entries which broke the current parsing mechanism - it was returning empty order books.

```
{
    "timestamp": 1553777858097,
    "asks": [
        [
            "0.03430012",
            "0.01498781",
            1553777852095
        ],
        [
            "0.03430013",
            "0.00478876",
            1553777849476
        ],
```

I've implemented parsing of these type of arrays in a more proper way


timmolter:
please fix conflicts. Thanks.

PR #3021, opened 2019-03-29
[core] support null values for currency pair

andre77:
support null values for currency pair inside OpenOrdersParamCurrencyPair class


timmolter:
In which situation is this needed?

andre77:
its needed for calls, where  the currency pair parameter is __*optional*__ 
For example at Kucoin: https://docs.kucoin.com/#list-orders
the symbol is optional, so basically if you provide a pair, you get the list for that pair, but if you don't, you get the list for all pairs (symbols)

Issue #3008, opened 2019-03-19
[NEW EXCHANGE]  Derebit


dinozof:
Hi everybody,

I saw that Derebit is not supported yet. I think it would be nice to implement this new exchange since it's gaining some popularity 
https://www.deribit.com/

Cheers!

jakubkalicki:
I will start working on this in upcoming days.

jakubkalicki:
Basic market data requests are implemented now.

earce:
This was implemented

PR #3022, opened 2019-03-30
Deribit Implementation

jakubkalicki:
#3008 

Implementing market data end points!


jakubkalicki:
Following market data end points have been implemented so far:

- https://docs.deribit.com/v2/#public-get_instruments
- https://docs.deribit.com/v2/#public-get_currencies
- https://docs.deribit.com/v2/#public-get_order_book
- https://docs.deribit.com/v2/#public-get_last_trades_by_instrument
- https://docs.deribit.com/v2/#public-ticker

@timmolter have a look, please. I think that we can merge it in such state.

walec51:
could you improve error handling to follow best practices stated at:
https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-7

jakubkalicki:
@walec51, @timmolter 
Implemented basic error handling and added examples. Please, review.

xchange-deribit/src/main/java/org/knowm/xchange/deribit/v2/Deribit.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.ws.rs.core.MediaType;
+import org.knowm.xchange.deribit.v2.dto.marketdata.response.*;
+
+@Path("/api/v2/public")
+@Produces(MediaType.APPLICATION_JSON)
+public interface Deribit {
+
+  @GET
+  @Path("get_instruments")
+  DeribitInstrumentsResponse getInstruments(@QueryParam("currency") String currency)
```
walec51: please create a DeribitException similar to https://github.com/knowm/XChange/blob/develop/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/dto/BittrexException.java
and add it to each throws declaration here

in your case the DeribitException will have a field

```
  @JsonProperty("error")
  private DeribitError error;
```

this way you will not have to do response parsing in the error adapter

your DeribitException should also implement the HttpStatusExceptionSupport interface as I see you need it in the error adapter

you can find more explanation on how its done here: https://github.com/mmazi/rescu/wiki/Exception-handling

xchange-deribit/src/main/java/org/knowm/xchange/deribit/v2/dto/DeribitResponse.java
```diff
@@ -0,0 +1,86 @@
...
+      V result,
+      DeribitError error,
+      boolean testnet,
+      long usIn,
+      long usOut,
+      long usDiff) {
+    this.jsonRPC = jsonRPC;
+    this.id = id;
+    this.result = result;
+    this.error = error;
```
walec51: if there is an error then throw a ExceptionalReturnContentException here instead of saving the error DTO here

this will force the client to parse and throw a DeribitException  of which I mentioned in the previous comment

xchange-deribit/src/main/java/org/knowm/xchange/deribit/v2/dto/marketdata/response/DeribitOrderBookResponse.java
```diff
@@ -0,0 +1,22 @@
+package org.knowm.xchange.deribit.v2.dto.marketdata.response;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.knowm.xchange.deribit.v2.dto.DeribitError;
+import org.knowm.xchange.deribit.v2.dto.DeribitResponse;
+import org.knowm.xchange.deribit.v2.dto.marketdata.DeribitOrderBook;
+
+public class DeribitOrderBookResponse extends DeribitResponse<DeribitOrderBook> {
```
walec51: why not just use DeribitResponse<DeribitOrderBook> directly in the client interface?
these type of classes seams to add nothing usefull

xchange-deribit/src/main/java/org/knowm/xchange/deribit/v2/service/DeribitMarketDataServiceRaw.java
```diff
@@ -0,0 +1,121 @@
...
+    super(exchange);
+  }
+
+  public List<DeribitInstrument> getDeribitActiveInstruments(String currency) throws IOException {
+
+    DeribitInstrumentsResponse response = null;
+
+    try {
+      response = deribit.getInstruments(currency);
+    } catch (HttpStatusIOException ex) {
```
walec51: when you implement my suggestions with the DeribitException then it will get thrown here instead

methods in *ServiceRaw classes should throw out DeribitException 

they should be caught and adapted in *Service methods

jakubkalicki: Ok, but it will cause that some *ServiceRaw methods will miss adaptation (e,g. getDeribitActiveInstruments method), right?

walec51: yes, that's the point, raw methods should use exchanges DTO (DeribitException is also somewhat a DTO for error situations), adaption should happen in methods that implement generic interfaces

jakubkalicki: Ok, now i get it. Thanks.

xchange-deribit/src/main/java/org/knowm/xchange/deribit/v2/dto/marketdata/response/DeribitOrderBookResponse.java
jakubkalicki: Valid point, i think i was driven by other exchange implementation.

jakubkalicki:
@walec51, all your suggestions are implemented now.

timmolter:
Thanks, @walec51 and @jakubkalicki

PR #3046, opened 2019-04-15
[bitfinex] Fix NPE when fetching open stop-limit orders

badgerwithagun:
Also report unhandled order types in the future and warn where the data returned won't be accurate.

This code really needs to be moved to the v2 API which fully supports all order types.


badgerwithagun:
Wrong branch.

Issue #1544, opened 2017-06-21
more metadata needed?


npomfret:
I've just tried executing a trade on the BTC/LTC market on hitbtc which failed with the message 'Rounding necessary'.  Now it seems that LTC and BTC are represented on the 8 exchange with 8 decimal places, as is the price on the BTC/LTC exchange.  However, the quantity that is tradable on that market is rounded to just 1 decimal place.  Is there a common place in this API (like the .json metadata files) to specify that value?

I was thinking something like:

```
{
  "currency_pairs": {
    "LTC/BTC": {
      "price_scale": 8,
      "quantity_scale": 2
    }
  }

  ...

  "BTC": {
    "scale": 8
  },
  "LTC": {
    "scale": 8
  },
```

timmolter:
The metadata may or may not have specifically that, I don't know for sure as I've not yet used that specific feature. The json could also very well be outdated. You'll have to investigate it yourself unless someone else chimes in here.

npomfret:
Thanks for the reply - I'll give it some thought.

molexx:
I'd be interested to know any progress/thoughts on this, thanks.

npomfret:
I'm quite new to this project, so my ideas might not be quite thought out enough yet... 

It feels like in order to trade on any exchange, each needs some common metadata.  Specifically, so far I've come across:

_currency scale_ - for use when getting your wallet data
_price scale_ - for use when making a limit order
_quantity scale_ - for use when setting the quantity/size of any order (which doesn't seem to be represented anywhere)

And I think these should be present in the exchange metadata for __all__ exchanges.

_fees_ - these vary from exchange to exchange.  Some have taker v's maker fees, some have none at all (coin floor for example).  They are usually expressed as a percentage of trade size.  But some exchange have more complex models, with discounts for example.  Some have funding fees, as well as trade fees. And sometimes fees vary based on what's being traded (as well as the trade size).

I think maybe the fees are a bit trickier to model and perhaps should be removed from the static data al together? 

I would be very useful if all trades returned by `getTradeHistory` would give amounts without fees applied and so the user can apply the fees separately.  So far I've sound some implementations that do as I've described, and some that give amounts net (after fees have been removed from the trade value).  This inconsistency makes it difficult to do accounting. 

In addition, we also may have:

_rate limits_ - to prevent a client spamming the exchange with requests.  I think these are typically easy to model and probably belong in the static data.

[...]

PR #3066, opened 2019-05-08
[core] - Added support for tradingFeeCurrency to CurrencyPairMetaData

nielsdraaisma:
This is a relatively small PR to address one of the issues raised in https://github.com/knowm/XChange/issues/1544 , it adds a new field `tradingFeeCurrency` to the CurrencyPairMetaData class. 

Future changes can leverage this to better calculate and present fee's to the client even if not returned by the API.

I have not gone through all the exchanges to populate the metadata but am happy to cover some exchanges if this PR gets merged. We could add a default mechanism to always fallback to the counter but that might give users a false sense of quality so i did not do that as part of this PR.


nielsdraaisma:
Hi @timmolter , would you be able to give some feedback on this PR ?.

PR #3071, opened 2019-05-10
Binance + CoinbasePro meta data all loaded via API

amjad1210:
We shouldn't use hard coded values but rather load them from the API if they're available.

All metadata for CoinbasePro + Binance is loaded via the API now.
This includes base amount scale, counter amount scale, wallet amount scale and withdrawal fees.

Added baseScale field which is the amount precision for a BID order type.


amjad1210:
FYI

CurrencyPairMetaData:
  - baseScale is the precision amount for the base currency.
  - priceScale is the precision amount for the counter currency.

timmolter:
could you please fix the conflicting files?

amjad1210:
I've fixed the conflicts.

walec51:
you now get a NPE when you try to use Binances Public API without passing any credentials because of this PR ....

cymp:
Yes I can confirm...

walec51:
quick fix: https://github.com/knowm/XChange/pull/3099
however this whole change should be rethinked

amjad1210:
Sorry I've always used it with credentials completely slipped my mind that people may be using it without credentials. I'll add a fix within the next few hours please bare with me.

amjad1210:
Fix in #3100

Issue #3073, opened 2019-05-11
[Binance] Meta data for symbols is wrong


gregoreficint:
The meta information from Binance contains incorrect currency pairs and currencies. I will illustrate this point with a concrete example:
The meta data of the Binance exchange contains the following data:
`{CurrencyPair@5246} "IOT/XBTC" -> {CurrencyPairMetaData@5247} "CurrencyPairMetaData [tradingFee=0.1, minimumAmount=1, maximumAmount=9E+7, priceScale=8, amountStepSize=1]"
{CurrencyPair@5248} "IOT/XETH" -> {CurrencyPairMetaData@5249} "CurrencyPairMetaData [tradingFee=0.1, minimumAmount=1, maximumAmount=9E+7, priceScale=8, amountStepSize=1]"`

Please notice that the exchange has three strange currencies (XBTC, XETH and IOT).

If you use the native API call (https://api.binance.com/api/v1/exchangeInfo) you will notice that the correct currency pairs are:
```
symbol | "IOTXBTC"
status | "TRADING"
baseAsset | "IOTX"
quoteAsset | "BTC"

symbol | "IOTXETH"
status | "TRADING"
baseAsset | "IOTX"
quoteAsset | "ETH" 
```
Thus, it should have been "IOTX", "BTC" and "ETH" instead.

Fix is in the pull request  #3074

PR #3074, opened 2019-05-11
[Binance] Fix for wrong information for some currencies and currency pairs in meta data

gregoreficint:
As described in #3073, the meta data for certain currencies and currency pairs such as IOTX/BTC is not correctly stored and retrieved for Binance. 

The root cause of this problem is in the initialization routine for Binance.  

Within the `remoteInit()` method in `BinanceExchange.java` we find the following code:
```Symbol[] symbols = exchangeInfo.getSymbols();

for (BinancePrice price : marketDataService.tickerAllPrices()) {
  CurrencyPair pair = price.getCurrencyPair();

  for (Symbol symbol : symbols) {
    if (symbol
        .getSymbol()
        .equals(pair.base.getCurrencyCode() + pair.counter.getCurrencyCode())) {
```

Interestingly, this routine combines two for-loops while only one is necessary since you want to iterate just once over the currency pairs. Additionally, we see two ways of receiving the information: Once the information about the symbols is received via the designated way (`exchangeInfo.getSymbols()` maps to `GET /api/v1/exchangeInfo`) including all the details. In the second way the currency pairs are received via the price information (`marketDataService.tickerAllPrices(marketDataService.tickerAllPrices()` maps to `GET api/v1/ticker/allPrices`). 

If we look at the responses from Binance we can guess already why this problem exists. The designated way gets a response that clearly shows which currencies are used in the currency pair IOTXETH, i.e. IOTX and ETH
```
symbol | "IOTXETH"
status | "TRADING"
baseAsset | "IOTX"
baseAssetPrecision | 8
quoteAsset | "ETH"
quotePrecision | 8
orderTypes | […]
icebergAllowed | true
isSpotTradingAllowed | true
isMarginTradingAllowed | false
```
while in the second way the response contains only the string “IOTXETH”, which is later not split correctly into its currencies.
```
symbol	"IOTXETH"
price	"0.00005001"
```

If we dig deeper, we find the reason for this behaviour. Getting the prices triggers the routine `getCurrencyPairFromString(String currencyPairString)` from the `CurrencyPairDeserializer` class that tries to guess the right way to split the string. While this function is really cool it cannot be expected to guess every currency pair correctly. One currency pair, where it fails, is “IOTXETH”, where it splits the string into “IOT” and “XETH”. 

There are two ways to fix the problem. First, one could try to fix the `CurrencyPairDeserializer` for all currency pairs where the function is working incorrectly. However, this function is already really great at doing its job and I believe it will not be possible to get this function right for every current and future currency pair. Secondly, we might change the for-loop in the Binance class and dispense the price function to receive symbols altogether. This seems way better since we get the full and accurate information already with the `exchangeInfo.getSymbols()` so that we should not abuse the price function to do something it is not designed for. I assume that the author used the price function to only get the currency pairs that are currently traded on Binance. However, we can get this information also from the exchange symbols themselves, which have attached a status. We would filter out all symbols with the status “BREAK” since these symbols have been delisted by Binance. We would keep all symbols with other status values since trading might only be stopped for a temporary period. Thus, our fixed function looks like:
```
for (Symbol symbol : symbols) {
  if (!symbol.getStatus().equals("BREAK")) { // Symbols with status "BREAK" are delisted
```

In order to keep the correct behaviour in future implementations, we also add two test cases that previously would have failed but now show the correct behaviour:
```
currPair =
    currencyPairs.keySet().stream()
        .filter(cp -> "IOTX/ETH".equals(cp.toString()))
        .collect(StreamUtils.singletonCollector());
Assert.assertNotNull(currPair);

curr =
    currencies.keySet().stream()
        .filter(c -> c.equals(new Currency("IOTX")))
        .collect(StreamUtils.singletonCollector());
Assert.assertNotNull(curr);
```

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
@@ -8,7 +8,6 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import org.knowm.xchange.BaseExchange;
 import org.knowm.xchange.ExchangeSpecification;
-import org.knowm.xchange.binance.dto.marketdata.BinancePrice;
 import org.knowm.xchange.binance.dto.meta.exchangeinfo.BinanceExchangeInfo;
 import org.knowm.xchange.binance.dto.meta.exchangeinfo.Filter;
 import org.knowm.xchange.binance.dto.meta.exchangeinfo.Symbol;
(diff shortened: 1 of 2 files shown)


walec51:
you now get a NPE when you try to use Binance without passing any credential because of this PR ....

gregoreficint:
@walec51 I honestly believe that this PR is not the cause of this behaviour since this PR just removes one function call but does not add one. Thus, the behaviour should not have changed regarding not using any credentials. Could you verify, please, that this PR causes this behaviour rather than a different one?

walec51:
I'm sorry, your right, this is the PR that introduced the bug: https://github.com/knowm/XChange/pull/3071/files

PR #3088, opened 2019-05-21
[kucoin] funding methods like deposit and withdrawal

andre77:

PR #3092, opened 2019-05-23
[kucoin] Reduce number of trades or orders fetched by default

badgerwithagun:
Kucoin has introduced a limit.


andre77:
I was working on other stuff at kucoin, and did the same changes in my PR, see https://github.com/knowm/XChange/pull/3088
just fyi
sometimes just too lazy to open an extra PR :-)

timmolter:
I guess I can close this then.

PR #3097, opened 2019-05-25
[Bibox] Fixed get balances

cymp:


cymp:
@timmolter Please have a look at 1a654ba. Any idea why the compilation fails and forces me to do that? I would expect the lombok dependency to be resolved with the dependency to core...

cymp:
It's in the root pom though, what should we do?

timmolter:
actually, they way you did it is correct. It's only defined in the parent pom, and only the modules that depend on it should explicitly define it.

cymp:
ok so I think you can merge :)

PR #3099, opened 2019-05-27
[binance] fixed NPE in remoteInit if you pass no credentials

walec51:
This PR made the Binances public API unusable without any credentials: https://github.com/knowm/XChange/pull/3071

My PR fixed the NPE case. However the author of the PR mentioned above should rethink his change in the context of such an use case.


amjad1210:
Hey could you remove this PR i'll redesign it to work without credentials. This is just a hack to remove the nullpointer and doesn't resolve the issue.

amjad1210:
Can you close this PR please i've put a full fix in #3100

PR #3100, opened 2019-05-27
Fix assetdetails NP for users using public API

amjad1210:
Fix NP for users using the public API without credentials.

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
@@ -120,27 +120,33 @@ public class BinanceExchange extends BaseExchange {
                         because their API returns current fee directly */
                   stepSize,
                   null));
-          if (!currencies.containsKey(currentCurrencyPair.base)) {
-            AssetDetail baseAsset = assetDetailMap.get(currentCurrencyPair.base.getCurrencyCode());
-            BigDecimal baseWithdrawlFee =
-                baseAsset != null ? baseAsset.getWithdrawFee().stripTrailingZeros() : null;
-            currencies.put(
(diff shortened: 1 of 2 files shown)


xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
```diff
@@ -120,29 +120,35 @@ public void remoteInit() {
...
-          if (!currencies.containsKey(currentCurrencyPair.counter)) {
-            AssetDetail counterAsset =
-                assetDetailMap.get(currentCurrencyPair.counter.getCurrencyCode());
-            BigDecimal counterWithdrawlFee =
-                counterAsset != null ? counterAsset.getWithdrawFee().stripTrailingZeros() : null;
-            currencies.put(
-                currentCurrencyPair.counter,
-                new CurrencyMetaData(counterPrecision, counterWithdrawlFee));
-          }
+
```
walec51: move the declaration in line 83 `Map<String, AssetDetail> assetDetailMap = accountService.getAssetDetails();` to here

walec51:
thanks for fixing this, just one small remark in the review

xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
amjad1210: Hey @walec51 

`Map<String, AssetDetail> assetDetailMap = accountService.getAssetDetails();` is an API call to get all the assets and should only be called once. Placing it on line 123 put it within the for loop which will call it over and over again resulting in a possible ban due to exceeding the rate limit.

walec51: ok, your right, didn't notice the loop above

PR #3101, opened 2019-05-29
[Okex]  API v3

roeiba:
New functionality added using API v3:
* Withdrawals support
* Transfer support
* Request deposit address

Instead of writing new v3 API over mazi, I took a different approach and implemented the missing methods using the Okex official java SDK. 

As the repo is not on maven, I also committed the SDK project (taken from https://github.com/okcoin-okex/open-api-v3-sdk/tree/master/okex-java-sdk-api) 

In order to use api V3 methods, user must set provide 'apikey_v3' and 'secretkey_v3' on exchange specifications.

### DIFF
--- a/xchange-okcoin/pom.xml
+++ b/xchange-okcoin/pom.xml
@@ -28,6 +28,12 @@
             <artifactId>xchange-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        
+        <dependency>
+            <groupId>com.okcoin.commons</groupId>
+            <artifactId>okex-java-sdk-api</artifactId>
+            <version>1.0.0</version>
+        </dependency>
 
     </dependencies>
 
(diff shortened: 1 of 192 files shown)


roeiba:
Need to fix broken build.

andre77:
hi @roeiba 
just stumbled upon your PR. 
Last days a spent also a couple of hours to implement the Okex API v3, but used the "mazi" way, as most of the implementations here, see https://github.com/knowm/XChange/pull/3118
may be we should bundle our work or synchronise it some how to avoid the work being done twice :-)

later on i will also implement the deposit/withdrawal methods of the API, but again using mazi
just to let you know

roeiba:
Hi Andrea,

Thank you sharing.

I have withdrawals and deposits already implemented on the branch of the
pull request.

I just had problems adding the OKEx Sdk as dependent project as it’s not on
maven.

Solution would be to adding the sdk jar, but I don’t have the time to do it
right now.

Regards,
Roei

andre77:
like i said, i would go the "mazi" way, not the OKEx Sdk

timmolter:
Yeah, definitely don't use the SDK jar.

Issue #3126, opened 2019-06-27
[Kraken] UserTrades sorted by Ids by default - though the Ids are not "sortable"


gregoreficint:
The function getTradeHistory(TradeHistoryParams params) in KrakenTradeService sorts the trades by ID. However, the ids in Kraken have no natural order so the trade history should be sorted by timestamp as default. The ids have the format of A-B-C which A, B, C being strings. I believe the strings refer to the individual ledger entries in Kraken.

These are some real Kraken trades to illustrate my point.

These are the trades sorted (desc) by time in Unix millis:

Ids | Time in unix millis | Time human readable | Order type | Pair | Price | Quantity
-- | -- | -- | -- | -- | -- | --
TJZAEP-FPC7J-6VFPRB | 1560893096740 | 06-18-19 21:24:56 | sell/market | BCH/EUR | € 366.40 | 0.01
TN4UDE-GEXR2-HYMBRA | 1560893089389 | 06-18-19 21:24:49 | buy/market | BCH/EUR | € 366.50 | 0.01
T7CDWJ-EDVRC-4RQWZQ | 1560893066091 | 06-18-19 21:24:26 | sell/market | BCH/EUR | € 366.00 | 0.007
TQDQHD-3VZVV-KA67YK | 1560893058067 | 06-18-19 21:24:18 | buy/market | BCH/EUR | € 366.50 | 0.007
TY5PSH-OJALS-5ZAVOC | 1560893050784 | 06-18-19 21:24:10 | sell/market | BCH/EUR | € 365.70 | 0.007
TJAT6V-VTASQ-G5WOKH | 1560893044339 | 06-18-19 21:24:04 | buy/market | BCH/EUR | € 366.00 | 0.007
TQPAPV-FQHAD-S7UWDL | 1560893038586 | 06-18-19 21:23:58 | sell/market | BCH/EUR | € 365.70 | 0.007
T652ON-XWOK5-AGL2U7 | 1560893031653 | 06-18-19 21:23:51 | buy/market | BCH/EUR | € 365.80 | 0.007
TA6CBD-A3OVV-ANE2IX | 1560893026030 | 06-18-19 21:23:46 | sell/market | BCH/EUR | € 365.60 | 0.007
TSIFMS-22CSF-YPKP5I | 1560893020487 | 06-18-19 21:23:40 | buy/market | BCH/EUR | € 365.80 | 0.007
[...]

PR #3127, opened 2019-06-27
[Kraken] Changed deafult sorting to timestamps rather than ids

gregoreficint:
The function getTradeHistory(TradeHistoryParams params) in KrakenTradeService sorts the trades by ID. However, the ids in Kraken have no natural order so the trade history should be sorted by timestamp as default. The ids have the format of A-B-C which A, B, C being strings. I believe they refer to individual ledger entries in Kraken.

Fixes issue #3126.


gregoreficint:
@timmolter Could you please check the Travis CI? I believe the CI itself failed for Java 11 not the change. Please let me know if I am wrong. Thank you!

Issue #1816, opened 2017-10-15
Multi-Value Deposit Addresses


digitalchaotics:
The current AccountService API does not accommodate multi-value deposit addresses.  For example on Bittrex, to deposit XMR you are required to provide two values:  the "Payment_id" and the "Base Address".  In another case (BTS) the two values are labeled "Memo" and "Registered Acct".  This issue impacts 18 coins on Bittrex.  I'm not aware of this issue on other exchanges.

This issue impacts the utility of two AccountService API methods:  'withdrawFunds' and 'requestDepositAddress'.  Neither method provides a way to use multi-value addresses.

One solution to this issue is to expand the API to include a DepositAddress class which would be accepted by 'withdrawFunds' and returned by a new 'getDepositAddress' method, e.g. ...

- public DepositAddress getDepositAddress(Currency currency);
- public String withdrawFunds(Currency currency, BigDecimal amount, DepositAddress address);

Suggestions?  Maybe there's a workaround I'm not aware of?  Has anyone else faced this issue?

npomfret:
This is a bit of a work in progress.  I started to add support for _complex_ addresses when I added the support for XRP withdrawals, see `RippleWithdrawFundsParams`.

I don't think there are any other implementations yet.

digitalchaotics:
Yes - the issue seems to be everywhere for XRP, but I've seen additional examples.  Lots of them on Bittrex, but I've also seen it on other exchanges.

npomfret:
I'd be happy to work on this stuff but it won't be for a couple of weeks (new puppy means I don't have time to wash let alone work!)

tmmrt:
is there any update on requestDepositAddress on accountService? I would like to request a usable address that has the depositaddress as well as tag/paymentid for XRP, XMR etc.

amjad1210:
I need this feature myself so will create a PR.

PR #3139, opened 2019-07-17
Binance & Coinbase destination tag support

amjad1210:
#1816


amjad1210:
@timmolter Any chance you can merge this soon?

timmolter:
can you fix the errors please?
<https://travis-ci.org/knowm/XChange/jobs/560157055>

amjad1210:
> can you fix the errors please?
> https://travis-ci.org/knowm/XChange/jobs/560157055

Fixed.

xchange-core/src/main/java/org/knowm/xchange/dto/account/DepositAddress.java
```diff
@@ -0,0 +1,50 @@
+package org.knowm.xchange.dto.account;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+public final class DepositAddress implements Serializable {
```
walec51: we already have a very generic interface WithdrawFundsParams which you can extend to add a tag 
we also already have this address in DefaultWithdrawFundsParams which already has the address
we also already have RippleWithdrawFundsParams that adds a tag for ripple and MoneroWithdrawFundsParams with add paymentId <- this is a little unfortunate as they are in essence the same this

please remove DepositAddress , add destinationTag to DefaultWithdrawFundsParams with the following documentation:

```
  /**
   * Various cryptocurrencies allow to add an additional identifier / tag to you transaction which helps them
   * to process you payment. Ripple for example has a tag, Monero has a payment id.
   */
  public String getDestinationTag() {
```
and us it instead in Binance and Coinbase

in another PR I'll then refactor the code to deprecate RippleWithdrawFundsParams and MoneroWithdrawFundsParams and add some better WithdrawFundsParams  interfaces

walec51: I would rename this to `AddressWithTag` as it is used both for deposits and withdrawals

xchange-core/src/main/java/org/knowm/xchange/dto/account/FundingRecord.java
```diff
@@ -18,6 +18,9 @@
   /** Crypto currency address for deposit/withdrawal */
   private final String address;
 
+  /** Crypto currency destination tag for deposit/withdrawal */
+  private String destinationTag;
```
walec51: final

xchange-core/src/main/java/org/knowm/xchange/dto/account/DepositAddress.java
amjad1210: Done.

xchange-core/src/main/java/org/knowm/xchange/dto/account/FundingRecord.java
amjad1210: Fixed.

PR #3140, opened 2019-07-17
[coindeal] more meta-data and bettter exception handling

walec51:


makarid:
Thanks for this update @walec51

PR #3154, opened 2019-07-30
[bitfinex] refactoring to use both api's v1 and v2 at the same time

andre77:
the raw services like `org.knowm.xchange.bitfinex.service.BitfinexMarketDataServiceRaw` should be able to use both api's v1 _and_ v2 at the same time
now it is the case


walec51:
+1 for the general direction, I think that this is how we should do API migration in all Xchange modules, its much better then exposing two Exchanges for each API version

you need to fix xchange-examples still

PR #3163, opened 2019-08-07
[Core] Fix order DTO builders

Nephtys:
Fix order builders:
- missing fee/cumulativeAmount copy in Builder.from methods
- MarketOrder.fee method was setting cumulativeAmount instead of fee
- missing javadoc
- missing tests


walec51:
+ good fix

PR #3198, opened 2019-09-02
[binance] better order mapping

walec51:


walec51:
@milczarekIT, @makarid - could you take a look at my code?
I don't want to merge this PR without any review

amjad1210:
@walec51 sorry for the late response I took a look over this and it looks really good. I'm going to use this as a guideline and do something similar with CoinbasePro.

milczarekIT:
Also I saw your message after your merge

PR #3199, opened 2019-09-03
Removed Abucoins exchange which is no longer operating

milczarekIT:
Removed Abucoins exchange which is no longer operating: https://abucoins.com/en


walec51:
isn't their API still available to download old users trades?

milczarekIT:
Hi Adam,
I am not sure, because I do not have account here and I am not able to open it, because this exchange is dead. 'Register' button is still available, but it causes following error: "Due to company liqiduation process registrations are punt on hold."

(Public) API to download historical trades or order books does not exist.

makarid:
I believe it is better to delete it. The exchange had ultra low volume even when it operated so not much damage will be done if we delete it.

timmolter:
thanks!

PR #3201, opened 2019-09-04
[Core] Adding new PriceUtility class for tickRounding

makarid:


walec51:
-1 no tests, this is just rounding to a given decimal point?
please learn how to use BigDecimal for that

makarid:
Hello @walec51 i will add a test in a sec. This method is not just rounding. It rounds the price with specific tickSize. For example if you put tickSize = 0.25 as param, the method will always round the price to the nearest of the following decimals[ 0.0, 0.25, 0.5, 0.75]. Only these 4 decimals suffixes can be used on the price.

walec51:
ok, then its something like stepSize which is used by Binance also for order amounts and we also store this in CurrencyPairMetaData

because of this lets make this method more generic and use BigDecimal for calculations instead of primitives like double:

```
public class BigDecimalUtil {

  public static BigDecimal roundToStepSize(BigDecimal value, BigDecimal stepSize) {
    return roundToStepSize(value, stepSize, RoundingMode.HALF_DOWN);
  }

  public static BigDecimal roundToStepSize(BigDecimal value, BigDecimal stepSize, RoundingMode roundingMode) {
    BigDecimal divided = amount.divide(stepSize, MathContext.DECIMAL32)
                .setScale(0, roundingMode);
    return divided.multiply(stepSize, MathContext.DECIMAL32).stripTrailingZeros();
  }
}
```

makarid:
So do you want me to change double into BigDecimal?

walec51:
replace your PriceUtils with my BigDecimalUtils

walec51:
see this: https://github.com/knowm/XChange/pull/3203

PR #3203, opened 2019-09-04
[core] BigDecimalUtils with rounding to step size

walec51:
functionality suggested by @makarid  in https://github.com/knowm/XChange/pull/3201

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/utils/Assert.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/utils/Assert.java
@@ -64,4 +64,8 @@ public abstract class Assert {
       throw new IllegalArgumentException(message);
     }
   }
+
+  static void ass(boolean b) {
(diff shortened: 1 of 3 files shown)


makarid:
Thanks @walec51

walec51:
if this is ok with you then I'll merge

I'll also use it in my OrderValuesAdjuster for order amounts (I plan to contribute it next week)

makarid:
Yeah it is ok. I will add it to Bitmex placeOrder and Itbit placeOrder method also when you merge

walec51:
@makarid I don't think that rounding the orders limit price / amount in placeOrder is a good idea.

You should handle this when building the order by either:
a) validating the input data in accordance to CurrencyPairMetaData
b) adjust the values then constructing an order
I'll submit a PR with a helper class for that today.

walec51:
@makarid this helper will solve your problem at the proper stage - while building the order, not when placing it: https://github.com/knowm/XChange/pull/3204

PR #3204, opened 2019-09-06
[core] new OrderValuesHelper to validate and / or adjust order values

walec51:
new OrderValuesHelper to validate and / or adjust order values based on CurrencyPairMetaData

it also introduces a new style (behavioral) of writing tests in OrderValuesHelperTest - tell me if you like it


makarid:
Nice PR @walec51. Thanks

Issue #3202, opened 2019-09-04
[Feature] Adding minor support for margin trading on Wallet.class


makarid:
Hello, i would like to add 3 optional fields in the Wallet class in order to have at least some support for margin trading. 
The fields can be: 
1) boolean isMarginSupported
2) long maxLeverage
3) double currentLeverage.

What do you think about that? Thanks

makarid:
@timmolter @walec51 what do you think?

walec51:
why would those attributes be assigned to a wallet?
you want add information about "margin trading" or "margin lending"?

makarid:
Thanks for the response @walec51 .Yes i would like the library to have support for at least some basic margin trading functionality

walec51:
you did not answer my questions

makarid:
Sorry, about margin trading

makarid:
I believe this feature is very small and will not break anything. It is a quick fix until we can fully support margin trading.

walec51:
ok, is it like that on most exchanges?
with which ones to you have experience with in margin trading?

PS. Small features added to the core API become unrefactorable without braking backwards  compatibility really fast. Because of this we have to think them through.

walec51:
on exchanges that support multiple wallets can you have one wallet with margin trading and one without it?

isn't the maxLeverage global for the entire exchange?

makarid:
At least the following exchanges offer margin trading: Bitfinex,Bitmex,Kraken,Kraken futures(crypto facilities),liquid. Maybe there are more which i don't know or use them. 
About your backward compatibility issue maybe then it would be better to create a new class for example MarginWallet which extends Wallet and add all the margin features there.

makarid:
> on exchanges that support multiple wallets can you have one wallet with margin trading and one without it?
> 
> isn't the maxLeverage global for the entire exchange?

Yes we can have 2 wallets. The maxLeverage is global for the entire exchange. If we go with this approach we need to create 2 constant variables(or an enum) for the Wallet.name attribute in order to have specific names (for example "spot","margin"), so we can distinct which one is.

walec51:
I'm not asking which exchanges offer margin trading. I'm asking on which of them did you margin trading your self and learned how it works there.

Adding MarginWallet to xchange-core is still extending the API.

We already have accountInfo.getWallet(String id). Unfortunately this id is exchange specific as some exchanges like Bitfinex wale separate wallets for holding, trading and lending (remember that margin trading != margin lending).

[...]

PR #3206, opened 2019-09-07
[Core] New Wallet feature

makarid:
Hello, i have implement the new Wallet feature that we have discussed on the #3202 .  I have implement it this way because i don't want to have many compatibility issues at the beginning. Until more people find it useful and start to implement it on other exchanges it is best to not break the previous implementation. Please check it and tell me if you like it.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
@@ -1,12 +1,9 @@
 package org.knowm.xchange.dto.account;
 
 import java.io.Serializable;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Objects;
+import java.math.BigDecimal;
+import java.util.*;
+
 import org.knowm.xchange.currency.Currency;
 
 /**
@@ -24,6 +21,12 @@ public final class Wallet implements Serializable {
   private String id;
   /** A descriptive name for this wallet. Defaults to {@link #id} */
   private String name;
+  /** All wallet operations*/
+  private final Set<WalletFeature> walletFeatures;
+  /** Maximum leverage that wallet supports*/
+  private BigDecimal maxLeverage = BigDecimal.ZERO;
+  /** Current leverage of wallet*/
+  private BigDecimal currentLeverage = BigDecimal.ZERO;
 
   /**
    * Constructs a {@link Wallet}.
@@ -31,8 +34,11 @@ public final class Wallet implements Serializable {
    * @param id the wallet id
    * @param name a descriptive name for the wallet
    * @param balances the balances, the currencies of the balances should not be duplicated.
+   * @param walletFeatures all the features that wallet supports
+   *
+   * maxLeverage and currentLeverage are BigDecimal.ZERO for the default constructor
    */
-  public Wallet(String id, String name, Collection<Balance> balances) {
+  public Wallet(String id, String name, Collection<Balance> balances,Set<WalletFeature> walletFeatures) {
 
     this.id = id;
     if (name == null) {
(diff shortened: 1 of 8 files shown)


xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
```diff
@@ -24,15 +20,24 @@
   private String id;
   /** A descriptive name for this wallet. Defaults to {@link #id} */
   private String name;
+  /** All wallet operations */
+  private final Set<WalletFeature> walletFeatures;
```
walec51: rename attribute to `features` and getter to `getFeatures()`

further more I think this attribute should never be null
for constructors that don't have it in the args I think we can safely assume the following features: FUNDING, TRADING <- as this is the most common case

xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
```diff
@@ -24,15 +20,24 @@
   private String id;
   /** A descriptive name for this wallet. Defaults to {@link #id} */
   private String name;
+  /** All wallet operations */
```
walec51: /** Features supported by this wallet */

xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
```diff
@@ -24,15 +20,24 @@
...
+  private final Set<WalletFeature> walletFeatures;
+  /** Maximum leverage that wallet supports */
```
walec51: /** Maximum leverage for margin trading supported by this wallet */

xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
```diff
@@ -24,15 +20,24 @@
...
+  private BigDecimal maxLeverage = BigDecimal.ZERO;
+  /** Current leverage of wallet */
```
walec51: /** Current leverage for margin trading done on this wallet */

xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
```diff
@@ -24,15 +20,24 @@
...
    * Constructs a {@link Wallet}.
    *
    * @param id the wallet id
    * @param name a descriptive name for the wallet
    * @param balances the balances, the currencies of the balances should not be duplicated.
+   * @param walletFeatures all the features that wallet supports
+   *     <p>maxLeverage and currentLeverage are BigDecimal.ZERO for the default constructor
    */
-  public Wallet(String id, String name, Collection<Balance> balances) {
+  public Wallet(
```
walec51: these constructors are becoming a mess
maybe we should add a Builder class and deprecate them all?

xchange-core/src/main/java/org/knowm/xchange/dto/account/WalletFeature.java
```diff
@@ -0,0 +1,12 @@
+package org.knowm.xchange.dto.account;
+
+public enum WalletFeature {
```
walec51: Lest make this an inner enum of the Wallet class

xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
makarid: Totally agree

xchange-core/src/main/java/org/knowm/xchange/dto/account/Wallet.java
makarid: Can i add a new "getter" in order to return the default features?

walec51: I don't understand how a new getter helps with the problem of getFeatures() returning null in all the current modules

makarid:
I have implement the solution with your new insight and i really like it. It is MUCH better. I will make a PR shortly.

makarid:
With this new Builder pattern you can easily add whatever feature you want in the Wallet.class and i believe it would be much simpler to extend new functionality in the future if needed.

timmolter:
Looking good. I think this warrants a version bump to 4.5.0.

makarid:
Thanks for the reply @timmolter.

makarid:
Can we merge this?

timmolter:
I'll bump the version.

PR #3209, opened 2019-09-08
Add dynamic trading fee API support for cexio

BenjaminTrapani:


xchange-cexio/src/main/java/org/knowm/xchange/cexio/CexIOAdapters.java
```diff
@@ -356,4 +360,20 @@ private static CurrencyPair adaptCurrencyPair(String pair) {
     // Currency pair is in the format: "BCH:USD"
     return new CurrencyPair(pair.replace(":", "/"));
   }
+
+  private static Fee adaptFeeDetails(FeeDetails feeDetails) {
+    // It might be worth expanding the Fee structure in xchange-core to contain both buy and sell
+    // fees
```
walec51: interesting, does any other exchange have this type of separation of fees?

PR #3210, opened 2019-09-09
fix when check server time with proxy setting

nicqchen:


walec51:
build using maven to fix formatting and push it to this PR

timmolter:
please see: https://github.com/knowm/XChange/wiki/Code-Style. We can then see more clearly what your changes are. Thanks!

PR #3215, opened 2019-09-12
[Kraken] Adding Kraken margin wallet support

makarid:


makarid:
Is it OK now?

xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenAccountService.java
```diff
@@ -38,9 +37,22 @@ public KrakenAccountService(Exchange exchange) {
...
+
+    Wallet marginWallet =
+        Wallet.Builder.from(tradingWallet.getBalances().values())
+            .id("margin")
+            .features(EnumSet.of(Wallet.WalletFeature.FUNDING, Wallet.WalletFeature.MARGIN_TRADING))
+            .maxLeverage(BigDecimal.valueOf(5))
+            .currentLeverage(
+                krakenTradeBalanceInfo
+                    .getCostBasis()
+                    .divide(krakenTradeBalanceInfo.getTradeBalance(), MathContext.DECIMAL32))
```
walec51: just one thing to consider: can getTradeBalance return zero?
if so this will fails

makarid: Eagle eye! I believe it is possible if you don't have any funds on the exchange. I will check for zero value and return zero.

makarid:
I have fix the divide by zero issue.

PR #3216, opened 2019-09-13
[Binance,CoinbasePro] Added cumulativeQuoteAmount field into Order

amjad1210:
- Added cumulativeQuoteAmount which is the quote amount which is filled against the order.

Binance
- Set correct stopPrice on StopOrder.

CoinbasePro
- Refactored adaptOrder method.

@walec51

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
@@ -136,6 +136,13 @@ public abstract class Order implements Serializable {
     this.cumulativeAmount = cumulativeAmount;
   }
 
+  public BigDecimal getCumulativeCounterAmount() {
+    if (cumulativeAmount != null && averagePrice != null && averagePrice.compareTo(BigDecimal.ZERO) > 0) {
+      return cumulativeAmount.multiply(averagePrice);
+    }
+    return null;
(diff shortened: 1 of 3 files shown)


xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
```diff
@@ -86,6 +88,31 @@ public Order(
...
 
+  public Order(
+      OrderType type,
+      BigDecimal originalAmount,
+      CurrencyPair currencyPair,
+      String id,
+      Date timestamp,
+      BigDecimal averagePrice,
+      BigDecimal cumulativeAmount,
+      BigDecimal cumulativeQuoteAmount,
```
walec51: I don't think we need to store this value is easly calculated: `cumulativeAmount.multiply(averagePrice, ...)`
it would be more universal for all modules to add a convinance method that calculates this on the fly if cumulativeAmount and averagePrice are not null

walec51: besides that we don't use the term "quote", we call it "counter"

amjad1210: @walec51 Thanks for clarifying pushed new changes.

PR #3217, opened 2019-09-13
 [poloniex] Added new Transfer-Object: PoloniexAdjustment

Yampersand:
Added new Type PoloniexAdjustment to the PoloniexDepositsWithrawalsResponse-Object and changed adapter to convert it to a Deposit-FundingRecord

API mothod for AccountService.getFundingHistory returns an additional type named Adjustment, which can be airdrops other other forms of "gifted" funds:
[Documentation](https://docs.poloniex.com/#returndepositswithdrawals)

This commit adds the new Type to the PoloniexDepositsWithdrawalsResponse and proposes a non-invasive way to adapt it to a Deposit-FundingRecord. It (miss)uses the description field to place all the extra information until there is a better solution to handle this kind of deposit variation.


Yampersand:
To have a better way to adapt this data, I made an proposal in this issue #3218

PR #3223, opened 2019-09-15
[Bitmex] Reverting back stepSize on bitmex placeOrder

makarid:
This stepSize works only for XBT/USD pair. I am sorry for this mistake on the implementation.


walec51:
+1 adjusting order values should be done while building it based on CurrencyPairMetaData

PR #3224, opened 2019-09-15
[Core] Replace ioutils.closequitely with try-with-resources

makarid:
Hello, can you please tell me if it is ok, the way i have implemented? Thanks

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/BaseExchange.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/BaseExchange.java
@@ -74,26 +74,24 @@ public abstract class BaseExchange implements Exchange {
     if (this.exchangeSpecification.getMetaDataJsonFileOverride()
         != null) { // load the metadata from the file system
 
-      try(InputStream is = new FileInputStream(this.exchangeSpecification.getMetaDataJsonFileOverride())) {
+      try (InputStream is =
+          new FileInputStream(this.exchangeSpecification.getMetaDataJsonFileOverride())) {
         loadExchangeMetaData(is);
       } catch (IOException e) {
-        logger.error(
-            "An exception occured while loading the metadata file from the classpath. It may lead to unexpected results, so it's better to address it.",
-            e);
+        throw new ExchangeException(e);
       }
 
     } else if (this.exchangeSpecification.getExchangeName()
         != null) { // load the metadata from the classpath
 
-      try(InputStream is = BaseExchangeService.class
+      try (InputStream is =
+          BaseExchangeService.class
               .getClassLoader()
               .getResourceAsStream(getMetaDataFileName(this.exchangeSpecification) + ".json")) {
 
         loadExchangeMetaData(is);
-      } catch (IOException e){
-        logger.error(
-                "An exception occured while loading the metadata file from the classpath. It may lead to unexpected results, so it's better to address it.",
-                e);
+      } catch (IOException e) {
+        throw new ExchangeException(e);
       }
 
     } else {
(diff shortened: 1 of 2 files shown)


xchange-core/src/main/java/org/knowm/xchange/BaseExchange.java
```diff
@@ -76,30 +76,26 @@ public void applySpecification(ExchangeSpecification exchangeSpecification) {
     if (this.exchangeSpecification.getMetaDataJsonFileOverride()
         != null) { // load the metadata from the file system
 
-      InputStream is = null;
-      try {
-        is = new FileInputStream(this.exchangeSpecification.getMetaDataJsonFileOverride());
+      try(InputStream is = new FileInputStream(this.exchangeSpecification.getMetaDataJsonFileOverride())) {
         loadExchangeMetaData(is);
-      } catch (FileNotFoundException e) {
+      } catch (IOException e) {
```
walec51: I don't think that if the metadata file is malformed then only a warn in the logs is enough, I would prefer the exception to get out

xchange-core/src/main/java/org/knowm/xchange/BaseExchange.java
```diff
@@ -76,30 +76,26 @@ public void applySpecification(ExchangeSpecification exchangeSpecification) {
...
+              .getClassLoader()
+              .getResourceAsStream(getMetaDataFileName(this.exchangeSpecification) + ".json")) {
+
         loadExchangeMetaData(is);
-      } finally {
-        IOUtils.closeQuietly(is);
+      } catch (IOException e){
+        logger.warn(
+                "An exception occured while loading the metadata file from the classpath. This is just a warning and can be ignored, but it may lead to unexpected results, so it's better to address it.",
+                e);
```
walec51: same here

xchange-core/src/main/java/org/knowm/xchange/BaseExchange.java
makarid: Do you mean to log the change the log level to error and log the error?

makarid: This: log.error(e.getMessage(),e);

walec51: no, I would prefer the IOException to get thrown out of this method into my application code

makarid: when i am trying to throw the exception in the catch block, i get "Unhandled exception: IOException error". So i don't understand how else i can throw the exception? Do i need to add throws IOException on the method?

walec51: ok, I didn't notice that applySpecification has no throws IOException

you have an example how to handle this at the end of the `applySpecification' method

```
   catch (IOException e) {
        throw new ExchangeException(e);
      }
```

makarid: Thanks a lot for your response @walec51 ,although i have not understand how to do it? Do you want to write a TODO in order to add it later?

walec51: just replace those two catches with this code:

```
 catch (IOException e) {
        throw new ExchangeException(e);
      }
```

you have to do a simple copy-paste, I don't know how to explain this simpler...

makarid: xaxaxaxa. Sorry man, i thought it would throw the same error. I am making the push.

PR #3232, opened 2019-09-19
[acx] Added support for trade history, updated metadata.

nielsdraaisma:
- Updated metadata
- Implemented getTradeHistory
- Fixed bug that failed to convert btcusdt to the correct currencypair

### DIFF
--- a/xchange-acx/src/main/java/org/knowm/xchange/acx/AcxApi.java
+++ b/xchange-acx/src/main/java/org/knowm/xchange/acx/AcxApi.java
@@ -2,12 +2,7 @@ package org.knowm.xchange.acx;
 
 import java.io.IOException;
 import java.util.List;
-import javax.ws.rs.FormParam;
-import javax.ws.rs.GET;
-import javax.ws.rs.POST;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
+import javax.ws.rs.*;
 import javax.ws.rs.core.MediaType;
 import org.knowm.xchange.acx.dto.AcxTrade;
 import org.knowm.xchange.acx.dto.account.AcxAccountInfo;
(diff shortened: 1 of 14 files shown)


xchange-acx/src/main/java/org/knowm/xchange/acx/AcxApi.java
```diff
@@ -59,6 +54,52 @@ AcxOrderBook getOrderBook(
...
+   *
+   * @param market Unique market id. It's always in the form of xxxyyy, where xxx is the base
+   *     currency code, yyy is the quote currency code, e.g. 'btcaud'. All available markets c an be
+   *     found at /api/v2/markets.
+   */
+  @GET
+  @Path(
+      "/trades/my.json?market={market}&access_key={access_key}&tonce={tonce}&signature={signature}&limit={limit}&order={order}")
+  List<AcxTrade> getMyTrades(
+      @PathParam("access_key") String accessKey,
```
walec51: use one method with @QueryParam instead of @PathParam
you can pass null if you do not want it in the request so one method will be enough

xchange-acx/src/main/java/org/knowm/xchange/acx/AcxMapper.java
```diff
@@ -133,4 +157,28 @@ public String getOrderType(OrderType type) {
...
+        .id(trade.id)
+        .orderId(trade.orderId)
+        .price(trade.price)
+        .originalAmount(trade.volume)
+        .timestamp(trade.createdAt)
+        .type(mapOrderSide(trade.side))
+        .build();
+  }
+
+  private Map<String, CurrencyPair> marketMap = new ConcurrentHashMap<>();
```
walec51: move this declaration to the beginning of he class

xchange-acx/src/main/java/org/knowm/xchange/acx/service/trade/AcxTradeService.java
```diff
@@ -91,13 +100,74 @@ public boolean cancelOrder(String orderId) throws IOException {
...
+    Long startId = null;
+    Long endId = null;
+    String order = null;
+    if (acxTradeHistoryParams.endTime != null) {
+      timestamp = acxTradeHistoryParams.endTime.toInstant().getEpochSecond();
+    }
+    String market = getAcxMarket(acxTradeHistoryParams.currencyPair);
+    //    if (acxTradeHistoryParams.endId != null) {
+    //      endId = Long.valueOf(acxTradeHistoryParams.endId);
+    //    }
```
walec51: why is this commented out?

xchange-acx/src/main/java/org/knowm/xchange/acx/AcxApi.java
nielsdraaisma: This wasn't working initially, turned out it required a small change to the digest code to make it work with a single method.

xchange-acx/src/main/java/org/knowm/xchange/acx/AcxMapper.java
nielsdraaisma: done

xchange-acx/src/main/java/org/knowm/xchange/acx/service/trade/AcxTradeService.java
nielsdraaisma: done since it's now using a single method.

PR #3234, opened 2019-09-19
Binance margin trading integration methods

golart:


xchange-binance/pom.xml
```diff
@@ -8,7 +8,8 @@
         <version>4.3.22-SNAPSHOT</version>
     </parent>
 
-    <artifactId>xchange-binance</artifactId>
+    <artifactId>xchange-binance-custom</artifactId>
+    <version>4.3.22.3</version>
```
walec51: almost all of the maven changes in this PR are unacceptable

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/BinanceNewOrder.java
```diff
@@ -2,8 +2,11 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigDecimal;
+import java.util.Date;
+import org.knowm.xchange.binance.BinanceAdapters;
+import org.knowm.xchange.dto.trade.LimitOrder;
 
-public final class BinanceNewOrder {
+public final class BinanceNewOrder extends LimitOrder {
```
walec51: binances order does not have the same field names as our LimitOrder, remove this extends
DTO are NOT mappers

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/BinanceOrderStatus.java
```diff
@@ -12,9 +12,9 @@
   EXPIRED;
 
   @JsonCreator
-  public static OrderStatus getOrderStatus(String s) {
+  public static BinanceOrderStatus getOrderStatus(String s) {
```
walec51: doesn't Jackson has a similar behaviour by default for enums if you do not suggest a @JsonCreator in them?
this seams unnecessary in all enums

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -99,30 +100,11 @@ public OpenOrders getOpenOrders(OpenOrdersParams params) throws IOException {
 
   @Override
   public String placeMarketOrder(MarketOrder mo) throws IOException {
-
-    return placeOrder(OrderType.MARKET, mo, null, null, null);
+    return Long.toString(placeOrder(OrderType.MARKET, mo, null, null, null).orderId);
   }
 
-  @Override
-  public String placeLimitOrder(LimitOrder lo) throws IOException {
```
walec51: you want to remove support for LimitOrders in binance ?!

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -99,30 +100,11 @@ public OpenOrders getOpenOrders(OpenOrdersParams params) throws IOException {
...
-
-      while (orderFlagsIterator.hasNext()) {
-        IOrderFlags orderFlag = orderFlagsIterator.next();
-        if (orderFlag instanceof TimeInForce) {
-          tif = (TimeInForce) orderFlag;
-        }
-      }
-    }
-    return placeOrder(type, lo, lo.getLimitPrice(), null, tif);
+  public BinanceNewOrder placeMarketOrder2(MarketOrder mo) throws IOException {
```
walec51: remove this from here, just use the *Raw method in your application code if you want to get binances DTO

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -153,35 +135,60 @@ public String placeStopOrder(StopOrder so) throws IOException {
...
       throw BinanceErrorAdapter.adapt(e);
     }
   }
 
+  @Override
+  public String placeLimitOrder(LimitOrder lo) throws IOException {
+    return Long.toString(placeLimitOrder2(lo).orderId);
+  }
+
+  public BinanceNewOrder placeLimitOrder2(LimitOrder lo) throws IOException {
```
walec51: same as above

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/service/BitfinexAdapters.java
```diff
@@ -103,7 +103,14 @@ private BitfinexAdapters() {}
...
   public static String adaptBitfinexCurrency(String bitfinexSymbol) {
-    return bitfinexSymbol.toUpperCase();
+    String currency = bitfinexSymbol.toUpperCase();
+    if (currency.equals("DSH")) {
+      currency = "DASH";
+    }
+    if (currency.equals("QTM")) {
+      currency = "QTUM";
+    }
+    return currency;
```
walec51: don't, until we have a global solution for unifying currency code exchanges modules should use code actually used by the implemented exchange

xchange-bitfinex/src/main/resources/bitfinex.json
```diff
@@ -46,6 +46,12 @@
       "trading_fee": 0.002
     },
     "DSH/BTC": {
+    "DASH/USD": {
```
walec51: this is not a valid JSON now

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexBaseService.java
```diff
@@ -14,11 +11,18 @@
...
+import java.util.Map;
+
 @SuppressWarnings({"WeakerAccess", "unused"})
 public class BitmexBaseService extends BaseExchangeService<BitmexExchange> implements BaseService {
 
   protected final BitmexAuthenticated bitmex;
   protected final ParamsDigest signatureCreator;
+  protected final String RATE_LIMIT_LIMIT = "X-RateLimit-Limit";
+  protected final String RATE_LIMIT_REMAINING = "X-RateLimit-Remaining";
+  protected final String RATE_LIMIT_RESET = "X-RateLimit-Reset";
```
walec51: this should be static also

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexDigest.java
```diff
@@ -27,10 +32,64 @@ public static BittrexDigest createInstance(String secretKeyBase64) {
...
+    String invocationUrl = getInvocationUrl(restInvocation);
     Mac mac = getMac();
     mac.update(invocationUrl.getBytes());
 
     return String.format("%0128x", new BigInteger(1, mac.doFinal()));
   }
+
+  // copy paste methods from si.mazi.rescu.RestInvocation
+  static String getInvocationUrl(RestInvocation restInvocation) {
+    // HARDCODED API URL. Reason: evade INVALID_SIGNATURE error. Url used as digest payload
```
walec51: when does this INVALID_SIGNATURE happen?

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
```diff
@@ -47,6 +47,14 @@ public String placeMarketOrder(MarketOrder marketOrder) throws IOException {
     }
   }
 
+  public LimitOrder placeMarketOrder2(MarketOrder marketOrder) throws IOException {
```
walec51: placing a market order results in a limit order ?

xchange-exmo/src/main/java/org/knowm/xchange/exmo/dto/meta/ExmoCurrencyPairMetaData.java
```diff
@@ -0,0 +1,38 @@
+package org.knowm.xchange.exmo.dto.meta;
+
+import java.math.BigDecimal;
+import org.knowm.xchange.dto.meta.CurrencyPairMetaData;
+import org.knowm.xchange.dto.meta.FeeTier;
+
+/** @author ujjwal on 26/02/18. */
+public class ExmoCurrencyPairMetaData extends CurrencyPairMetaData {
```
walec51: DTOs are not mappers and those two aspects should never be mixed

xchange-exmo/src/main/java/org/knowm/xchange/exmo/service/ExmoAdapters.java
```diff
@@ -52,14 +55,32 @@ public static Balance adaptBalance(
...
 
   public static Ticker adaptTicker(CurrencyPair currencyPair, Map<String, String> data) {
+
+    final BigDecimal last = new BigDecimal(data.get("last_trade"));
+    final BigDecimal buyPrice = new BigDecimal(data.get("buy_price"));
+
+    BigDecimal priceChange = null;
+    BigDecimal priceChangePercent = null;
+
+    if (ExmoMarketDataHolder.getCloseBuyPrice(currencyPair) != null) {
```
walec51: this assumes that some one is adapting ticker in a specified order and will fail in multi threaded applications or any other use case that does not respect this hidden requirement

xchange-binance/pom.xml
golart: fixed

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
golart: Not useful, maybe it would be better to make public current private method placeOrder? Cause it contains some business logic and exception handling. What do you think?

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexBaseService.java
golart: fixed

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
golart: yes, a market order is a special case of limit order

xchange-exmo/src/main/java/org/knowm/xchange/exmo/service/ExmoAdapters.java
golart: Agree with you, but there is no way to calc priceChange and priceChangePercent without holding data. And without that it works correctly only without this general ticker data fields

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/BinanceOrderStatus.java
golart: Do you mean revert specific naming for binance order status? Does it seem useful for coding?

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
golart: Look 167 line. It supports, but more useful as a specific case of calling of placeLimitOrder2, which return full new order object as a value

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexDigest.java
golart: Fort example if you use proxying. In this case signature creates with your proxy url, bur should be with exchange api url

xchange-binance/pom.xml
walec51: I still see strange repositories and custom artifacts in the poms

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/BinanceOrderStatus.java
walec51: I mean that if we remove this method, Jackson should use BinanceOrderStatus.valueOf by it self

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
walec51: line 167 don't change the fact that you are removing an implementation of a method defined in an interface which is the main API of this library and is used by most users

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
walec51: you can have public methods in *Raw clases
non *Raw classes should return / throw generic DTOs / exception
*Raw classes should return / throw exchange specific DTOs / exception

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexDigest.java
walec51: then all this code can be replaced with:
```
// define this in BinanceExchange
public static String DEFAULT_SSL_URI = "https://bittrex.com/api/"

// pass the exchangeSpecification to BittrexDigest constructor and then:
String invocationUrl = restInvocation.getInvocationUrl();
String digsetUrl = invocationUrl;
if (!invocationUrl.startsWith(DEFAULT_SSL_URI)) {
  String specUri = exchangeSpecification.getSslUri();
  digsetUrl = invocationUrl.replace(specUri, DEFAULT_SSL_URI);
}
```

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
walec51: no, its not, it has no limit price

xchange-exmo/src/main/java/org/knowm/xchange/exmo/service/ExmoAdapters.java
walec51: you should calculate this in your application code on a given list, NOT in a builder of a single ticker, the way you are implementing this is unacceptable and prone to errors

Issue #3226, opened 2019-09-17
version 4.4.0 OkexAccountService.getAccountInfo() error


suihs11:
version 4.4.0 OkexAccountService.getAccountInfo() throws an Exception
ExceptionMessage: duplicate wallets passed to AccountInfo

makarid:
With just getAccountInfo() method or with getAccountInfo().getWallet();

walec51:
full stack trace pls

sizmestiev:
Suppose the issue is in the copy+paste error: two `Wallet.WalletFeature.FUNDING` usages instead of `FUNDING`/`TRADING`.

https://github.com/knowm/XChange/blob/5b79b8ff103bc39622e7ee79260077520063770d/xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/service/OkexAccountService.java#L58

https://github.com/knowm/XChange/blob/5b79b8ff103bc39622e7ee79260077520063770d/xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/service/OkexAccountService.java#L61

makarid:
Yes it is

suihs11:
The full stack trace:

java.lang.IllegalArgumentException: duplicate wallets passed to AccountInfo
	at org.knowm.xchange.dto.account.AccountInfo.<init>(AccountInfo.java:104)
	at org.knowm.xchange.dto.account.AccountInfo.<init>(AccountInfo.java:77)
	at org.knowm.xchange.dto.account.AccountInfo.<init>(AccountInfo.java:114)
	at org.knowm.xchange.dto.account.AccountInfo.<init>(AccountInfo.java:40)
	at org.knowm.xchange.okcoin.v3.service.OkexAccountService.getAccountInfo(OkexAccountService.java:62)
	at io.renren.modules.exchangeApi.service.impl.ExchangeApiServiceImpl.callExchangeCash(ExchangeApiServiceImpl.java:115)
	at io.renren.modules.exchangeApi.task.CollectMethod.doTask1(CollectMethod.java:89)
	at io.renren.modules.exchangeApi.task.CollectMethod$$FastClassBySpringCGLIB$$ef5d846.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:747)
[...]

PR #3235, opened 2019-09-20
Fixing issue #3226

makarid:


walec51:
this will still fail, restore the wallet ids that your commit removed
just copy paste and test the code given at: https://github.com/knowm/XChange/issues/3226#issuecomment-533464699

walec51:
did you build this change using maven and test it?

makarid:
Now i did.Thanks for the reminder.

Issue #3238, opened 2019-09-23
[Binance] Fee missing from query order


amjad1210:
Currently when you request query orders the fee is always null. It seems the only time you can get the fee from Binance is when you create a new order with the parameter newOrderRespType set to FULL and the order fills immediately. https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#new-order--trade

But if you query the order using the GET request the fills are missing so there's no way to get the fee. https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#query-order-user_data

The only workaround I can see is requesting data from https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#account-trade-list-user_data and filtering the orderId to match an order.

I really need the total fee taken from the exchange for the order.

@walec51 Any input on this? 

Will create a PR for review.

walec51:
This is a more general problem.

Orders from Binance don't provide fees because Binances API does not provide it on the orders level.

On other exchanges orders can even have the `averagePrice == null` because this information is not present on the orders level there.

For both fields we can choose one of the following solutions:

1. Add documentation to those fields that they can be null if the underlining exchange does not provide them on an order level. To remedy this situation we can also add a new method to `TradeService` called `<T extends Order> T assureTradeResultsFilled(T order)` which would make the additional effort fetch trades and calculate those fields if they are missing in the order. One would have to explicitly call it to indicated that he is prepared to make that effort.

2. Calculate those fields for every order if the underlining exchange does not provide them. This may be a very unoptimised solution for people that don't needed them on the order because they always download all their trades and base their decisions on that. This problem is escalated by the fact that exchanges limit call rates aggressively.

So this is a case of ease of use vs optimisation. I would like to hear from more people which direction is preferable.

Ben21k:
Does Anybody know how to get the referral friend code (not my code) by Binance API?

amjad1210:
@walec51 I like the proposal of adding a new method which does extra checks to calculate missing values only when requested. The current methods will stay the same with no extra overhead unless the user explicitly call it to indicate tradeResultsFilled == true. Which would do any extra checks for orders to populate missing data. This sounds like the best solution moving forward across all exchanges.

amjad1210:
@walec51 I've implemented, tested and pushed a commit to solution 1. Please review the changes.

[...]

PR #3239, opened 2019-09-23
[Binance] Add order fees

amjad1210:
#3238

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
@@ -2,11 +2,7 @@ package org.knowm.xchange.binance.service;
 
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import org.knowm.xchange.Exchange;
 import org.knowm.xchange.binance.BinanceAdapters;
@@ -371,5 +367,58 @@ public class BinanceTradeService extends BinanceTradeServiceRaw implements Trade
     } catch (BinanceException e) {
       throw BinanceErrorAdapter.adapt(e);
     }
+  }
+
+  @Override
+  public Collection<Order> assureTradeResultsFilled(Collection<Order> orders) throws IOException {
+    // Group orders by currencyPair
+    Map<CurrencyPair, List<Order>> ordersGrouped =
+        orders.stream().collect(Collectors.groupingBy(Order::getCurrencyPair));
+
+    LinkedList<UserTrade> trades = new LinkedList<>();
+    for (Map.Entry<CurrencyPair, List<Order>> entry : ordersGrouped.entrySet()) {
+      List<Order> pairOrders = entry.getValue();
+      // Get oldest timestamp from orders
+      Optional<Date> oldestOrderTimestamp =
+          pairOrders.stream().map(Order::getTimestamp).min(Date::compareTo);
+
+      int limit = 1000; // Default 500; max 1000.
+      BinanceTradeHistoryParams binanceTradeHistoryParams =
+          new BinanceTradeHistoryParams(entry.getKey());
+      binanceTradeHistoryParams.setLimit(limit);
+      oldestOrderTimestamp.ifPresent(binanceTradeHistoryParams::setStartTime);
+
+      // Iterate to make sure we get all trades
+      while (true) {
+        UserTrades userTrades = getTradeHistory(binanceTradeHistoryParams);
+        List<UserTrade> userTradeList = userTrades.getUserTrades();
+        trades.addAll(userTradeList);
+        if (userTradeList.size() < limit) {
+          break;
+        }
+
+        // Get last fromId and remove from our list to prevent duplicates
+        String fromId = trades.removeLast().getId();
+        binanceTradeHistoryParams.setStartId(fromId);
+        binanceTradeHistoryParams.setStartTime(null);
+      }
+    }
+
+    // Group trades by orderId
+    Map<String, List<UserTrade>> tradesGrouped =
+        trades.stream().collect(Collectors.groupingBy(UserTrade::getOrderId));
+    orders.forEach(
+        o -> {
+          List<UserTrade> orderTrades = tradesGrouped.get(o.getId());
+          if (orderTrades != null) {
+            // Get total fee for order
+            BigDecimal fee =
+                orderTrades.stream()
+                    .map(UserTrade::getFeeAmount)
+                    .reduce(BigDecimal.ZERO, BigDecimal::add);
+            o.setFee(fee);
+          }
+        });
+    return orders;
   }
 }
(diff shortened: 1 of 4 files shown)


xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -367,9 +363,52 @@ public OpenOrdersParams createOpenOrdersParams() {
...
+  private Collection<Order> addFees(Collection<Order> orders) {
+    List<BinanceTrade> binanceTrades = new ArrayList<>();
+    orders.stream()
+        .map(Order::getCurrencyPair)
+        .distinct() //Optimization to reduce overhead
+        .forEach(
+            c -> {
+              try {
+                binanceTrades.addAll(
+                    super.myTrades(
```
walec51: this will not work for older orders

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -367,9 +363,52 @@ public OpenOrdersParams createOpenOrdersParams() {
...
+                        null,
+                        null,
+                        null,
+                        null,
+                        (Long)
+                            exchange
+                                .getExchangeSpecification()
+                                .getExchangeSpecificParametersItem("recvWindow"),
+                        getTimestamp()));
+              } catch (IOException e) {
```
walec51: this is like the worst way you could handle an error
leaving the fields you think should always be filled null and a stack trace outside of the logging system

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
amjad1210: Yeh older orders are also missing cummulativeQuoteQty from the API. Nothing we can do regarding older orders.

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
amjad1210: I wrote this up fairly quickly to see if it would work. Please review the changes again as I've rectified the issues.

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -372,4 +375,33 @@ public OpenOrdersParams createOpenOrdersParams() {
       throw BinanceErrorAdapter.adapt(e);
     }
   }
+
+  @Override
+  public Collection<Order> assureTradeResultsFilled(Collection<Order> orders) throws IOException {
```
walec51: if today you did 500 trades on a given currency pair and you pass orders from yesterday here then this will fail because by default binance gives you only 500 last trades

you should set `TradeHistoryParamsTimeSpan#setStartTime` to the oldest orders timestamp on a given currency pair and you should iterate using `TradeHistoryParamsIdSpan#fromId` if the trades count you receive is > limit

xchange-core/src/main/java/org/knowm/xchange/service/trade/TradeService.java
```diff
@@ -297,6 +297,22 @@ default void verifyOrder(MarketOrder marketOrder) {
...
+   * @return the order as it is on the exchange.
+   * @throws ExchangeException - Indication that the exchange reported some kind of error with the
+   *     request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the
+   *     requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the
+   *     requested function or data, but it has not yet been implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  default Collection<Order> getOrderAssureResultsFilled(String... orderIds) throws IOException {
```
walec51: change name to `getOrderWithTradeResults`

xchange-core/src/main/java/org/knowm/xchange/service/trade/TradeService.java
```diff
@@ -321,4 +337,37 @@ default void verifyOrder(MarketOrder marketOrder) {
...
+   * @return the order as it is on the exchange.
+   * @throws ExchangeException - Indication that the exchange reported some kind of error with the
+   *     request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the
+   *     requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the
+   *     requested function or data, but it has not yet been implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  default Collection<Order> getOrderAssureResultsFilled(OrderQueryParams... orderQueryParams)
```
walec51: change name to `getOrderWithTradeResults`

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
walec51: look at my recent comment, if I do a lot of trades event orders from today may fail to get filled

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
amjad1210: I'll look into a solution for this.

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -251,8 +247,7 @@ public boolean cancelOrder(CancelOrderParams params) throws IOException {
     }
   }
 
-  @Override
-  public UserTrades getTradeHistory(TradeHistoryParams params) throws IOException {
+  private List<BinanceTrade> getBinanceTrades(TradeHistoryParams params) throws IOException {
```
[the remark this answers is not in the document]
amjad1210: This is completely my fault. Looking further into getTradeHistory(..) I see the fee is available in UserTrade.

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
amjad1210: Added a solution to this and pushed the changes.

xchange-core/src/main/java/org/knowm/xchange/service/trade/TradeService.java
```diff
@@ -321,4 +337,37 @@ default void verifyOrder(MarketOrder marketOrder) {
...
+   * @return the order as it is on the exchange.
+   * @throws ExchangeException - Indication that the exchange reported some kind of error with the
+   *     request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the
+   *     requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the
+   *     requested function or data, but it has not yet been implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  default Collection<Order> assureTradeResultsFilled(Collection<Order> orders) throws IOException {
```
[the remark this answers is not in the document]
amjad1210: Only issue with this is i'm calling getOrderWithTradeResults(..) by default for multiple exchanges instead of getOrders(...) and need to pass through the orders even if nothing is done on getOrderWithTradeResults(...) for that exchange.

Issue #3171, opened 2019-08-13
Support for Futures markets


andre77:
Currently the central key point of the XChange API is the CurrencyPair, which is passed as parameter for almost all methods. This works fine so far for spot markets. But if we want to support futures markets the CurrencyPair is not enough, for example see [Deribit Instruments](https://www.deribit.com/api/v2/public/get_instruments?currency=BTC&expired=false&kind=future), there we have three "products" with the same pair BTC/USD.
To support this we would need to replace the CurrencyPair by `Product` class, which could look like this:

```
public class Product {
  private String base;
  private String counter;
  private String symbol;	// this can be null, ie for spot products
  private ProductType productType;
}

public enum ProductType {
  spot, perpetual, future;
}
```

To stay with the deribit example, we would have three instances for products:
```
{base: BTC, counter: USD, symbol: BTC-27SEP19, productType: future}
{base: BTC, counter: USD, symbol: BTC-27DEC19, productType: future}
{base: BTC, counter: USD, symbol: BTC-PERPETUAL, productType: perpetual}
```

So like already said, inside the interface methods we could replace the CurrencyPair by Product (or Instrument or what ever name). The old methods could actually stay for a while and be marked as deprecated. This way we would be still compatible with older versions.

if the majority thinks it could make sense, i will create a PR

makarid:
I like the idea. Maybe we can add options support also. They just need expiration date field i believe. What do you thing?

andre77:
futures also have expiration date, i do not mind to add an optional Date field, which would be null for 'spot' and 'perpetual' products

makarid:
Great.

timmolter:
sounds good to me.

makarid:
Have anyone implement this or at least started?

andre77:
not yet unfortunately

douggie:
@andre77 wondered if you had any thoughts on the product PR idea, just I was going to look at implementing TFX which would need a simliar solution, I was going to approach it a bit different but may be your way is better. 

https://github.com/knowm/XChange/issues/3258

andre77:
no, unfortunately did not find any time to make some progress here

mdvx:
>> Maybe we can add options support also. They just need expiration date field and strike price i believe.We can create a second constructor for option support. 

[...]

PR #3243, opened 2019-09-24
[deribit] additional methods ie positions etc.

andre77:


walec51:
I see DeribitTradeService has only cancel implemented from the generic methods
Do you plan to add more in this PR or should we merge it as is for now?

PS. If you will be working further on Deribit it would be nice to add better exception handling in non-*Raw methods in accordance to: https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-7

andre77:
as for now, i do not plan to make further changes here. deribit in an exchange for futures, where the simple handling with CurrencyPair's is not possible. see https://github.com/knowm/XChange/issues/3171
so currently you can use the *Raw methods, like i do

PR #3245, opened 2019-09-24
[bitfinex] further implementation of finex api v2

andre77:
added cnht support


xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/BitfinexExceptionV2.java
```diff
@@ -0,0 +1,24 @@
...
+@ToString
+public class BitfinexExceptionV2 extends RuntimeException {
+
+  private String type;
+  private Long errorCode;
+  private String message;
+
+  // ["error",10100,"apikey: invalid"]
+  @JsonCreator
+  public BitfinexExceptionV2(Object[] arr) {
```
walec51: can jackson handle `Object[]`? I have my doubts, shouldn't this be `String[]`?
could you provide a test using wiremock which shows a BitfinexExceptionV2 escaping a *Raw method?

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/BitfinexAuthenticated.java
```diff
@@ -1,11 +1,64 @@
...
+  public static final String BFX_NONCE = "bfx-nonce";
+
+  /** https://docs.bitfinex.com/v2/reference#rest-auth-positions */
+  @POST
+  @Path("auth/r/positions")
+  List<Position> activePositions(
+      @HeaderParam(BFX_NONCE) SynchronizedValueFactory<Long> nonce,
+      @HeaderParam(BFX_APIKEY) String apiKey,
+      @HeaderParam(BFX_SIGNATURE) ParamsDigest signature,
+      EmptyRequest empty)
```
walec51: is this needed?
wouldn't this work without this `empty` argument?

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/service/BitfinexMarketDataServiceRaw.java
```diff
@@ -132,4 +133,13 @@ public BitfinexLendDepth getBitfinexLendBook(String currency, int limitBids, int
       throw new BitfinexException(e.getHttpBody());
     }
   }
+
+  public List<Status> getStatus(List<CurrencyPair> pairs) throws IOException {
+    try {
+      return bitfinexV2.getStatus(
+          "deriv", BitfinexAdapters.adaptCurrencyPairsToTickersParam(pairs));
+    } catch (HttpStatusIOException e) {
```
walec51: if you have a proper BitfinexExceptionV2 then this should not get thrown
please see my comment on BitfinexExceptionV2

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/BitfinexHmacSignature.java
```diff
@@ -0,0 +1,58 @@
...
+  @Override
+  public String digestParams(RestInvocation i) {
+
+    // /api/${apiPath}${nonce}${JSON.stringify(body)}
+
+    String path = i.getPath();
+    try {
+      path = URLDecoder.decode(path, "utf8");
+    } catch (UnsupportedEncodingException ignored) {
+    }
```
walec51: silent errors are evil, even the "its almost impossible" ones ;)
just surround the entire methods body with try - catch
you have a good catch that handles it in the end

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/BitfinexExceptionV2.java
andre77: sure it can handle `Object[]`, see the comment over the method, this is, how an error could like passed by bitfinex `["error",10100,"apikey: invalid"]`

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/BitfinexAuthenticated.java
andre77: yes, the body must be a valid json structure, so you need to pass the `EmptyRequest` here to get curly brackets `{}`

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/BitfinexHmacSignature.java
andre77: i do not see a reason to change this, sorry
basically, we try to url decode the "path", if it fails, so what, we simply use the "path" like it is
the worst thing in case of wrong "path" value would be, that we end up having an invalid signature, which would lead to a failed request on the server with an according error. but this scenario does not happen in all my tests. so what can be improved here ?

walec51: ok, then at least log something in this catch

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/BitfinexAuthenticated.java
walec51: ok

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/BitfinexExceptionV2.java
walec51: then why does this happen in your code:

```
  public List<Status> getStatus(List<CurrencyPair> pairs) throws IOException {
    try {
      return bitfinexV2.getStatus(
          "deriv", BitfinexAdapters.adaptCurrencyPairsToTickersParam(pairs));
    } catch (HttpStatusIOException e) {
```

if BitfinexExceptionV2 works correctly then it should be thrown by `bitfinexV2.getStatus` instead of HttpStatusIOException

we can argue or we can implement a test that proves what happens ;)
you can find an example on how to write a test with wiremock here: https://github.com/knowm/XChange/blob/develop/xchange-livecoin/src/test/java/org/knowm/xchange/livecoin/service/TradeMockedIntegrationTest.java

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/EmptyRequest.java
```diff
@@ -0,0 +1,8 @@
+package org.knowm.xchange.bitfinex.v2.dto;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.NONE)
+public class EmptyRequest {
+  public static final EmptyRequest instance = new EmptyRequest();
```
walec51: INSTANCE <- uppercase

andre77:
@walec51 feel free to close the PR, i'm fine with that

PR #3260, opened 2019-10-07
[Core,Binance] Renamed destinationTag to addressTag

amjad1210:
[Core]
- Renamed destinationTag to addressTag

[Binance]
- Added addressTag to withdraw and deposit history.

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/DepositList.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/DepositList.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.List;
+import lombok.Data;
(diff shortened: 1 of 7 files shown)


walec51:
The destination tag field is used mostly for moneros payment id, stellars memo id, ripples destination tag, etc. I don't see how changing destinationTag to addressTag in the Core API improves anything. It does however break backwards comparability.

amjad1210:
> The destination tag field is used mostly for moneros payment id, stellars memo id, ripples destination tag, etc. I don't see how changing destinationTag to addressTag in the Core API improves anything. It does however break backwards comparability.

@walec51 When I initially implemented the destinationTag field my focus was around XRP which is where the term came from. The reason for this change is destinationTag is a term solely used for Ripple whereas Monero uses Payment ID and EOS uses Memo ID etc... I wanted to rename the field to be more generic and addressTag seems to be the correct choice.

If you still disagree I can revert the changes for destinationTag to addressTag.

walec51:
Well ok, the name seam a little better. I guess this particular DTO broke backwards compatibility multiple times in the past so it won't hurt that much.

PR #3270, opened 2019-10-18
Add and test canonical JSON serialization in lots of places

badgerwithagun:
Slightly opinionated approach.

I am intending to introduce a lot of IPC using these objects so need reliable, clear JSON serialisation.


xchange-core/src/main/java/org/knowm/xchange/currency/Currency.java
```diff
@@ -302,8 +305,9 @@
   public static final Currency STORJ = createCurrency("STORJ", "Storj", null);
   public static final Currency MOD = createCurrency("MOD", "Modum", null);
 
-  private final String code;
-  private final CurrencyAttributes attributes;
+  @JsonProperty private final String code;
```
walec51: i think this class should serialize to a simple string, not an object, same goes for `CurrencyPair`

when you serialize an order you want it to look like this:

```
{
  "type" : "BID",
  "originalAmount" : "0.234234",
  "currencyPair" : "ETH/BTC",
  // ...
}
```
**NOT like this:**

```
{
  "type" : "BID",
  "originalAmount" : "0.234234",
  "currencyPair" : {
    "base" : {
      "code" : "ETH"
      "attributes" : [
        // tons of metadata that repeat in every file
      ]
    },
    "counter" : {
      "code" : "BTC"
      "attributes" : [
        // tons of metadata that repeat in every file
      ]
    }
  },
  // ...
}
```

I think that just using `@JsonCreator` on `getInstance` and `@JsonValue` on `getCurrencyCode` should do the trick

xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
```diff
@@ -185,6 +197,7 @@ public Date getTimestamp() {
     return timestamp;
   }
 
+  @JsonProperty("flags")
```
walec51: property names in JSON should be the same as in the classes

xchange-core/src/test/java/org/knowm/xchange/dto/trade/StopOrderTest.java
```diff
@@ -85,6 +88,10 @@ public void testBuilderFrom() {
     final StopOrder copy = StopOrder.Builder.from(original).build();
 
     assertThat(copy).isEqualToComparingFieldByField(original);
+
+    StopOrder jsonCopy = ObjectMapperHelper.viaJSON(copy);
+    assertThat(jsonCopy).isEqualToIgnoringGivenFields(copy, "cumulativeAmount");
+    assertTrue(jsonCopy.getCumulativeAmount().compareTo(copy.getCumulativeAmount()) == 0);
```
walec51: this mixes to units into one test

please separate and other tests like this

just build your own order and pass it viaJSON

xchange-core/src/main/java/org/knowm/xchange/currency/Currency.java
badgerwithagun: Done

xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
badgerwithagun: OK, for consistency I have renamed the field to match the property :)

badgerwithagun: I've also had to add a `JsonProperty("orderFlags")` on the builder.

Basically I'm aware of the convention but:

- Builder was using `flags`
- Field was called `flags`
- Getter and setter were called `getOrderFlags/setOrderFlags`

So I went with the most common one!

I've reversed this to standardise them all on `orderFlags`.

badgerwithagun:
All done, @walec51

xchange-core/src/main/java/org/knowm/xchange/currency/CurrencyPair.java
```diff
@@ -405,6 +407,7 @@ public CurrencyPair(String baseSymbol, String counterSymbol) {
   /**
    * Parse currency pair from a string in the same format as returned by toString() method - ABC/XYZ
    */
+  @JsonCreator
```
walec51: I think we should also add `@JsonValue` to the `toString` method in this class

xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
```diff
@@ -408,12 +421,14 @@ protected Builder(OrderType orderType, CurrencyPair currencyPair) {
       this.currencyPair = currencyPair;
     }
 
+    @JsonProperty("type")
```
walec51: ok, now I understand why you say that the naming convention in this file is a mess :)
for now the getters/setters of the main class should be the main determinatior of a name of a given property
in a 5.X.X release I think we should ditch the word Order in any property here as it is obvious that it relates to the order

xchange-core/src/test/java/org/knowm/xchange/dto/trade/UserTradeTest.java
```diff
@@ -83,6 +85,9 @@ public void testBuilderFrom() {
     assertThat(copy.getOrderId()).isEqualTo(original.getOrderId());
     assertThat(copy.getFeeAmount()).isEqualTo(original.getFeeAmount());
     assertThat(copy.getFeeCurrency()).isEqualTo(original.getFeeCurrency());
+
+    UserTrade jsonCopy = ObjectMapperHelper.viaJSON(copy);
+    assertThat(jsonCopy).isEqualToComparingFieldByField(copy);
```
walec51: move to separate test

xchange-core/src/test/java/org/knowm/xchange/dto/trade/MarketOrderTest.java
```diff
@@ -75,6 +77,9 @@ public void testBuilderFrom() {
     final MarketOrder copy = MarketOrder.Builder.from(original).build();
 
     assertThat(copy).isEqualToComparingFieldByField(original);
+
+    MarketOrder jsonCopy = ObjectMapperHelper.viaJSON(copy);
+    assertThat(jsonCopy).isEqualToComparingFieldByField(copy);
```
walec51: move to separate test

xchange-core/src/test/java/org/knowm/xchange/dto/trade/LimitOrderTest.java
```diff
@@ -82,6 +85,10 @@ public void testBuilderFrom() {
     final LimitOrder copy = LimitOrder.Builder.from(original).build();
 
     assertThat(copy).isEqualToComparingFieldByField(original);
+
+    LimitOrder jsonCopy = ObjectMapperHelper.viaJSON(copy);
+    assertThat(jsonCopy).isEqualToIgnoringGivenFields(copy, "cumulativeAmount");
+    assertTrue(jsonCopy.getCumulativeAmount().compareTo(copy.getCumulativeAmount()) == 0);
```
walec51: move to separate test

xchange-core/src/test/java/org/knowm/xchange/dto/trade/UserTradeTest.java
```diff
@@ -50,7 +52,7 @@ public void testBuilder() {
   }
```
walec51: please add one test that will allow us to check if CurrencyPair is serialized as a String:

```

  @Test
  public void testSerialize() throws IOException {
    UserTrade original = ...;
    String serialized = ObjectMapperHelper.toCompactJSON(original);
    assertThat(serialized).contains("\"currencyPair\":\"BTC/USD\"");
  }
```

xchange-core/src/test/java/org/knowm/xchange/dto/account/BalanceTest.java
```diff
@@ -0,0 +1,28 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.IOException;
+import java.math.BigDecimal;
+import org.junit.Test;
+import org.knowm.xchange.currency.Currency;
+import org.knowm.xchange.utils.ObjectMapperHelper;
+
+public class BalanceTest {
+
```
walec51: please add one test that will allow us to check if Currency is serialized as a String:

```

  @Test
  public void testSerialize() throws IOException {
    Balance balanse = ...;
    String serialized = ObjectMapperHelper.toCompactJSON(balanse);
    assertThat(balanse).contains("\"currency\":\"ADA\"");
  }
```

xchange-core/src/main/java/org/knowm/xchange/currency/CurrencyPair.java
badgerwithagun: Removed the custom converter, since it's no longer necessary

xchange-core/src/test/java/org/knowm/xchange/dto/trade/UserTradeTest.java
badgerwithagun: This is a valid extension to the same test. Done

badgerwithagun:
Review fixes in.

walec51:
thanks, merged

PR #3273, opened 2019-10-23
Hitbtc

roeiba:
I am suggesting here my solution 

I defined user reference as part of LimitOrder (builder updated accordingly). In any case order user reference is not provided, it is generated by the Builder.

I also fixed Hitbtc to use this user reference properly. After the fix, HitbtcLimitOrder class becomes redundant and can deleted, but I fixed as well just for backward compatibility until it will be deprecated.

I also improved Hitbtc getFundingHistory to support filters.

This solves both #2774 and #2779


walec51:
I'm a little worried by the fact that most of the exchanges will just ignore this field.
However I understand that there is no way around this to make HitBTCs integration work and other ways client id ware handled are even worse like: putting the clientId to order flags or doing some adapting DTO hibrid like `HitbtcLimitOrder extends LimitOrder`.

I no one objects to adding a `Order#userReference` property that will only by supported on few exchanges then I'll merge this in a few days.

PR #3281, opened 2019-11-07
Make testing for JSON serialization/deserialization stricter

badgerwithagun:
Previously it allowed deserialization not to work when new properties were added.
Fixed a number of broken properties.


badgerwithagun:
Build failed for entirely unrelated reasons. Closing to kick it off again

PR #3282, opened 2019-11-08
Make testing for JSON serialization/deserialization stricter

badgerwithagun:
Previously it allowed deserialization not to work when new properties were added.
Fixed a number of broken properties.

### DIFF
--- a/xchange-core/src/test/java/org/knowm/xchange/dto/account/BalanceTest.java
+++ b/xchange-core/src/test/java/org/knowm/xchange/dto/account/BalanceTest.java
@@ -2,16 +2,21 @@ package org.knowm.xchange.dto.account;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
 import java.math.BigDecimal;
 import org.junit.Test;
 import org.knowm.xchange.currency.Currency;
-import org.knowm.xchange.utils.ObjectMapperHelper;
 
 public class BalanceTest {
 
   @Test
   public void testSerializeDeserialize() throws IOException {
+
+    // This deliberately doesn't use FAIL_ON_UNKNOWN_PROPERTIES so that
+    // we ensure serialization/deserialization is symmetrical.
+    ObjectMapper objectMapper = new ObjectMapper();
+
     Balance balance =
         new Balance.Builder()
             .available(new BigDecimal("0.12"))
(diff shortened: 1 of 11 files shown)


badgerwithagun:
Test failures are in `LatokenMarketDataServiceTest` and are nothing to do with this change. Reverted and they still happen.

xchange-core/src/test/java/org/knowm/xchange/dto/account/BalanceTest.java
```diff
@@ -2,16 +2,21 @@
...
-import org.knowm.xchange.utils.ObjectMapperHelper;
 
 public class BalanceTest {
 
   @Test
   public void testSerializeDeserialize() throws IOException {
+
+    // This deliberately doesn't use FAIL_ON_UNKNOWN_PROPERTIES so that
+    // we ensure serialization/deserialization is symmetrical.
+    ObjectMapper objectMapper = new ObjectMapper();
```
walec51: instead of creating a new `ObjectMapper` for each test can't we just use `ObjectMapperHelper` and the `readValueStrict` method added by you?

badgerwithagun: That's done, @walec51. Back with you.

xchange-core/src/test/java/org/knowm/xchange/dto/trade/LimitOrderTest.java
```diff
@@ -92,6 +93,11 @@ public void testBuilderFrom() throws IOException {
 
   @Test
   public void testSerializeDeserialize() throws IOException {
+
+    // This deliberately doesn't use FAIL_ON_UNKNOWN_PROPERTIES so that
+    // we ensure serialization/deserialization is symmetrical.
+    ObjectMapper objectMapper = new ObjectMapper();
```
walec51: this is not used anywhere in the method

xchange-core/src/test/java/org/knowm/xchange/dto/trade/UserTradeTest.java
```diff
@@ -110,10 +115,10 @@ public void testSerializeDeserialize() throws IOException {
             feeAmount,
             feeCurrency);
 
-    String json = ObjectMapperHelper.toCompactJSON(original);
+    String json = objectMapper.writeValueAsString(original);
```
walec51: why not just use ObjectMapperHelper.viaJSON ?

badgerwithagun: Done

xchange-core/src/test/java/org/knowm/xchange/dto/trade/LimitOrderTest.java
badgerwithagun: Fixed

PR #3283, opened 2019-11-11
[coinbasepro] Added support for "limit" parameter in trade history request

Yampersand:
Added support for request parameter "limit" in trade history request to better support paging.

Also "fixed" a test which makes a live call and the response seems to have changed.


Yampersand:
I know there are tests failing, but I don#t think I'm responsible. They fail in project 'latoken' which I haven't touched ...

PR #3292, opened 2019-11-15
Skip PoloniexLevel when adapting order book

mdvx:
and save big on memory consuption when running with XStream
NB: also remove PoloniexLevel class


walec51:
ok, ad the commit that removes `PoloniexLevel`, its not used any ware else

PR #3300, opened 2019-11-19
[Binance] Stop order automatically selects the wrong type

SergeyLukashevich:
Add requirement to manually set stop order type.

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
@@ -1,8 +1,9 @@
 package org.knowm.xchange.binance.dto.trade;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import org.knowm.xchange.dto.Order.IOrderFlags;
 
-public enum OrderType {
+public enum OrderType implements IOrderFlags {
   LIMIT,
   MARKET,
   TAKE_PROFIT_LIMIT,
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
@@ -1,9 +1,8 @@
 package org.knowm.xchange.binance.dto.trade;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
-import org.knowm.xchange.dto.Order.IOrderFlags;
 
-public enum OrderType implements IOrderFlags {
+public enum OrderType {
   LIMIT,
   MARKET,
   TAKE_PROFIT_LIMIT,
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAdapters.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAdapters.java
@@ -188,4 +188,24 @@ public class BinanceAdapters {
     }
     return new CurrencyMetaData(precision, withdrawalFee, minWithdrawalAmount);
   }
+
+  public static org.knowm.xchange.binance.dto.trade.OrderType adaptOrderType(StopOrder order) {
+
+    if (order.getIntention() == null) {
+      throw new IllegalArgumentException("Missing intention");
+    }
+
+    switch (order.getIntention()) {
+      case STOP_LOSS:
+        return order.getLimitPrice() == null
+            ? org.knowm.xchange.binance.dto.trade.OrderType.STOP_LOSS
+            : org.knowm.xchange.binance.dto.trade.OrderType.STOP_LOSS_LIMIT;
+      case TAKE_PROFIT:
+        return order.getLimitPrice() == null
+            ? org.knowm.xchange.binance.dto.trade.OrderType.TAKE_PROFIT
+            : org.knowm.xchange.binance.dto.trade.OrderType.TAKE_PROFIT_LIMIT;
+      default:
+        throw new IllegalStateException("Unexpected value: " + order.getIntention());
+    }
+  }
 }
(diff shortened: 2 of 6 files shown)


walec51:
your change will break currently working code

whats wrong with the current mapping?

SergeyLukashevich:
Because order type calculation is wrong:

```    OrderType orderType;
    if (so.getType().equals(Order.OrderType.BID)) {
      orderType = so.getLimitPrice() == null ? OrderType.TAKE_PROFIT : OrderType.TAKE_PROFIT_LIMIT;
    } else {
      orderType = so.getLimitPrice() == null ? OrderType.STOP_LOSS : OrderType.STOP_LOSS_LIMIT;
    }
```

The assumption that buy limit order always is take-profit limit order - is not true. The same is for sell. 
If some code relies on it, it can be fixed by explicitly setting the needed type.

walec51:
Ok, your right, the current implementation is just wrong.

But the way you are trying to fix it is also very hacky, you allow to much order flags and someone will by able to do a StopOrder with a MARKET + LIMIT flag monstrosity. Besides that, having a generic API that requires an exchange specific flag to work ruins the point of having a generic API. 99% of developer when they execute a 'placeStopOrder' expect a stop loss order to get placed - NOT a IllegalArgumentException or a Take Profit order.

The root of the current problem is laziness or lack of understanding that a take profit order is not a special type of stop loss order - its something opposite. Lets not continue this mistake.

Please add a new type of order to xchange-core - `TakeProfitOrder` - it will be very similar to `StopOrder` but do something opposite - this should be reflect in its javadocs. Afterwards add a

```
String placeTakeProfitOrder(TakeProfitOrder takeProfitOrder) throws IOException
```

method to `TradeService` and implement this feature in Binance in a proper manner by separating stop orders from take profit orders.

SergeyLukashevich:
> you allow to much order flags and someone will by able to do a StopOrder with a MARKET + LIMIT flag monstrosity

Agree, the combination of MARKET and LIMIT order doesn't make sense. Probably it causes exchange return an error, but anyway we have to reject such combination. Moreover we don't them at all, as soon
Binance expect only STOP_LOSS* and TAKE_PROFIT* flags for a stop order.

> 99% of developer when they execute a 'placeStopOrder' expect a stop loss order to get placed - NOT a IllegalArgumentException or a Take Profit order.

Stop order is not a synonym of stop-loss order. Stop-loss and take-profit orders are just some orders that are used to close a position. We can use a stop order to close position both by stop loss and take profit condition. It's may vary between different trading platforms, but at least it's valid in pure theory. The fact that Binance links stop order with exit condition, it's their internal interpretation of how stop order should look like, and I purpose to do not use this approach in a common interface.

How about these changes in PR:
1. Flags shouldn't have any of: MARKET or LIMIT
2. Flags should include one of: STOP_LOSS, STOP_LOSS_LIMIT, TAKE_PROFIT, and TAKE_PROFIT_LIMIT
3. On invalid flags combination - error should be raised and demand to set valid flag explicitly with a short explanation, why.

walec51:
Flag validation in your proposition is just a way of managing a hacky solution. Our core API where possible should not allow to construct orders that do not make sense nor should it require some exchange-specific flags to work in any capacity. Adding validation does not fix bad API design.

I see your point on how you view **Stop Order** as a generic term that applies to both **Stop loss order** and **Take profit order**. Binances API suggests that. However on Bitmex this is a different order type: https://www.bitmex.com/api/explorer/#!/Order/Order_new

So I went to more general literature: https://www.investopedia.com/terms/s/stop-limitorder.asp
I see that a stop order has a simpel definition and the fact is it a 'stop loss' or 'take profit' is determined by the fact is your stop price higher or lower from the current price. However in the world of wildly changing prices of crypto I see that exchanges prefer that you state your intent explicitly.

Taking that into considerations my earlier proposal was also bad.

My new proposal would be to just add a new optional enum field to StopOrder:

```
public StopOrder extends Order {
    // ...
   public Intention intention;
   
   public static enum Intention {
     STOP_LOSS, TAKE_PROFIT;
   }
   // ...
}
```

The javadoc of this field should state that on some exchanges this field is required and if you want your code to work across many cross exchange then you should always set it.

What do you think of this approach?

SergeyLukashevich:
Yes, it sounds reasonable.

So, using new enum for Binance, its order type will be made in that way:
* STOP_LOSS: OrderType.MARKET + Intention.STOP_LOSS
* STOP_LOSS_LIMIT: OrderType.LIMIT + Intention.STOP_LOSS
* TAKE_PROFIT: OrderType.MARKET + Intention.TAKE_PROFIT
* TAKE_PROFIT_LIMIT: OrderType.LIMIT + Intention.TAKE_PROFIT

walec51:
Not exactly. In the solution I'm proposing you do not have to set any OrderFlags at all in `StopOrder`. The adaptation to a Binance order DTO would look like this:

```
switch (order.getIntention()) {
  case Intention.STOP_LOSS: return order.getLimitPrice() == null ? STOP_LOSS : STOP_LOSS_LIMIT;
  case Intention.TAKE_PROFIT: return order.getLimitPrice() == null ? TAKE_PROFIT : TAKE_PROFIT_LIMIT;
```

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/OrderType.java
```diff
@@ -1,8 +1,9 @@
 package org.knowm.xchange.binance.dto.trade;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import org.knowm.xchange.dto.Order.IOrderFlags;
 
-public enum OrderType {
+public enum OrderType implements IOrderFlags {
```
walec51: everything is ok, just remove this `implements IOrderFlags` change

PR #3304, opened 2019-11-21
[Poloniex] Memory Optimization #2

mdvx:
Save 1 LimitOrder.Builder (which also contains 1 HashMap for OrderFlags) per row of Market depth

### DIFF
--- a/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAdapters.java
+++ b/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAdapters.java
@@ -92,10 +92,17 @@ public class PoloniexAdapters {
 
     for (List<BigDecimal> rawlevel : rawLevels) {
       LimitOrder limitOrder =
-          new LimitOrder.Builder(orderType, currencyPair)
-              .originalAmount(rawlevel.get(1))
-              .limitPrice(rawlevel.get(0))
-              .build();
+          new LimitOrder(
+              orderType,
+              rawlevel.get(1),  // originalAmount
+              currencyPair,
+              null,
+              null,   
+              rawlevel.get(0),  // limitPrice
+              null,
+              null,
+              null,
+              Order.OrderStatus.UNKNOWN);
       orders.add(limitOrder);
     }
     return orders;


timmolter:
interesting. How much of a difference does that make?

mdvx:
The volume of data generated by Poloniex, when subscribed to a streaming order book, sends the Garbagge collector into a craze, after a few minutes running.  These optimizations increase the time the VM runs smoothly by a factor of 4.  YourKit profiler identifies 99% of objects being collected, as those created here.

mdvx:
After a few seconds.  Note the 71% is also HashMap and ArrayLists generated by orderbook

image

walec51:
I'm against this change.

Constructors with dozens of arguments should be deprecated in time in favour of builders which provide much better code readability.

If we want to optimize something then lets optimize how those builders work. For example by not allocating a HashMap or ArrayList if its not needed and building an order with Collections.EMPTY_MAP / EMPTY_LIST if it was not allocated. The builder it self is a temporary object which get easily cleaned in the nearest GC run.

mdvx:
I agree with the argument, for normal objects.  What I am seeing is "OutOfMemoryError: GC overhead limit exceeded" after about 5-15 minutes of subscribing just to BTC/USD on Poloniex (I have five other exchanges connected, but this one breaks my JVM).  Looking into how Poloniex was implemented, the OrderBook is full recreated with every update message, so for a depth of 100, thats (updates/second) * (depth) 100 * (bid/ask) 2 * (objects/row) 6 => 1200 * updates/second => (if we assume 2) 2,400 objects for GC collect per second.

The Poloniex orderbook handling should probably be reimplemented using a mutable orderbook, and that would solve the problem, but thats an awful lot more work.

Would you be ameable to adding a 5 parameter constuctor to LimitOrder(symbol,side,qty,price,timestamp) instead?  My preference is to an interfaces ILimitOrder, that does not unpack and repack the fields, but hold a reference to the raw data plus the getters.

walec51:
I don't think that allocating 2400 small objects should be a problem nor should it overwhelm your GC unless your running on some embedded device.

What might be escalating the problem is the `orderFlags` `HashSet` which is created for every `Order` and `Order.Builder` instance. Its default initial capacity is 16. Using `Collections.EMPTY_SET` by default and adding some code to the builders `flags` and `flag` methods so that they create an actual non empty HashSet at first usage should greatly improve the situation.

As for making DTOs mutable or changing them to interfaces with implementations working as adapters - this would be a radical change for this library and I don't think it is justified before we try simpler optimizations.

mdvx:
Ok, i can implement in a higher layer.

mdvx:
@walec51 I did not see you comment when I closed this, using Collections.EMPTY_SET should be a good work around for the builder.  

I have been running the modified code for a few days, and it really did fix my GCs (I have a fast dev machine).  But I understand the reluctance, and since I work mostly in the with XStream (and alike) libraries, we have to deal with many more updates per second than XChange.

Also, I have had a lot of problems with the Order object as implemented, especially since properties like limit,stop and orderflags are not friendly.  (I just had one today with Kraken, trying to detect the fee currency).

Still, having XChange as a base library to start from is great.  (any plans to add CurrencyPair.getInstance()? like Currency.getInstance*( has - another one that creates thousands/millions of objects - but i can add my own cache for that).

PR #3309, opened 2019-11-26
Dead man's switch (auto cancel)

andre77:
implemented for
 - bitmex https://www.bitmex.com/app/wsAPI#Dead-Mans-Switch-Auto-Cancel
 - crypto facilities https://www.cryptofacilities.com/resources/hc/en-us/articles/360005994214-Cancel-All-Orders-After

### DIFF
--- a/xchange-bitmex/pom.xml
+++ b/xchange-bitmex/pom.xml
@@ -35,6 +35,9 @@
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 		</dependency>
-
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+        </dependency>
 	</dependencies>
 </project>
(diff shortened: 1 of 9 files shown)


walec51:
great PR, just one question: can't we use \@Value instead of \@Data for DTOs?
or does jackson have a problem with the constructor added by \@Value ?

andre77:
@walec51 good point, i would actually prefer to use immutable classes with `@Value` annotation
but indeed it does not work out of the box with jackson, since there is no default constructor
i get such an exception:
> com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.knowm.xchange.bitmex.dto.trade.BitmexCancelAll` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
afaik jackson needs a default constructor with setter methods
may be there is a way to solve it however, let me know :-)

walec51:
Ok, thanks for checking out. I'm pretty sure I was able to configure Jackson to work with constructors generated by `@Value` in another project. I'll experiment in another PR.

PR #3313, opened 2019-12-04
[ItBit] Fixing inconsistency of getTradeHistory method. Itbit returns…

makarid:
… orders with nested trades and we should separate these trades in indepentent orders, because you would get different results which depended on the time of a trade will get accepted.

### DIFF
--- a/xchange-itbit/src/main/java/org/knowm/xchange/itbit/ItBitAdapters.java
+++ b/xchange-itbit/src/main/java/org/knowm/xchange/itbit/ItBitAdapters.java
@@ -217,47 +217,44 @@ public final class ItBitAdapters {
     List<UserTrade> trades = new ArrayList<>();
 
     for (String orderId : tradesByOrderId.keySet()) {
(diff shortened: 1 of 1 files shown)


walec51:
I agree that setting tradeId = orderId is bad.

But your solution seams also bad because the way you set tradeId does not guarantee that the same trades will have the same ids in different requests. If an order has trades that happened between 12:00 - 17:00 and I filter the trade history using startTime > 15:00 then the first trade that happened after 15:00 will have tradeId = orderId-1 despite the fact that they could have been trades before it.

If the exchange does not have trade ids then they should just be set null.

walec51:
ps. in their API I see that each trade has an `executionId`
https://api.itbit.com/docs#trading-get-trades-get
isn't this a de facto trade id?

makarid:
Thank you for your response, @walec51 . I will definitely see that option. I have contact with the support and they told me that they know this design issue and they have told it to their developers. So, i am not sure if the executionId is the correct one, because if it was they would told me to use that one. But i will check it.Thanks

PR #3314, opened 2019-12-04
Fix string to currency pair conversion when base name starts with "t"

davidjirovec:
Was not working correctly for example for TKN/USD


walec51:
please add unit test for this case

davidjirovec:
Done

PR #3315, opened 2019-12-05
[Bitstamp] Invalid nonce errors

mdvx:
Keeping up with the nanotimes


walec51:
this will exceed the maximum nonce value in the year 2262...

... ok, approved ;)

Issue #3262, opened 2019-10-14
Bitfinex not initializing because of funding currencies.


bxabi:
https://docs.bitfinex.com/reference#rest-public-tickers

Apparently bitfinex added funding currencies, and the result array is different from the regular ones, so the Json parsing fails.
There are 7 more values returned, and even the order is different :(

Maybe I can work on this with your help (I made a workaround for me just adding 7 fields in the BitfinexTicker.

PR #3322, opened 2019-12-15
Update bitfinex ticker fields

TonnTamerlan:
Bitfinex tickers can have two different types
[documentation](https://docs.bitfinex.com/reference#rest-public-ticker)


TonnTamerlan:
https://github.com/knowm/XChange/issues/3262

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTicker.java
```diff
@@ -2,104 +2,50 @@
...
-
-  public BigDecimal getLow() {
-    return low;
-  }
-
-  public BigDecimal getAsk() {
-    return ask;
-  }
+  private Object placeHolder0;
+  private Object placeHolder1;
```
walec51: do you have some example values this placeholder may have?
if so then please add then to your test

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTicker.java
```diff
@@ -2,104 +2,50 @@
...
+  private BigDecimal frrAmountAvailable;
 
   @Override
   public String toString() {
     return "BitfinexTicker{"
         + "symbol='"
         + symbol
         + '\''
+        + ", frr="
+        + frr
```
walec51: if we're going with lombok we can just use `@ToString` and remove this method

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTicker.java
TonnTamerlan: these placeholders are reserved for the future and they are always null

TonnTamerlan: @walec51 these placeholders are reserved for the future and they are always null

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTicker.java
TonnTamerlan: @walec51 fixed

TonnTamerlan:
@walec51 Could you please merge this pull request

PR #3332, opened 2019-12-18
[CoinbasePro] add support for CoinbasePrime part#2

mdvx:
CoinbasePrime does not return MetaData in the same way


mdvx:
deprends on #3331

PR #3333, opened 2019-12-18
[Binance, CoinbasePro] Add canDeposit + canWthdraw in CoinMetaData

amjad1210:
[Binance]
- Remove hard coded currencies when loading dynamically.
- Add canDeposit + canWithdraw for coins.

[CoinbasePro]
- Add canDeposit + canWithdraw for coins.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/meta/CurrencyMetaData.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/meta/CurrencyMetaData.java
@@ -19,13 +19,9 @@ public class CurrencyMetaData implements Serializable {
   @JsonProperty("min_withdrawal_amount")
   private final BigDecimal minWithdrawalAmount;
 
-  /** Deposit enabled */
-  @JsonProperty("deposit_enabled")
-  private boolean depositEnabled;
-
-  /** Withdraw enabled */
-  @JsonProperty("withdraw_enabled")
-  private boolean withdrawEnabled;
+  /** Wallet health */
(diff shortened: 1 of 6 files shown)


xchange-core/src/main/java/org/knowm/xchange/dto/meta/CurrencyMetaData.java
```diff
@@ -19,15 +19,35 @@
   @JsonProperty("min_withdrawal_amount")
   private final BigDecimal minWithdrawalAmount;
 
+  /** Deposit enabled */
+  @JsonProperty("deposit_enabled")
+  private boolean depositEnabled;
+
+  /** Withdraw enabled */
+  @JsonProperty("withdraw_enabled")
+  private boolean withdrawEnabled;
```
walec51: I don't like the fact that most exchanges will return false as most APIs do not share this information

maybe introducing an enum would be better?

```
enum WalletsHealth {
    /** You can deposit and withdraw founds from the exchange */
    ONLINE,
    DEPOSITS_DISABLED,
    WITHDRAWALS_DISABLED,
    /** You cannot deposit nor withdraw founds from the exchange */
    OFFLINE,
    /** The exchange does not inform us about the health of this wallet */
    UNKNOWN;
}
```

amjad1210: @walec51 Makes perfect sense i'll make the changes.

amjad1210:
@walec51 Please review changes.

PR #3334, opened 2019-12-19
[LGO] Bug fix on price increment check + small enhancements.

BodySplash:


xchange-lgo/src/main/java/org/knowm/xchange/lgo/dto/product/LgoProductCurrency.java
```diff
@@ -14,7 +15,7 @@ public LgoProductCurrency(
       @JsonProperty("increment") BigDecimal increment,
       @JsonProperty("limits") LgoLimit limits) {
     this.id = id;
-    this.increment = increment;
+    this.increment = increment != null ? increment.stripTrailingZeros() : null;
```
walec51: DTOs should represent what came from the exchange, this is premature adaptation and should not be here

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
```diff
@@ -47,9 +36,7 @@ public static ExchangeMetaData adaptMetadata(
       BigDecimal minAmount = product.getBase().getLimits().getMin();
       BigDecimal maxAmount = product.getBase().getLimits().getMax();
       Integer baseScale = currency.get(Currency.getInstance(product.getBase().getId())).getScale();
-      Integer priceScale =
-          currency.get(Currency.getInstance(product.getQuote().getId())).getScale();
-      BigDecimal increment = product.getQuote().getIncrement();
+      BigDecimal increment = product.getQuote().getIncrement().stripTrailingZeros();
```
walec51: is this actually better then the previous implementation?

xchange-lgo/src/main/java/org/knowm/xchange/lgo/service/LgoTradeService.java
```diff
@@ -131,61 +116,69 @@ private LgoProduct getProduct(CurrencyPair currencyPair) {
...
 
   @Override
-  public boolean cancelOrder(String orderId) throws IOException {
-    LgoPlaceOrder lgoOrder = LgoAdapters.adaptCancelOrder(orderId, new Date());
-    placeOrder(lgoOrder);
-    return true;
-  }
-
-  private String placeOrder(LgoPlaceOrder lgoOrder) throws IOException {
+  public String placeMarketOrder(MarketOrder marketOrder) throws IOException {
```
walec51: why is sending the unencrypted order by default better?

xchange-lgo/src/main/java/org/knowm/xchange/lgo/dto/product/LgoProductCurrency.java
BodySplash: Fix. Thank you.

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
BodySplash: Sorry, my PR description wasn't quite explicit. 
The issue here was that while the quote currency (USD) by itself has a scale of 4, in practice in a CurrencyPair it is limited to the price increment scale (without trailing 0).
This implementation avoid some confusions between the two, while retaining the capacity for LGO to change price increment value in the future. 

I also updated the `verifyOrder` in `LgoTradeService` to be more explicit (hopefully)

xchange-lgo/src/main/java/org/knowm/xchange/lgo/service/LgoTradeService.java
BodySplash: Good question. Encryption is about anti front running. It's a way for LGO to prove that no manipulation has occurred on a given batch. We have some explanations [here](https://lgono.de/)

The unencrypted support was added a few months ago, so now it's a matter of choice for a market participant to activate or not anti frontrunning protection. Of course, that raises the question of which version should be the default here, given the fact that LGO, as a regulated exchange, is also audited by more regular means. 
Since encryption comes with a cost, if you are not aware of this feature, that means that you would always be a little bit slower to place your orders than customer not using it. 

Thinking out loud, maybe a better implementation here to avoid custom methods would be to add a configuration parameter for LgoExchange, and check its value here.

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
walec51: ok

xchange-lgo/src/main/java/org/knowm/xchange/lgo/service/LgoTradeService.java
walec51: I agree, having a exchange specific parameter in exchange specification that would determine if we send encrypted orders or not would be better

xchange-lgo/src/main/java/org/knowm/xchange/lgo/service/LgoTradeService.java
```diff
@@ -131,61 +119,69 @@ private LgoProduct getProduct(CurrencyPair currencyPair) {
...
-      throw LgoErrorAdapter.adapt(e);
+  /** Place a market order encrypting it's content. */
+  public String placeEncryptedMarketOrder(MarketOrder marketOrder) throws IOException {
+    LgoPlaceOrder lgoOrder = LgoAdapters.adaptEncryptedMarketOrder(marketOrder);
+    return placeEncryptedOrder(lgoOrder);
+  }
+
+  @Override
+  public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
+    if (orderParams instanceof CancelOrderByIdParams) {
```
walec51: an exception should be thrown if the required param is missing

BodySplash: I couldn't find any other exchange implementations that throws in this method. Do you expect the orderId to be null ?

xchange-lgo/src/main/java/org/knowm/xchange/lgo/service/LgoTradeService.java
BodySplash: Done.

Nephtys:
I've also added the price history (candlesticks) feature

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
```diff
@@ -1,39 +1,35 @@
...
-import org.knowm.xchange.lgo.dto.trade.LgoUserTrade;
-import org.knowm.xchange.lgo.dto.trade.LgoUserTrades;
+import org.knowm.xchange.lgo.dto.product.*;
+import org.knowm.xchange.lgo.dto.trade.*;
 
 public final class LgoAdapters {
 
   private LgoAdapters() {}
 
+  private static SimpleDateFormat dateFormat;
```
walec51: `private static final SimpleDateFormat ISO_DATE_FORMAT;`

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
```diff
@@ -207,4 +201,8 @@ public static LimitOrder adaptEntryToLimitOrder(
         null,
         new BigDecimal(entry[0].toString()));
   }
+
+  public static String toISODateString(Date date) {
```
walec51: name this method `adaptDateParam`

xchange-lgo/src/main/java/org/knowm/xchange/lgo/dto/marketdata/LgoPriceHistory.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class LgoPriceHistory {
+
+  private final List<LgoCandlestick> prices;
+
+  private LgoPriceHistory(List<LgoCandlestick> prices) {
+    this.prices = prices;
+  }
+
+  public static LgoPriceHistory fromRawValues(List<List<Object>> rawPriceHistory)
```
walec51: use `@JsonFormat(shape = Shape.ARRAY)` with `LgoCandlestick` instead of doing manual parsing

Nephtys:
Sorry for the new feature, I thought this PR was already merged and did not verify

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
Nephtys: done

xchange-lgo/src/main/java/org/knowm/xchange/lgo/LgoAdapters.java
Nephtys: done

xchange-lgo/src/main/java/org/knowm/xchange/lgo/dto/marketdata/LgoPriceHistory.java
Nephtys: done. Nice Jackson feature that I wasn't aware of, thank you

PR #3335, opened 2019-12-19
[Bitfinex] Use detailed exception class when requesting data from exchange

SergeyLukashevich:
Add missing exceptions adaptation based on the returned error text

### DIFF
--- a/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/service/BitfinexMarketDataServiceRaw.java
+++ b/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/service/BitfinexMarketDataServiceRaw.java
@@ -6,6 +6,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import org.knowm.xchange.Exchange;
+import org.knowm.xchange.bitfinex.BitfinexErrorAdapter;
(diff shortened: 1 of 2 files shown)


xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/service/BitfinexMarketDataServiceRaw.java
```diff
@@ -77,15 +78,23 @@ public BitfinexLendDepth getBitfinexLendBook(String currency, int limitBids, int
...
-    List<CurrencyPair> currencyPairs = new ArrayList<>();
-    for (String symbol : bitfinex.getSymbols()) {
-      currencyPairs.add(BitfinexAdapters.adaptCurrencyPair(symbol));
+    try {
+      List<CurrencyPair> currencyPairs = new ArrayList<>();
+      for (String symbol : bitfinex.getSymbols()) {
+        currencyPairs.add(BitfinexAdapters.adaptCurrencyPair(symbol));
+      }
+      return currencyPairs;
+    } catch (BitfinexException e) {
```
walec51: this type of mapping should be done in non-Raw *Service classes
Raw *Service classes should throw exchange native exceptions

see: https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-7

PR #3336, opened 2019-12-19
[Bitmex] Add more input data validity checks

SergeyLukashevich:
Make sure we have all headers from exchange before rate limit calculation;
Make symbols parsing not so strict, and allow to skip unrecognisable symbols instead of error;
Polish tests logback.xml.

### DIFF
--- a/xchange-bitmex/src/test/resources/org/knowm/xchange/bitmex/logback.xml
+++ b/xchange-bitmex/src/test/resources/org/knowm/xchange/bitmex/logback.xml
@@ -2,17 +2,22 @@
 
 <configuration>
 
-    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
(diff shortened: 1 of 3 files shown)


xchange-bitmex/src/test/resources/org/knowm/xchange/bitmex/logback.xml
```diff
@@ -2,22 +2,17 @@
 
 <configuration>
 
-    <!-- Standard console appender for checking activity (short on detail) -->
-    <appender name="CONSOLE_APPENDER" class="ch.qos.logback.core.ConsoleAppender">
-        <!-- Simplified standard logging encoder -->
+    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
```
walec51: should this go to the PR?

PR #3338, opened 2019-12-21
[Bitmex]Adding support for page limit on Bitmex getTradeHistory

makarid:


xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/service/BitmexTradeService.java
```diff
@@ -176,9 +170,16 @@ public UserTrades getTradeHistory(TradeHistoryParams params) throws IOException
       startTime = timeSpan.getStartTime();
       endTime = timeSpan.getEndTime();
     }
+    int count = 100;
```
amjad1210: This will be a problem as you're limiting the count to a maximum of 100 trades but what if the trade count you receive is > limit.

Please refer to @walec51 comments on #3239

makarid: Hello, the 100 is the default limit if you do not put any number, that is why i put it.

amjad1210: @makarid My mistake i misunderstood that you were just adding support for limit.

PR #3339, opened 2019-12-23
Fixed bug in BinanceErrorAdapter

milczarekIT:
Fixed bug introduced in https://github.com/knowm/XChange/pull/3299.
It was introduced in 4.4.2-SNAPSHOT, so it was not released.

Bug description:
For error code -1013 from Binance was returned ExchangeUnavailableException instead of OrderAmountUnderMinimumException or OrderNotValidException.

Fixed in this PR.

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceErrorAdapter.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceErrorAdapter.java
@@ -36,7 +36,7 @@ public final class BinanceErrorAdapter {
           return new ExchangeException(message, e);
         }
       case -1013:
-        createOrderNotValidException(message, e);
+        return createOrderNotValidException(message, e);
       case -1016:
         return new ExchangeUnavailableException(message, e);
       case -1021:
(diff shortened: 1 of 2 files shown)


xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceErrorAdapter.java
```diff
@@ -45,15 +45,16 @@ public static ExchangeException adapt(BinanceException e) {
...
     }
-    return new ExchangeException(message, e);
   }
 
-  private static ExchangeException createOrderNotValidException(
-      String message, BinanceException e) {
+  private static ExchangeException createOrderNotValidException(String message, BinanceException e) {
     if (e.getMessage().contains("MIN_NOTIONAL")) {
       return new OrderAmountUnderMinimumException(message, e);
+    } else {
```
amjad1210: You don't need an else statement here. Even though both methods are correct it comes down to readability by reducing nesting in your code.

milczarekIT: I aligned this to the same convention as few lines above:

```
 if (e.getMessage().contains("insufficient balance")) {
          return new FundsExceededException(e.getMessage(), e);
 } else {
          return new ExchangeException(message, e);
 }
```

PR #3370, opened 2020-01-23
Support for Cryptowatch Market Data

massigerardi:
Added partial support for market data from Cryptowatch


xchange-cryptowatch/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <parent>
+        <groupId>org.knowm.xchange</groupId>
+        <artifactId>xchange-parent</artifactId>
+        <version>4.4.2-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>xchange-cryptowatch</artifactId>
+
+    <name>XChange Kraken</name>
```
timmolter: not Kraken

massigerardi: ops, sorry.
I will fix the name
Cryptowatch is owned by kraken anyway.

PR #3390, opened 2020-02-05
[OkCoin] Margin api support

e1ektr0:
Just add support margin api to raw service. Didn't modify exist code.


e1ektr0:
Not ready, detect problems after testing.

PR #3394, opened 2020-02-09
[Binance] Add asset dividend endpoint

### LINKED ISSUE #3393: [Binance] Distributions API endpoints
Some coins are generating gas coins and some coins are air droped by the exchange itself.
It doesn't exist any endpoint in Xchange with the capability of tracking "distributed/air droped" coins. 

See https://binance-docs.github.io/apidocs/spot/en/#asset-dividend-record-user_data 

Edit : AIRDROP type exist in FundingRecord Type

trixprod:
Fix #3393


xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/AssetDividendList.java
```diff
@@ -0,0 +1,49 @@
...
+  @Data
+  public static final class AssetDividend {
+
+    //    {
+    //           "amount":"10.00000000",
+    //            "asset":"BHFT",
+    //            "divTime":1563189166000,
+    //            "enInfo":"BHFT distribution",
+    //            "tranId":2968885920
+    //    },
```
EVRUSIN: Comments

trixprod: Hello, see DepositList, we also have an example of response structure.
It could be useful. 

But i can remove it if needed.

walec51: instead of comments please a an unit test which shows that the above JSON will be de-serialized into a proper DTO

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceAccountService.java
```diff
@@ -201,8 +195,16 @@ public TradeHistoryParams createFundingHistoryParams() {
...
+      Integer limit = null;
+
+      if (params instanceof TradeHistoryParamLimit) {
+        TradeHistoryParamLimit hpl = (TradeHistoryParamLimit) params;
+        limit = hpl.getLimit();
+      }
+
       boolean withdrawals = true;
       boolean deposits = true;
+      boolean airdrops = false;
```
walec51: if `HistoryParamsFundingType#getType` is not set then this should return all funding history

walec51: as binances documentations states those "dividends" wont always be airdrops, in fact they will be mostly stake rewards:

https://www.financemagnates.com/cryptocurrency/news/binance-launches-staking-service-for-coins-held-in-its-wallets/

until we implement better parsing of the `enInfo` field I'd suggest to add a new value to:

```
public class FoundingRecord {

  /** Enum representing funding transaction type */
  public enum Type {
    WITHDRAWAL(false),
    DEPOSIT(true),
    AIRDROP(true),
    /** Used for inflows that are not a regular users deposit and are either different the the inflows defined above or their nature could not have been deduced from the exchanges response */
    OTHER_INFLOW(true);
    /** Used for outflows that are not a regular users withdrawal and are either different the the outflows defined above or their nature could not have been deduced from the exchanges response */
    OTHER_INFLOW(false);
``` 

and name the boolean here differenty

```
boolean otherInflow = true;
```

trixprod:
See PR #3430

trixprod:
@EVRUSIN , @walec51 i'm closing this.
See PR #3430 for the rework

PR #3401, opened 2020-02-13
[bitfinex] Unexpected market order on book

mdvx:
"BitfinexAdapters - Unexpected market order on book. Defaulting to limit order"

I sent in a market order via XStream and got a dozen these warnings.  Two proposals:
a) this patch
b) just remove the warning at line 503

this is probably because XStream is also using this adaptOrders function when getting updates on the socket

Personally, I prefer this patch, as it extends the functionality of adaptOrders


mdvx:
BitfinexOrderStatusResponse [id=39458872000, symbol=ETHUSD, price=273.37, avgExecutionPrice=273.35, side=buy, type=exchange market, timestamp=1581559027.922, isLive=false, isCancelled=false, wasForced=false, originalAmount=0.1, remainingAmount=0, executedAmount=0.1]

It is worth noting that isLive==false, and therefor this is not technical an OpenOrder, however this also applies to LimitOrders and StopOrders

also, for every processAuthenticatedMessage adaptOrder is getting called a lot (10+ times)

walec51:
ok, however `adaptOrders` is in dire need for a refactor in the future

Issue #3358, opened 2020-01-06
[Core] Currency.compareTo


mdvx:
The following test fails for the last two lines.  compareTo returns 
   btc.compareTo(Currency.LTC) --> -10, 
   xbt.compareTo(Currency.LTC) --> 12

To fix, I think we need to add look at Currency.compareTo, subtracting hashcodes looks like it should work.  CurrencyPair.compareTo looks ok too.  so wondering how to fix.

Changing this may have a big impact, so any ideas?
````
  @Test
  public void testCompareTo() {
    Currency btc = Currency.BTC;
    Currency xbt = Currency.XBT;

    assertThat(btc.equals(btc));
    assertThat(btc.equals(xbt));
    assertThat(xbt.equals(btc));
    assertThat(xbt.equals(xbt));

    assertEquals(btc.compareTo(xbt), 0);
    assertEquals(xbt.compareTo(btc), 0);

    assertEquals(btc.compareTo(Currency.LTC), xbt.compareTo(Currency.LTC));
    assertEquals(btc.compareTo(Currency.ETH), xbt.compareTo(Currency.ETH));
  }
````

**Comment by mdvx** [comment]
My temprary solution seems to be holding out
```
  ....
  @Override
  public int compareTo(Currency o) {
    return attributes.equals(o.attributes) ? 0 : toString().compareTo(o.toString());
  }
  .....
  @Override
  public int compareTo(CurrencyPair o) {
    return equals(o) ? 0 : toString().compareTo(o.toString());
  }
  ....
````

mdvx:
https://github.com/knowm/XChange/issues/2886
https://github.com/knowm/XChange/issues/2720
https://github.com/knowm/XChange/issues/2535

According to LiveCoin XBT -> 
`{
name: "Bricktox",
symbol: "XBT",
walletStatus: "normal",
withdrawFee: 1488,
difficulty: null,
minDepositAmount: 0,
minWithdrawAmount: 0.01,
minOrderAmount: 583.3334
},`

however Bricktox last post is March 2018 https://twitter.com/bricktox?lang=en (the only month it was active, for it's ICO).  In April 2018, Bricktox crashed to almost 0, and it now lives on a zombie coin.

PR #3414, opened 2020-02-21
[Core] Currency.compareTo #3358

mdvx:
The following test fails for the last two lines. compareTo returns
btc.compareTo(Currency.LTC) --> -10,
xbt.compareTo(Currency.LTC) --> 12

To fix, I think we need to add look at Currency.compareTo, subtracting hashcodes looks like it should work. CurrencyPair.compareTo looks ok too. so wondering how to fix.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/currency/Currency.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/currency/Currency.java
@@ -482,13 +482,7 @@ public class Currency implements Comparable<Currency>, Serializable {
 
   @Override
   public int compareTo(Currency o) {
-
-    if (attributes.equals(o.attributes)) return 0;
-
-    int comparison = code.compareTo(o.code);
-    if (comparison == 0) comparison = getDisplayName().compareTo(o.getDisplayName());
-    if (comparison == 0) comparison = hashCode() - o.hashCode();
-    return comparison;
+    return attributes.equals(o.attributes) ? 0 : toString().compareTo(o.toString());
   }
 
   private static class CurrencyAttributes implements Serializable {


mdvx:
@Test
  public void testCompareTo() {
    Currency btc = Currency.BTC;
    Currency xbt = Currency.XBT;

    assertThat(btc.equals(btc));
    assertThat(btc.equals(xbt));
    assertThat(xbt.equals(btc));
    assertThat(xbt.equals(xbt));

    assertEquals(btc.compareTo(xbt), 0);
    assertEquals(xbt.compareTo(btc), 0);

    assertEquals(btc.compareTo(Currency.LTC), xbt.compareTo(Currency.LTC));
    assertEquals(btc.compareTo(Currency.ETH), xbt.compareTo(Currency.ETH));
  }

timmolter:
XBT is not the same as BTC.

mdvx:
https://themoneymongers.com/what-is-xbt-btc/
Is XBT the same as BTC?
Yes, XBT or BTC is the same thing, i.e., BTC. The only reason for the confusion is because some exchanges are using the term XBT while the majority of them are using BTC because that's more

mdvx:
XBT seems to be the ISO code, whereas BTC is the crypto friendly code

mdvx:
https://support.kraken.com/hc/en-us/articles/360001206766-Bitcoin-currency-code-XBT-vs-BTC

Bitcoin currency code: XBT vs BTC
Being that Bitcoin is decentralized, there is no standard, nor governing body, to dictate what notation should be used for it. Nonetheless, "BTC" has been the generally accepted abbreviation for Bitcoin stemming from the early days of Bitcoin.

"XBT" is a new abbreviation for Bitcoin that is starting to come into use, and reflects its growing legitimacy as an international currency.

The abbreviation "XBT" comes from the International Standards Organization (ISO) that maintains a list of internationally recognized currencies. The standard dictates that if a currency is not associated with a particular country, it should begin with an “X”, hence "XBT". Another example of this is the abbreviation for gold, "XAU".

mdvx:
I sense a lot of reluctance on this solution, what would make everyone more comfortable?

We are going to have to deal with all the X codes
https://en.wikipedia.org/wiki/ISO_4217#Cryptocurrencies

timmolter:
Some exchanges use XBT and some BTC. We've had discussions in the past about adding a "unified" symbol converter, but we don't want to do it. I think because there are some symbols that overlap.

PR #3430, opened 2020-02-28
[Binance] Rework of PR #3394

trixprod:
Add simple unit test
Move from airdrop to inflow


xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/AssetDividendResponse.java
```diff
@@ -38,5 +38,25 @@ public String toString() {
     private long divTime;
     private String enInfo;
     private long tranId;
+
+    public BigDecimal getAmount() {
+      return amount;
+    }
```
walec51: lombok is used here, there is no need to add these getter manualy

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceAccountService.java
```diff
@@ -197,8 +191,16 @@ public TradeHistoryParams createFundingHistoryParams() {
         }
       }
 
+      Integer limit = null;
+
+      if (params instanceof TradeHistoryParamLimit) {
+        TradeHistoryParamLimit hpl = (TradeHistoryParamLimit) params;
+        limit = hpl.getLimit();
```
walec51: just do

```
limit = ((TradeHistoryParamLimit) params).getLimit();
```

PR #3439, opened 2020-03-05
[Core] Adding openPosition DTO for Future market support

makarid:
#3171

We can of course change the fields on the OpenPosition.class later. What do  you think?
We could also not extend the Order.class


xchange-core/src/main/java/org/knowm/xchange/dto/trade/OpenPosition.java
```diff
@@ -0,0 +1,110 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.math.BigDecimal;
+import java.util.Date;
+import java.util.Set;
+import org.knowm.xchange.currency.CurrencyPair;
+import org.knowm.xchange.dto.Order;
+
+public class OpenPosition extends Order {
```
walec51: First of this should not be called `OpenPosition` but just `Position` for the same reasons we do not have an `OpenOrder` class.

Secondly: I think that the decision to extend `Order` should be tied to the question: to we want to reuse existing interface methods or not?

if `Position` is a special case of order then it should be returned by `TradeService#getOpernOrders`

on the other hand if we want to keep things separate then I think we can have separate interfaces and in this example add a `FuturesTradeService#getOpenPositions` - most exchanges will not bother with implementing them at all

makarid: Thank you for your input @walec51 . I will change the name.
I will also drop the extending of Order.class
About the last input. @timmolter suggests that we make the things simple and not pop up interfaces for 1 or 2 methods, because the new users will get confused or have a bigger learning curve. I agree with that also. Do you think that it would be necessary to create a FutureTradingService interface just to put this method in it? I believe that it should be best just to add this method to the TradeService interface.

walec51: Personally I'm afraid that our users will confuse methods for normal orders with their counterparts for futures. The difference between a `Position` and an `Order` is very subtle.  I the real world those words are actually synonyms.

For now this only looks like a few methods but as I understand when this will get future complete then each method in the current `TradeService` will have futures counterpart added.

I don't know if this will also be the case for `MarketDataService` and `AccountService` - we would have to research this before we make the decision.

PS. I we ware to add all those methods in `TradeService` then would suggest more verbose names like `FuturePosition`, `getOpenFuturePositions`, `placeFuturePosition`, etc.

makarid: OK, so i will add create this FutureTradeService interface and see how we will tweak it in the future?

walec51: Lets first talk what do we need in MarketDataService and AccountService to make futures work?

xchange-core/src/main/java/org/knowm/xchange/dto/trade/OpenPosition.java
```diff
@@ -0,0 +1,110 @@
...
+
+  public OpenPosition(
+      OrderType type,
+      BigDecimal originalAmount,
+      CurrencyPair currencyPair,
```
walec51: CurrencyPair is not good enough for contracts, see: https://github.com/knowm/XChange/issues/3171#issuecomment-595261528

makarid: Sure, but we don't have a new DTO in order to support that. That is why i will put CurrencyPair for the near future

xchange-core/src/main/java/org/knowm/xchange/dto/trade/OpenPosition.java
makarid: MarketDataService will be the same in my opinion.
On the AccountService we will need info about the real-time margin that we are using.

makarid: As i am trying to write the Position class, i saw that most of the fields and methods that Order is using will be used in Position also. So, maybe we need to extend Order.class in order to prevent a lot of duplication.

walec51: no, MarketDataService would stay the same only if you plan to continue hacking the CurrencyPair parameter as it was in Bitmex:

```
public static final CurrencyPair ETH_H18 = new CurrencyPair(Currency.ETH, Currency.H18)
```

I'm highly against such a solutions, a currency pair should be a currency pair, nothing more - it should not hold parameters specific to a futures contract

makarid: > no, MarketDataService would stay the same only if you plan to continue hacking the CurrencyPair parameter as it was in Bitmex:
> 
> ```
> public static final CurrencyPair ETH_H18 = new CurrencyPair(Currency.ETH, Currency.H18)
> ```
> 
> I'm highly against such a solutions, a currency pair should be a currency pair, nothing more - it should not hold parameters specific to a futures contract

When we will create an Instrument interface and this instrument will be the input of MarketDataService method's then where will be no currencypair hacking. Until this happens we can go one step at the time. Right now the only thing i can do in order to make this openPosition workable is via this PR using currencyPair as input. Why we need to change everything in order to just add one method that will be implemented only by a few exchanges.

Issue #3246, opened 2019-09-24
[bitfinex] Error createExchange


tiagoferraro:
line where the error is triggered
        exchange = ExchangeFactory.INSTANCE.createExchange( BitfinexExchange::class.java.name,APIKEY,SECRET)

Log error 
W/System.err: org.knowm.xchange.exceptions.ExchangeException: si.mazi.rescu.HttpStatusIOException: Cannot deserialize value of type `java.math.BigDecimal` from String "tBTCUSD": not a valid representation
        at [Source: (StringReader); line: 1, column: 3] (through reference chain: java.lang.Object[][0]->org.knowm.xchange.bitfinex.v2.dto.marketdata.BitfinexTicker["ask"])
W/System.err:     at org.knowm.xchange.BaseExchange.applySpecification(BaseExchange.java:117)
        at org.knowm.xchange.ExchangeFactory.createExchange(ExchangeFactory.java:78)

walec51:
just merged a PR that might help, please build the latest snapshot on your machine and test against it

walec51:
closing due to lack of response

GusWard:
I am still getting this issue on Android, seems to be attempting to parse the 'symbol' as a BigDecimal. Could this be something to do with a mismatched parsing library?

msgilligan:
I'm seeing this issue on Android, too. (Using `minSdkVersion 28` and `targetSdkVersion 29`)

veken0m:
I get the same issue on Android.

@GusWard Are you able to clean up your pull request so that it can be merged? The @JsonPropertyOrder fix resolves the issue because for some reason Android deserializes the properties in alphabetical order by default.

walec51:
I'm wondering does this library even work on Android after we moved to Java 11?

PR #3443, opened 2020-03-08
Fixed json order parsing issue

GusWard:
Added fallback to zero for null ticker low

### DIFF
--- a/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/BitfinexExchange.java
+++ b/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/BitfinexExchange.java
@@ -70,7 +70,7 @@ public class BitfinexExchange extends BaseExchange implements Exchange {
       Map<CurrencyPair, BigDecimal> lastPrices =
           Arrays.stream(dataService.getBitfinexTickers(null))
               .map(BitfinexAdapters::adaptTicker)
-              .collect(Collectors.toMap(t -> t.getCurrencyPair(), t -> t.getLast()));
+              .collect(Collectors.toMap(t -> t.getCurrencyPair(), t -> t.getLast() == null ? BigDecimal.ZERO : t.getLast()));
 
       final List<BitfinexSymbolDetail> symbolDetails = dataService.getSymbolDetails();
       exchangeMetaData =
--- a/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
+++ b/xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
@@ -2,12 +2,14 @@ package org.knowm.xchange.bitfinex.v2.dto.marketdata;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import java.math.BigDecimal;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.ToString;
 
 @JsonFormat(shape = JsonFormat.Shape.ARRAY)
+@JsonPropertyOrder({"symbol", "frr", "bid", "bidPeriod", "bidSize", "ask", "askPeriod", "askSize", "dailyChange", "dailyChangePerc", "lastPrice", "volume", "high", "low", "placeHolder0", "placeHolder1", "frrAmountAvailable"})
 @AllArgsConstructor
 @NoArgsConstructor
 @Getter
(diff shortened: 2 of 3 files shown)


xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/BitfinexExchange.java
```diff
@@ -70,7 +70,7 @@ public void remoteInit() throws IOException, ExchangeException {
       Map<CurrencyPair, BigDecimal> lastPrices =
           Arrays.stream(dataService.getBitfinexTickers(null))
               .map(BitfinexAdapters::adaptTicker)
-              .collect(Collectors.toMap(t -> t.getCurrencyPair(), t -> t.getLast()));
+              .collect(Collectors.toMap(t -> t.getCurrencyPair(), t -> t.getLast() == null ? BigDecimal.ZERO : t.getLast()));
```
walec51: null represents a lack of value
zero i zero, the fact that we did not find the last price does not mean it was zero

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
```diff
@@ -2,12 +2,14 @@
...
 import com.fasterxml.jackson.annotation.JsonFormat;
 import java.math.BigDecimal;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.ToString;
 
 @JsonFormat(shape = JsonFormat.Shape.ARRAY)
+@JsonPropertyOrder({"symbol", "frr", "bid", "bidPeriod", "bidSize", "ask", "askPeriod", "askSize", "dailyChange", "dailyChangePerc", "lastPrice", "volume", "high", "low", "placeHolder0", "placeHolder1", "frrAmountAvailable"})
```
walec51: this is just the natural order of properties

I don't see why this would be needed

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
```diff
@@ -36,4 +38,83 @@
   public boolean isFundingCurrency() {
     return true;
   }
+
+  @Override
+  public String getSymbol() {
+    return symbol;
+  }
```
walec51: lombok is used here, this code is not needed

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerTraidingPair.java
```diff
@@ -30,4 +32,59 @@
...
+  }
+
+  @Override
+  public BigDecimal getDailyChangePerc() {
+    return dailyChangePerc;
+  }
+
+  @Override
+  public BigDecimal getLastPrice() {
+    return null;
```
walec51: what?

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/BitfinexExchange.java
GusWard: How would you prefer to handle null value then? as this throws exception when last is null

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
GusWard: I added this to fix this issue - https://github.com/knowm/XChange/issues/3246 which i was seeing also

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
GusWard: Ok will remove these :)

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/BitfinexExchange.java
walec51: I would prefer to add this before the collect:

```
.fitler(t -> t.getLast() != null)
```

xchange-bitfinex/src/main/java/org/knowm/xchange/bitfinex/v2/dto/marketdata/BitfinexTickerFundingCurrency.java
walec51: #3246 relates to a different DTO

please add a unit test first that show that we are unable to parse these DTOs without `@JsonPropertyOrder` annotations because to my understanding on how Jackson works your code did not change anything

GusWard: The issue only happens on Android, i cant seem to reproduce now in unit tests although im sure i could yesterday. Do you think there is anything wrong with explicitly setting the order? if so no problem i will just use my fork

walec51: if its not needed then it should not be there - just that :)

if you don't add a unit test that tests it then one day someone will add a field here and forget to add it in the annotation

but that said we are lacking a CI job that would run tests on the Androids JVM in this project - something to consider

aligokalppeker:
Hey fix the CI @GusWard !!!!! and progress PR to merge !!!!!!!

walec51:
@GusWard will you have some time ti polish this MR?

walec51:
closed because of lack of response from author

Issue #3269, opened 2019-10-17
[core]UserTrade missing field "original volume"


Yampersand:
The UserTrade-Object is missing a field "OriginalVolume" similar to OriginalAmount. It should represent the actual volume of the trade.

At the moment this value has to be calculated using amount and price. That leads to rounding decisions which may differ from how it is handled by the exchange.
Some exchange offer this value via API: Bitcoin.de, Bitstamp, CEX.io (in some cases), Kraken, Poloniex.

An example:
Follow a trade on Bitstamp:
`{
	"usd": "-748.21",
	"btc_usd": 8317.45,
	"order_id": 4178040339,
	"datetime": "2019-10-01 23:15:12",
	"fee": "0.60",
	"btc": "0.08995704",
	"type": "2",
	"id": 98055730,
	"eur": 0.0
}`

resulting in this UserTrade:
`UserTrade[
    type=BID,
    originalAmount=0.08995704,
    currencyPair=BTC/USD,
    price=8317.45,
    timestamp=Wed Oct 02 01:15:12 CEST 2019,
    id=98055730,
    orderId='4178040339',
    feeAmount=0.60,
   feeCurrency='USD'
]`

Calculation the volume of the Trade results in 8317.45 (price) * 0.08995704 (amount) = 748.213182348‬ while the API report 748.21.

Is there a reason not to include this field? If not, I would like to add this.

amjad1210:
@Yampersand This is available in the Order DTO with the method getCumulativeCounterAmount() you could add something similar into UserTrade. The reason this value isn't stored in Order or UserTrade is because it can easily be calculated. As for the rounding issue you can solve this by getting the currency precision from CurrencyPairMetaData e.g.

```
var pairs = exchange.getExchangeMetaData().getCurrencyPairs();
var tradePair = pairs.get(new CurrencyPair("BTC/USD"));
var scale = orderType == OrderType.BID ? tradePair.getCounterScale()  : tradePair.getBaseScale();
```

#3216

Yampersand:
@amjad1210 That is a good hint, thank you for that.  I will see how far that gets me, even if it still feels weird ignoring a provided value just to recalulate it later

walec51:
@Yampersand many exchanges don't provide this value and calculating it is trivial
having code that sometimes calculates it and sometimes does not would be a unnecessary complexity

feel free to reopen this issue if you have a problem with the current approach

Yampersand:
[...]

PR #3447, opened 2020-03-11
UserTrade missing field "original volume"

Yampersand:
As I mentioned in my issue  #3269 some exchanges offer a "original volume" value, similar to OriginalAmount, which is missing from UserTrade.
The argument "[the value] can easily be calculated" is not entirly true as following example of a binace trade shows:
{"symbol":"ETHBTC","id":15919****,"orderId":59246****,"orderListId":-1,"price":"0.01938500","qty":"0.00700000","quoteQty":"0.00013569","commission":"0.00005069","commissionAsset":"BNB","time":15801309*****,"isBuyer":true,"isMaker":false,"isBestMatch":true}

The "CurrencyPairMetaData" defines the scale to 6 digitis which would result in a volume of 0.007 * 0.019385 = 0.000136 insteadt of "quoteQty = 0.00013569".

To compromise between not poluting the UserTrade object and returning the correct value I duplicated the behaviour of the KrakenTradeService, which return the KrakenUserTrade with an additional cost field.
I implemented UserTrade subclasses for Binace, Bitstamp and Poloniex.

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/BinanceTrade.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/BinanceTrade.java
@@ -10,6 +10,7 @@ public final class BinanceTrade {
   public final long orderId;
   public final BigDecimal price;
   public final BigDecimal qty;
+  public final BigDecimal quoteQty;
   public final BigDecimal commission;
   public final String commissionAsset;
   public final long time;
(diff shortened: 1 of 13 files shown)


walec51:
I'm against this PR because it mixes responsibilities of a DTO with that of an adapter

Yampersand:
Could you please clarify how this PR mixes responsibilities?

As I see it my PR only adds an additional field to a DTO, without adapting or manipulating it in any way. It only provides access to a field that can't be aquired any other way, without changing or breaking the API. It will be accesable only for those who explicitly need it.

Maybe my changes to the polinex module are a bit confusing, because I renamed an already existing PoloniesUserTrade object to PoloniexTrade and then added a new PoloniexUserTrade object.
The commit 15473bdaecb0e3e962cdcd381a6f3e6295fbd10d shows this most clearly

walec51:
you should have DTO that represent what the actual exchanges API is returning

and use *Adapter classes to ma from exhcnage specific DTOs to our generic DTOs

PR #3448, opened 2020-03-11
[poloniex] Fixed Adapter due to changed return value from poloniex

Yampersand:
The return value of reason seems to have changed to "airdrop_generic". These modifications should be able to handle old and current reason values.

The Adaptor also uses now the new `FundingRecord.Type OTHER_INFLOW` as adujstmends are no `DEPOSIT`s

### DIFF
--- a/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAdapters.java
+++ b/xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAdapters.java
@@ -1,8 +1,5 @@
 package org.knowm.xchange.poloniex;
 
-import static org.knowm.xchange.dto.account.FundingRecord.Type.DEPOSIT;
-import static org.knowm.xchange.dto.account.FundingRecord.Type.WITHDRAWAL;
-
 import java.math.BigDecimal;
 import java.math.RoundingMode;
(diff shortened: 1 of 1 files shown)


xchange-poloniex/src/main/java/org/knowm/xchange/poloniex/PoloniexAdapters.java
```diff
@@ -283,11 +282,11 @@ public static ExchangeMetaData adaptToExchangeMetaData(
   }
 
   private static FundingRecord adaptAdjustment(PoloniexAdjustment a) {
-    FundingRecord.Type type = DEPOSIT;
+    FundingRecord.Type type = OTHER_INFLOW;
```
walec51: could we detect deposits instead of falling back on OTHER_INFLOW for anything other then an airdrop?

Yampersand: Poloniex return type consists out of three different objects:
`{
	"adjustments": [ ... ],
	"deposits": [ ... ],
	"withdrawals": [ ... ]
}`

If their naming scheme is consistent adjustments can't be deposits. That's way I choose `OTHER_INFLOW`.

PR #3457, opened 2020-03-18
[bithumb] bug fixes - averagePrice

mdvx:
and cumulative quantity

### DIFF
--- a/xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
+++ b/xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
@@ -85,19 +85,15 @@ public class BithumbTradeService extends BithumbTradeServiceRaw implements Trade
 
   @Override
   public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
-
     try {
-      if (orderParams instanceof CancelOrderByPairAndIdParams) {
-          final CancelOrderByPairAndIdParams params = (CancelOrderByPairAndIdParams) orderParams;
-          return cancelBithumbOrder(params.getOrderId(), params.getCurrencyPair());
-      } else if (orderParams instanceof CancelOrderByIdParams && orderParams instanceof CancelOrderByCurrencyPair) {
-        String orderId = ((CancelOrderByIdParams) orderParams).getOrderId();
-        CurrencyPair pair = ((CancelOrderByCurrencyPair) orderParams).getCurrencyPair();
-        return cancelBithumbOrder(orderId, pair);
-      } else {
+      if (!(orderParams instanceof CancelOrderByIdParams && orderParams instanceof CancelOrderByCurrencyPair)) {
         throw new NotYetImplementedForExchangeException(
-            "Only CancelOrderByPairAndIdParams || (CancelOrderByIdParams && CancelOrderByCurrencyPair) supported");
+                "Only CancelOrderByPairAndIdParams || (CancelOrderByIdParams && CancelOrderByCurrencyPair) supported");
       }
+
+      String orderId = ((CancelOrderByIdParams) orderParams).getOrderId();
+      CurrencyPair pair = ((CancelOrderByCurrencyPair) orderParams).getCurrencyPair();
+      return cancelBithumbOrder(orderId, pair);
     } catch (BithumbException e) {
       throw BithumbErrorAdapter.adapt(e);
     }
--- a/xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
+++ b/xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
@@ -86,20 +86,20 @@ public class BithumbTradeService extends BithumbTradeServiceRaw implements Trade
   @Override
   public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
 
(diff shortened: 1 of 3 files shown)


xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
```diff
@@ -96,9 +93,13 @@ public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
       } catch (BithumbException e) {
         throw BithumbErrorAdapter.adapt(e);
       }
+    } else if (orderParams instanceof CancelOrderByIdParams && orderParams instanceof CancelOrderByCurrencyPair) {
```
walec51: `CancelOrderByIdParams` is meant to be used then you can cancel an order by just passing its id
`CancelOrderByCurrencyPair` i meant to be used then you can cancel all orders for a given currency pair

I don't think this is needed, just use `CancelOrderByPairAndIdParams`

besides this code doesn't do proper error handling

mdvx: I disagree with the premise her that interfaces a exclusive.  If this were the case then a client application has no way of knowing which variant of CancelOrderParams an exchange requires.

interface CancelOrderByCurrencyPair extends CancelOrderParams
interface CancelOrderByIdParams extends CancelOrderParams
interface CancelOrderByPairAndIdParams extends CancelOrderByIdParams 

these combinations are not consistent

One fix would be to:
public interface CancelOrderByPairAndIdParams extends CancelOrderByIdParams, CancelOrderByCurrencyPair {
}

however if we look at class TradeHistoryParamsAll,we see that these interfaces are clearly meant to be additive, rather than exclusive.
    implements TradeHistoryParamsTimeSpan,
        TradeHistoryParamPaging,
        TradeHistoryParamsIdSpan,
        TradeHistoryParamOffset,
        TradeHistoryParamCurrencyPair,
        TradeHistoryParamMultiCurrencyPair,
        TradeHistoryParamLimit 
since TradeHistoryParamCurrencyPair and TradeHistoryParamMultiCurrencyPair are mutually exclusive

NB: CancelOrderByPairAndIdParams is used by only two exchanges (bithumb and bx) and we are in the process of fixing bithumb

> besides this code doesn't do proper error handling
PR does no more than extend the existing error handling, if that is incorrect then it should be under another PR

mdvx: SimulatedTradeService:104
```
      if (orderParams instanceof CancelOrderByIdParams
          && orderParams instanceof CancelOrderByOrderTypeParams)
```
KrakenTradeService:60
```
    if (orderParams instanceof CancelOrderByIdParams) {
      return cancelOrder(((CancelOrderByIdParams) orderParams).getOrderId());
    }
    if (orderParams instanceof CancelOrderByUserReferenceParams) {
      return cancelOrder(((CancelOrderByUserReferenceParams) orderParams).getUserReference());
    }
```

xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
```diff
@@ -89,16 +86,20 @@ public String placeLimitOrder(LimitOrder limitOrder) throws IOException {
   @Override
   public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
 
-    if (orderParams instanceof CancelOrderByPairAndIdParams) {
-      try {
-        final CancelOrderByPairAndIdParams params = (CancelOrderByPairAndIdParams) orderParams;
-        return cancelBithumbOrder(params.getOrderId(), params.getCurrencyPair());
-      } catch (BithumbException e) {
-        throw BithumbErrorAdapter.adapt(e);
+    try {
```
walec51: ok, if we are going in this direction:

```
CancelOrderByPairAndIdParams extends CancelOrderByIdParams, CancelOrderByCurrencyPair
```

then this method can be implemented in a much simpler way

```
  @Override
  public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
    try {
      if (!(orderParams instanceof CancelOrderByIdParams && orderParams instanceof CancelOrderByCurrencyPair)) {
                throw new NotYetImplementedForExchangeException(
                    "Only CancelOrderByPairAndIdParams || (CancelOrderByIdParams && CancelOrderByCurrencyPair) supported");
      }
      String orderId = ((CancelOrderByIdParams) orderParams).getOrderId();
      CurrencyPair pair = ((CancelOrderByCurrencyPair) orderParams).getCurrencyPair();
      return cancelBithumbOrder(orderId, pair);
    } catch (BithumbException e) {
      throw BithumbErrorAdapter.adapt(e);
    }
  }
```

xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
walec51: >  interfaces are clearly meant to be additive, rather than exclusive.

ok, your right

xchange-bithumb/src/main/java/org/knowm/xchange/bithumb/service/BithumbTradeService.java
mdvx: I totally agree, i just didn't want to obfuscate it at that time

PR #3461, opened 2020-03-21
Support for Additional Instrument types such as futures and options

douggie:
#3171 
Initial PR for abstract instrument class on which additional concrete classes can be created


mdvx:
Not like a lot of great work, look forward to this being available

walec51:
this is breaking almost the entire API

few exchanges have futures / options and therefore in #3171 we talked to make either separate methods or even separate interfaces for them

btw wont the orders for options / futures have different fields and behaviors then the current `LimitOrder`, `MarketOrder` and `SpotOrder`?

douggie:
Hi Adam,

My understanding was that we agreed to have and abstract class of
instrument of which CurrencyPair is a subclass, and would allow other
concrete classes that inherit from instrument and would be compatible with
all markets, hence I made the use of instrument consistent.

I felt, possibly incorrectly that, we would want the core services such as
trade and market to be reusable across instrument types, hence adding
support for instrument in xchange-core. If we have separate instrument
types then we should need to have other core services interfaces to support
those new types which seemed a bit duplicated.


What parts of the API did you feel I broke most? I tried to make it
backwardly compatible . Happy for other simpler implementations suggestions

From an order type perspective, instrument and order types are really
orthogonal. We can have a limit order for futures, options, bonds, spot
etc. The only difference being different instruments, assuming  consistent
exchange API (which is really the case).

walec51:
lets move the discussion here: https://github.com/knowm/XChange/issues/3171#issuecomment-602098281

this pull request is unreviewable in the current state - we should first focus only on what changes must be made in the xchange-core module and to not modify exchange implementations until we agree on what API changes need to be made

PR #3462, opened 2020-03-21
formatting fix done by our mvn plugin

walec51:


walec51:
> So maven is reformatting the code? I was blaming intelij - lol

yes, you should always commit code that went through maven build and therefore reformatted by it

timmolter:
THere's an IntelliJ plugin that uses the same formatter as Maven, which I use as well.

timmolter:
https://github.com/knowm/XChange/wiki/Code-Style

PR #3463, opened 2020-03-22
[Binance] Add subAccount Transfert History

trixprod:
Add two funding record api fetch,
one for the master account ( with a mandatory sub account email ),
one for the subaccount ( nothing mandatory ).


xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/TransfertHistoryResponse.java
```diff
@@ -0,0 +1,80 @@
...
+  public static final class TransfertHistory {
+    private String from;
+    private String to;
+    private String asset;
+    private BigDecimal qty;
+    private long time;
+
+    public String getFrom() {
+      return from;
+    }
```
walec51: if your using lombok then these setters / getters are not needed

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceAccountService.java
```diff
@@ -287,6 +304,52 @@ public TradeHistoryParams createFundingHistoryParams() {
                 });
       }
 
+      final String finalEmail = email;
```
walec51: I don't think this is needed

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceAccountService.java
```diff
@@ -287,6 +304,52 @@ public TradeHistoryParams createFundingHistoryParams() {
...
+
+      if (email != null) {
+        super.getTransfertHistory(email, startTime, endTime, page, limit)
+            .forEach(
+                a -> {
+                  result.add(
+                      new FundingRecord(
```
walec51: please use builder

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceAccountService.java
```diff
@@ -287,6 +304,52 @@ public TradeHistoryParams createFundingHistoryParams() {
...
+      }
+
+      if (subAccount) {
+
+        Integer type = deposits && withdrawals ? null : deposits ? 1 : 0;
+        super.getSubUserHistory(asset, type, startTime, endTime, limit)
+            .forEach(
+                a -> {
+                  result.add(
+                      new FundingRecord(
```
walec51: please use builder

xchange-core/src/main/java/org/knowm/xchange/dto/account/FundingRecord.java
```diff
@@ -298,7 +298,13 @@ public String toString() {
      * outflows defined above or their nature could not have been deduced from the exchanges
      * response
      */
-    OTHER_OUTFLOW(false);
+    OTHER_OUTFLOW(false),
+
+    /** Used for inflows that come from a sub account to the master account */
+    SUBACCOUNT_INFLOW(true),
```
walec51: I think more generic names would be better as this is to Binance-specific

some exchanges have a flat structure with many account for a given currency and no master

I would suggest:

```
    /**
     * Withdrawal from the exchange a blockchain address or a bank account.
     *
     * Usually results in a withdrawal fee being subtracted
     */
    WITHDRAWAL(false),

    /** Deposit from a blockchain address or a bank account to the exchange */
    DEPOSIT(true),

    /** Used for transfers between exchanges accounts */
    INTERNAL_WITHDRAWAL(false);

    /** Used for transfers between exchanges accounts */
    INTERNAL_DEPOSIT(true),

    /** Used for transfers between exchanges accounts */
    INTERNAL_WITHDRAWAL(false);
```

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/TransfertHistoryResponse.java
trixprod: I haven't created unit test but if you remove getter, are you still able to create UT ?

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceAccountService.java
trixprod: Lambda need final field. email field is null then affected to a value if needed.

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/TransfertHistoryResponse.java
walec51: I'm not telling to remove getters
I'm telling to remove the getter code you have written by hand
lombok will generate getters for you - this is how the `@Data` annotation you have used here works https://projectlombok.org/features/Data

Spyros-Stylianou: file name has a typo

trixprod: > file name has a typo

Fixed all references from Transfert to Transfer ( without the 't' )

PR #3466, opened 2020-03-24
#3171 Support for futures markets in orders

douggie:
Unfortunately I had to refactor two xchange implementations a little for correct json deserialisation.


xchange-therock/src/main/java/org/knowm/xchange/therock/dto/marketdata/TheRockTicker.java
```diff
@@ -3,21 +3,71 @@
...
 import java.math.BigDecimal;
 import java.util.Date;
 import org.knowm.xchange.currency.CurrencyPair;
-import org.knowm.xchange.utils.jackson.CurrencyPairDeserializer;
+import org.knowm.xchange.therock.utils.jackson.TheRockTickerDeserializer;
 
 /** @author Matija Mazi */
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
-public class TheRockTicker {
+@JsonDeserialize(using = TheRockTickerDeserializer.class)
```
walec51: first of all is this needed here for the Instrument refactor to work?

secondly you should not write deserializer for DTOs that can be easily deserialized by Jackson it self if you just use Jacksons annotations properly

xchange-therock/src/test/java/org/knowm/xchange/therock/dto/TheRockDtoTest.java
```diff
@@ -1,115 +0,0 @@
...
-    return parse(filename, theClass);
-  }
-
-  private static <E> E parse(String filename, Class<E> type) throws java.io.IOException {
-    InputStream is = TheRockDtoTest.class.getResourceAsStream("/" + filename);
-    return mapper.readValue(is, type);
-  }
-
-  @Test
-  public void testTicker() throws Exception {
```
walec51: so now we have an over complicated deserialization mechanism to the rock DTOs with less test coverage

restore this test, remove the deserializers, use Jackson annotation on DTOs

besides this looks unrelated to the futures feature - if so then separate this to a different PR

xchange-therock/src/main/java/org/knowm/xchange/therock/dto/marketdata/TheRockTicker.java
douggie: Hi Adam,

The issue I was struggling with, if we leave it as is we have:

`@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
public class TheRockTicker {

  @JsonDeserialize(using = CurrencyPairDeserializer.class)
  private CurrencyPair fundId;`

But when we go to deserialize it, jackson cannot work out the type.

`com.fasterxml.jackson.databind.exc.InvalidTypeIdException: 
Missing type id when trying to resolve subtype of [simple type, class org.knowm.xchange.currency.CurrencyPair]: missing type id property 'type' (for POJO property 'fund_id')`

Thought the `JsonDeserialize` annotation would have resolved the type, but it did not, suggestions welcomed on which annotation might work.

walec51: yes and that is because these annotations:

```
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
@JsonSubTypes({@Type(value = CurrencyPair.class, name = "currencyPair")})
```

on Instrument currently make no sense if we serialize / deserialize CurrencyPair as a string 

a string will never have a property `type` - we will probably have to have a custom deserializer for Insturment to determine its concrete class

I see you have already provided one `InstrumentDeserializer` - which is better then having a custom deserializer for every DTO that uses the Insturment class - this would be a show stopper for this refactor

for now remove those annotation

we also must think how will we serialize / deserialize derivatives

in your `InstrumentDeserializer` you detect it the string is currency pair based by the number of `/` characters - that is ok

but how will we distinguish futures from options? could you give some examples how their ticker symbols look on exchanges?

xchange-therock/src/main/java/org/knowm/xchange/therock/dto/marketdata/TheRockTicker.java
```diff
@@ -3,21 +3,71 @@
...
+public class TheRockTicker implements Serializable {
+  /** */
+  private static final long serialVersionUID = -4662345648477360478L;
 
-  @JsonDeserialize(using = CurrencyPairDeserializer.class)
   private CurrencyPair fundId;
```
walec51: I think we could also just change this to 

```
private String fundId;
```

mapping this to an CurrencyPair here is to early

an adapter should do this - even if the mapping is trivial

xchange-therock/src/main/java/org/knowm/xchange/therock/dto/marketdata/TheRockTicker.java
douggie: Thanks so much for suggestions, yeah a customer deserializer seems the way to go.
It is not ideal, but deserialising from strings we can count the `/`
CurrencyPair (Base`/`Counter)
BTC`/`USD
Futures/Swaps (Base`/`Counter`/`Prompt)
BTC`/`USD`/`200925
Options (Base`/`Counter`/`Prompt`/`StrikePrice`/`Put?Call)
BTC`/`USD`/`200925`/`8956.67`/`P

xchange-therock/src/test/java/org/knowm/xchange/therock/dto/TheRockDtoTest.java
douggie: removed, custom deserializer added to instrument.

xchange-therock/src/main/java/org/knowm/xchange/therock/dto/marketdata/TheRockTicker.java
douggie: shall we just leave this unchanged, it seems to work with the custom instrument deserializer.

xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
```diff
@@ -224,10 +221,27 @@ public void setAveragePrice(BigDecimal averagePrice) {
...
   }
-
+  /**
+   * @deprecated CurrencyPair is a subtype of instrument <br>
+   *     use {@link #getInstrument()} instead like this:
+   *     <blockquote>
+   *     <pre>
+   * getInstrument()
+   * </pre>
+   *     </blockquote>
```
walec51: change this javadoc to

```
/**
  * @deprecated
  * CurrencyPair is a subtype of Instrument - this method will throw
  * an exception if the order was for a derivative
  * <p> use {@link #getInstrument()} instead
  */
```

that example in blockquote is not needed

douggie: updated

timmolter:
Looks good to me.

PR #3467, opened 2020-03-24
DSX Global

pchertalev:
DSX Exchage is going to migrate to new API very soon. Old API going to be stopped.

Announcement text:

Hi , 
We previously emailed you about a big change coming to DSX. This will take place on March, 25th and will take a whole day, so that our team can make sure migration is completed successfully. 
You will need to make sure that your active orders are either cancelled or completed by the time we start migrating to a new platform and perform routine maintenance. 
After the migration is completed, our mobile app will not be immediately available. However, we will do our best to restore it as soon as we can. 
After the upgrade is completed, our current API will be changed. Please take appropriate measures to get prepared so it doesn’t interrupt your trading. Our new API documentation you can find here (https://api.dsxglobal.com/). You will also need to create new API keys to access a new API. 
Your account will be assigned new crypto addresses for making deposits to DSX. Please use those new addresses, as your old ones will no longer be supported. 
Once the upgrade is complete, we’ll send you another email with instructions on how to access the new and improved DSX. 
Thank you for your patience. 
Best regards, 
The DSX Team

#3468


walec51:
please do not introduce a v2 package
just do everything in `org.knowm.xchange.dsx`

pchertalev:
image

timmolter:
@walec51 you are right, but I accepted the PR anyway. Thanks as always for your comments and efforts!!!!

walec51:
ok I'll change the package in a separate PR

 @pchertalev I know the exchanges API version changed but that does not mean that we have to break applications that use generic interfaces and just want to

```
ExchangeFactory.INSTANCE.createExchange(DsxExchange.class);
```

but cannot because DsxExchange package changes

on some exchanges we even use a mixture of V1 and V2 APIs but the package of Exchange and Service classes does not change, version migrations are transparent to our users that use generic interfaces

timmolter:
@walec51 Thanks!

walec51:
https://github.com/knowm/XChange/pull/3470

PR #3472, opened 2020-03-27
[core] add support of exchange candles

mdvx:


timmolter:
This is already handle by the Ticker DTO, no?

mdvx:
That's what I thought, at first. but no, not really, there are important difference between a quote and a Candle, the number one being the period of the bucket,

walec51:
I don't see any difference I render our Tickers in a candle chart every day

https://kryptowalutownia.pl/kryptowaluty/bitcoin/gieldy

also this is redundant:

```
  private final Date openTime, closeTime;
  private final long granularity;

  /** the timestamp of the ticker according to the exchange's server, null if not provided */
  private final Date timestamp;
```

> important difference between a quote and a Candle, the number one being the period of the bucket,

I don't understand, explain

are you just missing a `duration` field in the ticker?

how would you render a list of candles that have different duration?

timmolter:
I thought we had `duration` in the ticker and it defaulted to 24 hours. Maybe it got removed at some point??

walec51:
hmm, I don't see any `duration` field in the git history, I also think that providing a default value would be risky - my by that just my point of view because I mostly use 1h, 3h and 5 min candles

but still for me a list of tickers / candles where any one can have a different duration makes no sense, how would you render them?

I think a better design would be to add holder class that specifies he duration for all tickers inside, something like:

```
@Data
public class Tickers {

  private final List<Ticker> tickers;

  private final Duration interval;
}
```

mdvx:
openTime is start of bucket (modulus minutes)
closeTime is 
a) (openTime + interval) modulus minutes
**or**
b) current time  when (a) has not been reached

mdvx:
Sorry, a Candle and a Tick are different representations of the similar data, and whereas Ticker may have been reused in the past, it is not the right model.

mdvx:
for Historical Candles  (which I hope we can extend to include more data)

walec51:
```
private final Date openTime;
private final Date timestamp;
```

this is exactly the same thing

having a closeTime attribute is pointless if its always equal timestamp + duration

you didn't answer my questions and you didn't explain why its not the wright model

mdvx:
no, one (timestamp) represents when the data is collected, the other (open/close) represent the period over which the data is aggregated.

mdvx:
This a key difference between a Tick and a Candle.  The existing Ticker class over expanded it's reach, it has bid, ask, last, open, which are mix-mash of concepts.  Candle class separates that.

mdvx:
Since getHistoricPrice() returns a list, we may however want to define how that list is sorted.  For the moment I have taken care of that upstream, but that kinda worries me

mdvx:
A rule
a Candle can be derived from a Tick(er) (class instance) and list of Tickers
a Tick(er) can never be derived from a Candle (class instance)

walec51:
I think your just over complicating the fact that you have some missing functions in Ticker

as for `timestamp` your right - I see that various exchanges either pass the generation timestamp or the open timestamp there

so we could add a `openTimestamp` to `ticker`

as for the `duration` field the question is - should it be a property of:
a) a single Ticker / Candle
b) or a collection of them?

if (a) then I would like to see an example of an candle chart with different durations in each candle

walec51:
from a pure language semantics your right that the term `ticker` should only represent the last price of an symbol: https://www.investopedia.com/ask/answers/12/what-is-a-stock-ticker.asp

however in this libraries domain model its more of a candle then a ticker and I think we should keep it that way

mdvx:
My first pass it this was , great I will use Ticker class, but then i realized just how different the Concepts are, which is why I created Candle, at first i wanted derive Candle from Ticker, but that did not work.

now we can smudge them together, and create TickCandle, but lets ensure we know what we know mean by the Ticker class.

also remember, closeTime of a Candle does not always equal openTime + interval, especially when considering the last interval

mdvx:
marc@qandlestick.com

walec51:
in stock terminology a ticker is: https://www.investopedia.com/ask/answers/12/what-is-a-stock-ticker.asp

so in a perfect world a correct implementation would be

```
class Ticker {
  CurrencyPair currencyPair;
  BigDecimal lastPrice;
  Date timestamp:
}
```

and nothing more

however in the real world no exchange API returns just this in the ticker endpoint - they usualy return data that resemble the last candle 

also a list of ticker like the one above would have no sense so this method should not exist in MarketDataService:

```
  default List<Ticker> getTickers(Params params) throws IOException {
```

this is why a `Ticker` in our library is actually a `Candle`

a `TickCandle` makes no sense to me

I'm even in favor of changing the name of this class from `Ticker` to `Candle` in some future release and renaming `MarketDataService#getTicker` and `MarketDataService#getTickers` to `MarketDataService#getCurrenctCandle` and ``MarketDataService#getCandles`. But that's a topic for another PR.

walec51:
> also remember, closeTime of a Candle does not always equal openTime + interval, especially when considering the last interval

please give some examples other then the last current candle

mdvx:
> > also remember, closeTime of a Candle does not always equal openTime + interval, especially when considering the last interval
> 
> please give some examples other then the last current candle

An exchange that is open from 8am to 8pm.  the 8pm candle is 8pm-8am (12 hours).  This is however a little however a little stretch.  The import thing is the last 10 minute candle can be 1pm-1:01pm, then 1pm-1:02pm etc...

further down the line, Candles can be used to represent periods of equal volume, by varying time period, 1pm-1:05pm, 1:05pm-1:07pm, 1:07pm-1:11pm  - but again this a more advanced use case,

mdvx:
getTicker does not have support for granularity
getPriceHistory requests the Candles from the exchanges, at the desired granularity

mdvx:
image

timmolter:
From a historical perspective a `Ticker` has the fields that is does because that's how MtGox did it and all other exchanges that followed. `Ticker` or `Candle`, it's more or less the same thing and I don't want to change the name as it will cause more harm than good at this point. The original tickers that exchanges provided had a default duration of 24 hours. I understand now that some exchanges provide finer grained `Tickers`. The problem I have with your Candles is that it copypastas a lot of what's already in the ticker class. Additionally, the rule of thumb for adding generic DTOs and methods to XChange is that the addition is only  justified when the majority of exchanges provide it. However if adding a field somewhere (like in Ticker) in an unobtrusive way (like leaving it null as default), we'll most likely let it slip in. For one-off cases, the desired approach is to use the non-generic (i.e. *Raw) class directly.

PR #3482, opened 2020-04-02
[Gemini] Candles in raw service

mdvx:
and v2 tickers


xchange-gemini/src/main/java/org/knowm/xchange/gemini/v2/dto/marketdata/GeminiCandle.java
```diff
@@ -0,0 +1,28 @@
+package org.knowm.xchange.gemini.v2.dto.marketdata;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import java.math.BigDecimal;
+
+@JsonFormat(shape = JsonFormat.Shape.ARRAY)
+public class GeminiCandle {
+  public Long time;
+  public BigDecimal open, high, low, close, volume;
```
walec51, line 9: please make this a value object like `GeminiTicker2` with private final fields

as this is (de)serialized as an array you don't have to annotate constructor args with `JsonProperty` - just make sure the orders is correct

PR #3487, opened 2020-04-11
[bitmex, core] Improved getFundingRecordHistory of BitMEX

Yampersand:
- Added support for paging through results

- Created adapter methods to remove redundant mapping of "BTC/XBT" to "XBt" and vice versa.

- Added support for additional `FundingRecord.Type`s: `REALISED_PROFIT `and `REALISED_LOSS`, which is kind of a interest you get from your futures and is added to or subtracted form your balance

- Corrected the parsing of the Timestamp


xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/BitmexAdapters.java
```diff
@@ -352,37 +372,58 @@ private static OrderType convertType(String side) {
...
+    dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
 
-    try {
-      dateFunding = dateFormat.parse(walletTransaction.getTransactTime());
-    } catch (ParseException e) {
-      throw new IllegalArgumentException(e);
+    Date dateFunding = null;
+    if (walletTransaction.getTransactTime() != null) {
+      try {
+        dateFunding = dateFormat.parse(walletTransaction.getTransactTime());
```
walec51, line 381: we could just use `Date` instead of `String` for the `WalletTransaction#transactTime` attribute and proper Jackson annotations to make Jackson parse this

xchange-core/src/main/java/org/knowm/xchange/dto/account/FundingRecord.java
```diff
@@ -304,7 +304,13 @@ public String toString() {
...
 
     /** Used for transfers between exchanges accounts */
-    INTERNAL_DEPOSIT(true);
+    INTERNAL_DEPOSIT(true),
+
+    /** Used for realised losses from derivatives */
+    REALISED_LOSS(false),
+
+    /** Used for realised profits from derivatives */
+    REALISED_PROFIT(true);
```
walec51, line 313: is this something that exists on many exchanges with derivatives or is it something specific to bitmex?

could you provide an example of another exchange that has a similar type in its API?

xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/BitmexAdapters.java
Yampersand: Done, see 8014b9504e8811a175bed5ec506bd54f878ccb27

xchange-core/src/main/java/org/knowm/xchange/dto/account/FundingRecord.java
Yampersand: Realised Profit and Loss, Realised PNL or RPL, is a standard term in trading derivatives. (There is also UnrealisedPNL but I think that can be represented through to status `PROCESSING`)

I know that at least Deribit and CMC Markets (not support jet) allow trading with derivatives. Deribits explicitly calls those earning RPL in their UI. Through their API they return a type called `settlement` with has a field `profit_loss`:
```
{
"type":"settlement"
"timestamp":1587110300141
"session_profit_loss":4469.243150033
"profit_loss":14.925158857
"position":78675
"mark_price":169.98
"instrument_name":"ETH-PERPETUAL"
"index_price":170.06
"funding":-0.008446859
}
```

walec51: ok, thank you to explaining

PR #3493, opened 2020-04-17
Fixed Cryptowatch module

massigerardi:
- fixed Cryptowatch url
- fixed missing start metadata
- added adapter for CurrencyPair to String
- added configuration to log error in results
- added integration tests
- fixed OHLC format

### DIFF
--- a/xchange-cryptowatch/src/main/java/org/knowm/xchange/cryptowatch/service/CryptowatchBaseService.java
+++ b/xchange-cryptowatch/src/main/java/org/knowm/xchange/cryptowatch/service/CryptowatchBaseService.java
@@ -4,6 +4,7 @@ import org.knowm.xchange.Exchange;
 import org.knowm.xchange.cryptowatch.Cryptowatch;
 import org.knowm.xchange.service.BaseExchangeService;
 import org.knowm.xchange.service.BaseService;
+import si.mazi.rescu.ClientConfig;
 import si.mazi.rescu.RestProxyFactory;
 
 public class CryptowatchBaseService extends BaseExchangeService implements BaseService {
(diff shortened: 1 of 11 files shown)


xchange-cryptowatch/src/main/java/org/knowm/xchange/cryptowatch/service/CryptowatchBaseService.java
```diff
@@ -12,8 +13,10 @@
 
   public CryptowatchBaseService(Exchange exchange) {
     super(exchange);
+    ClientConfig clientConfig = getClientConfig();
+    clientConfig.setIgnoreHttpErrorCodes(true);
```
walec51, line 17: why would you do that?

massigerardi: I think it is a better way to get the error code  than to catch the exception in the code.
see CryptowatchMarketDataServiceRaw.checkResult:70

it is equivalent to set `rescu.http.ignoreErrorCodes = true` in `rescu.properties` 
>  the HTTP response body never be parsed as Exception but always as the method response type. 
By setting it directly in the config, it is ignored by other clients running alongside this client

walec51: in general I disagree - if you implement exceptions properly like described here: https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-7

then you can access both the status code and response body 

however this is a topic for another PR as exception handling in this hole modules should get rewritten

massigerardi: I think the main focus of the best practice is:

> Exchange-specific Exceptions should NOT be leaked out of the generic interface methods. 

handling the exception as part of the result would satisfy the condition above and the module will throw anyway an ExchangeException as required.

massigerardi:
@walec51 thanks

PR #3504, opened 2020-05-04
[okcoin] future bills

andre77:
added future bills method


andre77:
api doc of the method: https://www.okex.com/docs/en/#futures-query

xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/dto/MarginMode.java
```diff
@@ -1,6 +1,28 @@
...
+  empty // special case, for some wallets okex returns empty margin mode
+;
+
+  @JsonCreator
+  public static MarginMode create(String s) {
+    if (s == null) {
+      return null;
+    }
+    if (s.isEmpty()) {
+      return empty;
```
walec51, line 19: wouldn't it be better to just return null in this case also?

xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/service/OkexAccountServiceRaw.java
```diff
@@ -63,7 +85,8 @@ public OkexWithdrawalResponse withdrawal(OkexWithdrawalRequest req) throws IOExc
     FuturesPositionsResponse res =
         okex.getFuturesPositions(apikey, digest, timestamp(), passphrase);
     res.checkResult();
-    return res.getHolding().get(0);
+    List<List<FuturesPosition>> holding = res.getHolding();
+    return holding == null || holding.isEmpty() ? new ArrayList<>() : holding.get(0);
```
walec51, line 89: `Collections.EMPTY_LIST` instead of `new ArrayList<>()` would be better

xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/dto/account/FuturesBillsResponse.java
```diff
@@ -0,0 +1,44 @@
+package org.knowm.xchange.okcoin.v3.dto.account;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy.SnakeCaseStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import java.math.BigDecimal;
+import java.util.Date;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data
```
walec51, line 10: could we use `@Value` instead of `@Data`?

xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/dto/MarginMode.java
andre77: sure, we could use null here. But what if you need to differ between null and empty?
was not sure and decided to keep empty type in place

xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/service/OkexAccountServiceRaw.java
andre77: you mean Collections.emptyList() :-)

xchange-okcoin/src/main/java/org/knowm/xchange/okcoin/v3/dto/account/FuturesBillsResponse.java
andre77: sure, @Value is a good option here, but i simply wanted to be consistent to other classes in this package

PR #3518, opened 2020-05-09
[stream-core] Improve consistency of use of non-blocking calls

badgerwithagun:
A few improvements to use of RX in the core streaming API.

 * Support non-blocking rate limit acquisition
 * Allow callers to hook into completion of the (already non-blocking) sending of socket messages.
 * Fix (unobserved) bug where connection emitters were fired before the socket was ready

Still being tested. Presented as draft for comment.


gewure:
any update?

walec51:

PR #3526, opened 2020-05-20
Websocket ping on idle + default idle time

earce:
Ping is triggered on open websocket when idle time is reached. Idle time is defaulted to 15 seconds, this is required to trigger ` protected void handleIdle(ChannelHandlerContext ctx)`.


mdvx:
We should just watch out that this does not break any existing connections, that dont expect this Ping frame, right now we have less than ten, and I know a few already do this in the @Override of handleIdle.  I am ok with the risk because the fix is simple (override with an empty function) for any exchange that might have a problem with this.

earce:
Gemini and Lgo have it explicitly declared (will be redundant override after this change).

My tests have covered CoinbasePro, Binance, Bitmex, Kraken, Bitstamp. 

Bitfinex has it in their documentation.

I am comfortable moving forward with this change, let me know if anything else should be changed, otherwise who can approve the PR?

Issue #3530, opened 2020-05-21
Idex Unreconginzed field "slug"


gulteking:
When i tried to call getExchangeMetaData() metadata or getExchangeSymbols(), application throws exception below.

I found the cause, while getting metadata, It calls "https://api.idex.market/returnCurrencies" from IdexMarketDataService. Response contains `"slug"` filed, but assigned dto(`ReturnCurrenciesResponse`) not contains slug value. Because of this, we are getting error. 
to solve error, just need to ignore unknown fields while creating objectMapper. 

```
        ObjectMapper objectMapper = new ObjectMapper();
        objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
        return objectMapper.readerFor(ReturnCurrenciesResponse.class).readValue(inputStreamReader);
```

I will solve create a pull request for this soon.

Exception stacktrace: 

`com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "slug" (class org.knowm.xchange.idex.dto.IdexCurrMeta), not marked as ignorable (3 known properties: "decimals", "name", "address"])
 at [Source: (InputStreamReader); line: 1, column: 194] (through reference chain: org.knowm.xchange.idex.dto.ReturnCurrenciesResponse["PLU"]->org.knowm.xchange.idex.dto.IdexCurrMeta["slug"])
	at com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException.from(UnrecognizedPropertyException.java:61)
	at com.fasterxml.jackson.databind.DeserializationContext.handleUnknownProperty(DeserializationContext.java:843)
	at com.fasterxml.jackson.databind.deser.std.StdDeserializer.handleUnknownProperty(StdDeserializer.java:1206)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.handleUnknownProperty(BeanDeserializerBase.java:1592)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.handleUnknownVanilla(BeanDeserializerBase.java:1570)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:294)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:151)
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer._readAndBindStringKeyMap(MapDeserializer.java:527)
[...]

PR #3531, opened 2020-05-22
fix for issue:  #3530

gulteking:
Idex Unreconginzed field "slug"

### DIFF
--- a/xchange-idex/src/test/java/org/knowm/xchange/idex/IdexExchangeTest.java
+++ b/xchange-idex/src/test/java/org/knowm/xchange/idex/IdexExchangeTest.java
@@ -1,0 +1,34 @@
+package org.knowm.xchange.idex;
+
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.knowm.xchange.Exchange;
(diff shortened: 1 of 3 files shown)


xchange-idex/src/test/java/org/knowm/xchange/idex/IdexExchangeTest.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.knowm.xchange.Exchange;
+import org.knowm.xchange.ExchangeFactory;
+import org.knowm.xchange.currency.CurrencyPair;
+import org.knowm.xchange.dto.meta.ExchangeMetaData;
+
+import java.util.List;
+
+public class IdexExchangeTest {
```
badgerwithagun, line 13: Test name should be suffixed `..Integration` so it's run with failsafe instead of surefire. Otherwise if IDEX is down this test will fail and break the build.

gulteking:
Sorry guys for very late answer. @walec51 , thanks for commit

PR #3533, opened 2020-05-23
[streaming] ping related cleanup, inspection cleanup

earce:
Cleaning up some redundant parts of the code related to ping.

Log + Gemini both override method `protected void handleIdle(ChannelHandlerContext ctx)` which is now part of `NettyStreamingService.java` behavior so have removed these overrides. Since Gemini's idle time was 15 seconds which is Netty's default --have switched to use that default var.

In my QA environment I saw https://github.com/bitrich-info/xchange-stream/issues/502 present itself and `pingStatus` treated as an unknown message type which I have now added to the switch to log as a `warn` given this presents itself when we are hitting rate limits. Not sure if another log level makes more sense? Additionally, is this issue only a product of attempting to open many WebSocket connections quickly, or can it occur by hammering a single WebSocket channel with 15+ subscriptions?

Have removed `protected Completable openConnection()` override from `KrakenStreamingService.java` given that the idle ping sends a ping out every 15 seconds of being idle (this will now be a web socket ping rather then an application level ping). @mdvx I saw you had added this in, with what frequency were you seeing disconnects? Was the heartbeat not sufficient to keep the connection alive?

In `NettyStreamingService.java` have cleared up some inspection warnings.


mdvx:
@earce In my case, I can't tolerate the connection going down, and simply reconnecting, as I may miss updates, I need to clear my caches as soon as this is detected (and rebuild them after a fresh connect).

There are a couple PR from the old project I still need to bring over for this (I can only do one at a time)

earce:
@mdvx I ran several tests with the iteration of the code in this review and never saw a disconnect.

Are the PRs you are referring to related to the issues with Kraken connectivity?

mdvx:
No, not Kraken Specifically (but I can;t remember what prompted the changes).  I have migrated them to this new PR now https://github.com/knowm/XChange/pull/3534

badgerwithagun:
@mdvx are you comfortable with these changes, as they overlap with yours?

Are you aware that there is no way to either ensure the connection stays up or to detect if it drops 100% reliably? It's just not a feature of TCP.

I'm currently working on a background PR to add something I use in my app to the library: a wrapper for `Exchange` that adds backup polling when streaming and combines the two into a single stream. It's the best way I've found of resyncing streams when there are undetected lost frames. If you're interested I can put up a work-in-progress PR.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -57,6 +54,9 @@ protected void handleMessage(JsonNode message) {
       KrakenEventType krakenEvent;
       if (event != null && (krakenEvent = KrakenEventType.getEvent(event.textValue())) != null) {
         switch (krakenEvent) {
+          case pingStatus:
+            LOG.warn("PingStatus received: {}", message);
```
badgerwithagun, line 58: Warn is quite high. Does this warrant waking someone up for?

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -163,8 +163,7 @@ protected Completable openConnection() {
                 if (ssl) {
                   SslContextBuilder sslContextBuilder = SslContextBuilder.forClient();
                   if (acceptAllCertificates) {
-                    sslContextBuilder =
-                        sslContextBuilder.trustManager(InsecureTrustManagerFactory.INSTANCE);
+                    sslContextBuilder.trustManager(InsecureTrustManagerFactory.INSTANCE);
```
badgerwithagun, line 166: Looks like the fmt plugin needs rerunning here

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
earce: changed to info

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
earce: reran (nothing changed here but gemini streaming service did)

earce:
@badgerwithagun yes typically with failed writes and timeouts, and this can be tougher to do depending on why the socket connection was broken.

Yes I'd be curious to see how you are combining polling and streaming. To your point of resyncing streams, are do you mean adding in polling to reconcile missed data or something else all together?

Yes I would be curious to this PR. 

@mdvx let me know if there is anything with this review that you need changed.

Also tangental but I know probably due to the ongoing crisis affecting every one the cadence of these releases has changed but does the project have an idea when 5.0.0 may be cut?

mdvx:
> @mdvx I saw you had added this in, with what frequency were you seeing disconnects? Was the heartbeat not sufficient to keep the connection alive?
It was disconnecting within the minute, every minute.

earce:
> > @mdvx I saw you had added this in, with what frequency were you seeing disconnects? Was the heartbeat not sufficient to keep the connection alive?
> 
> It was disconnecting within the minute, every minute.

I didn't see any disconnects after running tests for this with implementation for some time.

mdvx:
We are all a bit out of sync, when I added the onIdle ctx write ping frame, the pingStatus message did not exist, it is possible we are doubling up now.  I am installing the PR as it stands, with my other PRs and checking Gemini works

mdvx:
I see no problems with this implementation :-) @earce @badgerwithagun

mdvx:
Right now, I can't implement 

`class GeminiStreamingExchange { 
  @Override
  public Observable<Throwable> reconnectFailure() {
    return streamingService.subscribeReconnectFailure();
  }
  @Override
  public Observable<Object> connectionSuccess() {
    return streamingService.subscribeConnectionSuccess();
  }
}
`

because Gemini is not using   private Map<CurrencyPair, GeminiProductStreamingService> productStreamingServices;

any thoughts?

badgerwithagun:
I'm merging this for now. Can we move your question to a separate issue, please, @mdvx?

I do think the `info` level logging for `pingStatus` is a bit excessive, but happy to leave that for now for Kraken users to debate over.

mdvx:
https://github.com/knowm/XChange/issues/3537

PR #3534, opened 2020-05-24
[Core-Stream] Roll up of two excluded open PR during merge

mdvx:
https://github.com/bitrich-info/xchange-stream/pull/522
https://github.com/bitrich-info/xchange-stream/pull/566

### DIFF
--- a/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
+++ b/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
@@ -376,8 +376,8 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
                 Subscription newSubscription = new Subscription(e, channelName, args);
                 try {
                   sendMessage(getSubscribeMessage(channelName, args));
-                } catch (IOException throwable) {
-                  e.onError(throwable);
+                } catch (Exception throwable) { // if getSubscribeMessage throws this, it is because it needs to report
+                  e.onError(throwable);         // a problem creating the message
                 }
                 return newSubscription;
               });
@@ -385,7 +385,11 @@ public abstract class NettyStreamingService<T> extends ConnectableService {
         .doOnDispose(
             () -> {
               if (channels.remove(channelId) != null) {
-                sendMessage(getUnsubscribeMessage(channelId));
+                try {
+                  sendMessage(getUnsubscribeMessage(channelId));
+                } catch (IOException e) {
+                  LOG.error("Failed to unsubscribe channel: {}",channelId);
+                }
               }
             })
         .share();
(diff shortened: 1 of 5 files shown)


badgerwithagun:
@mdvx could you run the fmt plugin on this? The formatting changes are making it tricky to diff.

mdvx:
fmt run

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -375,16 +374,23 @@ public void sendMessage(String message) {
...
                   });
             })
         .doOnDispose(
             () -> {
               if (channels.remove(channelId) != null) {
-                sendMessage(getUnsubscribeMessage(channelId));
+                try {
+                  sendMessage(getUnsubscribeMessage(channelId));
+                } catch (IOException e) {
+                  LOG.error("Failed to unsubscribe channel: {}", channelId);
```
earce, line 392: UseCGroupMemoryLimitForHeap

earce, line 392: Do we want to log more information about what the error that occurred here was?

mdvx: Well do we really care why it failed to unsubscribe on an IOException?  I assume the socket has gone anyway, and the message (should be warn) is just to avoid eating the exception.

earce: Probably not, but in some cases this information may yield more color as to what is going on, figured it made sense to consider.

walec51:
+1 looks good, ready to merge?

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
```diff
@@ -375,16 +374,25 @@ public void sendMessage(String message) {
...
                     }
                     return newSubscription;
                   });
             })
         .doOnDispose(
             () -> {
               if (channels.remove(channelId) != null) {
-                sendMessage(getUnsubscribeMessage(channelId));
+                try {
+                  sendMessage(getUnsubscribeMessage(channelId));
```
badgerwithagun, line 390: sendMessage itself is asynchronous - I've been thinking for a while that it should return a `Completable` (see the draft PR I have open). Is that likely to be a problem for you here? Are you only worried about exceptions from `getUnsubscribeMessage` and `getSubscribeMessage`?

badgerwithagun:
Notwithstanding my comment above, merging. It clearly does what it's intended to do for now.

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
mdvx: > Are you only worried about exceptions from getUnsubscribeMessage and getSubscribeMessage?

Yes, they are overrides, so code is not trustworthy

xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java
badgerwithagun: Hm, on principle, eating an unknown exception without a stack trace isn't the best idea IMO.   If you know what it is and want to suppress noise, fine, but as you say below, this could be caused by anything.  I should have spotted this, sorry.

PR #3536, opened 2020-05-25
[stream-core] [stream-wrapper] [WIP] Exchange -> StreamingExchange wrapper

badgerwithagun:
**Definitely work in progress. Posted for comment.**

**Rolls up #3518**

Introduces `PollingExchangeWrapper`, a wrapper for `Exchange` instances that performs several roles:

 - Adds `StreamingExchange` capability to non-streaming `Exchange` implementations, allowing them to be used in a non-blocking way like any other `StreamingExchange`. This avoids the complexity of having to mix-and-match client code when using a variety of different exchanges in the same application. Just use a `PollingStreamingExchangeWrapper` for every exchange and treat them all as RX sources.
 - In doing so, migrates all blocking code into dedicated threads, allowing you to ensure that computation threads remain fully active and can therefore be predictably sized (e.g. to the number of cores).
 - Avoids the need for client code to implement any resilience handling. Most common temporary exchange issues are automatically detected and logged, and exponential backoff is used to stop spamming failing endpoints. Dramatically reduces production failures requiring intervention.
 - Allows graceful fallback where exchanges partially implement the `StreamingExchange` API, detecting this and using polling instead for only those endpoints.
 - Attempts to commonise variations in exchange implementations so that every exchange can be addressed using the same API. Mostly these will be migrated into the individual exchanges over time to ensure consistency, but this works as a temporary stopgap while the difficulties of changing existing exchange implementations can be discussed. However, in some cases, such variations are unlikely to ever be resolved (such as Binance requiring all subscriptions to be made up-front via a `ProductSubscription`) and this removes the need to worry about that: the websocket is automatically disconnected and reconnected transparently without requiring Observables to be recycled.

Usage:

```java
   // Create the Exchange
   Exchange binanceExchange = ExchangeFactory.createExchange(BinanceExchange.class);
   // Create the wrapper
   StreamingExchange exchange = new PollingStreamingExchangeWrapper(
     "binance", binanceExchange, rateController, 1000, true);
   // Start the background worker
   exchange.connect().blockingAwait();
   // Use like any normal StreamingExchange
   // Shut down
   exchange.disconnect().blockingAwait();
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -215,6 +215,7 @@
     <module>xchange-stream-service-pubnub</module>
     <module>xchange-stream-service-pusher</module>
     <module>xchange-stream-service-wamp</module>
+    <module>xchange-stream-wrapper</module>
 
   </modules>
 
--- a/xchange-simulated/src/main/java/org/knowm/xchange/simulated/MatchingEngine.java
+++ b/xchange-simulated/src/main/java/org/knowm/xchange/simulated/MatchingEngine.java
@@ -48,7 +48,7 @@ final class MatchingEngine {
   private final Deque<Trade> publicTrades = new ConcurrentLinkedDeque<>();
   private final Multimap<String, UserTrade> userTrades = LinkedListMultimap.create();
 
-  private volatile Ticker ticker = new Ticker.Builder().build();
+  private volatile Ticker ticker;
 
   MatchingEngine(
       AccountFactory accountFactory,
@@ -69,6 +69,7 @@ final class MatchingEngine {
     this.priceScale = priceScale;
     this.minimumAmount = minimumAmount;
     this.onFill = onFill;
+    this.ticker = new Ticker.Builder().currencyPair(currencyPair).build();
   }
 
   public synchronized LimitOrder postOrder(String apiKey, Order original) {
@@ -172,6 +173,7 @@ final class MatchingEngine {
 
   private Ticker.Builder newTickerFromBook() {
     return new Ticker.Builder()
+        .currencyPair(currencyPair)
         .ask(asks.isEmpty() ? null : asks.get(0).getPrice())
         .bid(bids.isEmpty() ? null : bids.get(0).getPrice());
   }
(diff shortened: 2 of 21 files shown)


badgerwithagun:
@mdvx, @earce as discussed. I'd value any input.

Bear in mind this is very much work in progress. There's a lot I'm not quite happy with yet. I'm trying to take a lot of logic which was deeply embedded in and coupled to [Orko](https://github.com/gruelbox/orko) and make it general-purpose, so the API is a mess. I do think it's really valuably functionality though.

nielsdraaisma:
This is a great feature to add, i'm sure most users of the streaming library have similar functionality in place so it's great to see this added to XChange.

badgerwithagun:
@walec51 one of the big things I want to get right with this is using your new resilience4j based rate limiting in the background threads instead of the temporary approach I'm using here.

However, it needs to fall back to simply using the rate limits in the metadata if the new resilience API isn't implemented for the exchange, and needs to support throttling/backoff (see the `RateController` interface). I'd value any ideas you have on the best way of approaching this.

mdvx:
I actually do this already, with mixed results.  I have two pollers one XChangeOrderPoller and XChangeAccountPoller.  both have limited success.  first thing was to build a dynamic back off(on) policy, based in part on rate limits.

problem 1:  some exchanges require polling every currency pair, and some have a lot.
problem 2: you are never quite in sync.

Challenge 2: I am not certain the AccountPoller is really needed, it should be possible to tract spending via order Order updates, and just use account poll refreshes, to get back in sync, every now and then.  This also applies to some socket exchanges, that dont offer account updates.

@badgerwithagun  @earce

mdvx:
Another thought is to build a poller that is both time interval based, and update event driven.  for instances polling every minute by default, unless and event occurs, then we can get the data right away, assuming we dont blow the rate limits. 

That's not and easy one, but quite doable.

badgerwithagun:
My code does most of what you describe already @mdvx . Polling only if there hasn't been a socket update within a timeframe is a nice idea. I think I'll add it.

earce:
@badgerwithagun just had a chance to start looking through this, minor question why did you go with the `Phaser` as opposed to scheduling the loop with `ScheduledExecutorService.scheduledWithFixedDelay()`

Btw kudos on this, quite a large undertaking, I will be reviewing this in portions of time as it's quite large.

badgerwithagun:
The idea is that the background threads unpark immediately and process a change when it's requested, except when they are blocked due to a rate limiter. When they are working, they work through a queue of tasks until that queue is exhausted.  When there's nothing to do, the threads are parked and not doing anything.

earce:
@badgerwithagun so that's sort of what I figured I saw the wake and thread signaling, however couldn't you accomplish the same thing with the ThreadExecutorService:

- If someone submits a subscription and you need to process it immediately you can just submit a new loop task to that executor service and it will process it immediately.

- Using scheduledWithFixedDelay if you are rate limited you will just block the next run of the loop accomplishing the rate controlling.

gewure:
curious. any update?

PR #3539, opened 2020-05-26
[Streaming] Convert Coinmate streaming service to use new Pushpin API

stachon:


xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingAccountService.java
```diff
@@ -67,16 +66,15 @@ public CoinmateStreamingAccountService(PusherStreamingService service, String us
...
-    return service
-        .subscribeChannel(channelName, "user_balances")
+    return service.subscribeMessages()
         .map(
             (message) -> {
-              Map<String, CoinmateWebsocketBalance> balanceMap =
-                  StreamingObjectMapperHelper.getObjectMapper()
-                      .readValue(
-                          message, new TypeReference<Map<String, CoinmateWebsocketBalance>>() {});
+              Map<String, CoinmateWebsocketBalance> balanceMap = reader.readValue(message.get("balances"));
```
badgerwithagun, line 77: nit: can collapse all this down to an expression lambda

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
```diff
@@ -0,0 +1,53 @@
...
+import info.bitrich.xchangestream.coinmate.dto.auth.AuthParams;
+import info.bitrich.xchangestream.service.netty.JsonNettyStreamingService;
+import io.reactivex.Observable;
+import io.reactivex.subjects.PublishSubject;
+import sun.rmi.runtime.Log;
+
+import java.io.IOException;
+
+/**
+ * The new API uses one WebSocket connection per channel, unlike other exchanges,
```
badgerwithagun, line 13: This could also be achieved be expecting the `ProductSubscription`s to be passed on `connect()`, in the same way as `BinanceStreamingExchange`.  I _think_ this is better, since it removes the need to disconnect/reconnect to change subscriptions - something that always confuses Binance users - but we should be aware we're doing it (and probably move Binance to this approach instead and get rid of `ProductSubscription`)

Something to think about.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingMarketDataService.java
```diff
@@ -51,22 +51,20 @@
...
+              List<CoinmateWebSocketTrade> list = reader.readValue(s);
               return list;
             })
         .flatMapIterable(coinmateWebSocketTrades -> coinmateWebSocketTrades)
         .map(
-            coinmateWebSocketTrade ->
-                CoinmateAdapters.adaptTrade(
-                    coinmateWebSocketTrade.toTransactionEntry(
-                        CoinmateUtils.getPair(currencyPair))));
+            coinmateWebSocketTrade -> CoinmateStreamingAdapter.adaptTrade(coinmateWebSocketTrade, currencyPair)
```
badgerwithagun, line 67: A few problems here:

- Since the socket is scoped to a single subscription, there needs to be an `onDispose` in all these methods to close the `StreamingService` when the `Observable` is disposed
- **However**, it's not a great idea to scope a socket to a single subscription.  The pattern used elsewhere is that the `Observable`s are `share()`ed, so if multiple subscriptions are made, the separate subscribers share the same socket. What I think you need to do here is maintain a `ConcurrentMap` of `CoinmateStreamingService`, re-use the existing one if it already exists and use `Observable.using()` to close the socket when the last subscriber disposes.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
```diff
@@ -0,0 +1,53 @@
...
+ * which use channels over one WebSocket. This is a hack to use JsonNettyStreamingService as
+ * single channel source.
+ */
+public class CoinmateStreamingService extends JsonNettyStreamingService {
+
+  private PublishSubject<JsonNode> messages;
```
badgerwithagun, line 19: I don't think you want or need the `PublishSubject` here. The existing channel subscription approach should work fine. Did you hit problems with it?

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingServiceFactory.java
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  public CoinmateStreamingService createAndConnect(String endpoint, boolean needsAuth) {
+    String url = baseUrl + "/" + endpoint;
+    if (needsAuth && authParams != null) {
+      url += "?" + authParams.toParams();
+    }
+    CoinmateStreamingService service = new CoinmateStreamingService(url);
+    // block until connected, because of nonce conflicts when connecting to multiple channels
+    service.connect().blockingAwait();
```
badgerwithagun, line 23: Rather than blocking the calling thread, which breaks the RX guarantee, how about this:

- Use a `static` `Lock` to prevent concurrent connections
- Return `Completable<CoinmateStreamingService>` from this method (i.e. just return the result of `service.connect()`)

The, the parent service's `connect()` method will work asynchronously as expected.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingTradeService.java
```diff
@@ -1,65 +1,66 @@
...
-            + "-"
+        "channel/my-open-orders/"
             + CoinmateStreamingAdapter.getChannelPostfix(currencyPair);
 
-    return service
-        .subscribePrivateChannel(channelName, "open_orders")
+    ObjectReader reader = StreamingObjectMapperHelper.getObjectMapper()
+        .readerFor(CoinmateWebsocketOpenOrder.class);
+
+    CoinmateStreamingService service = serviceFactory.createAndConnect(channelName, true);
```
badgerwithagun, line 37: Ditto

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingAccountService.java
stachon: Thanks, fixed

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingMarketDataService.java
stachon: Thanks for the suggestions, I am not that good with RxJava, so I am struggling to implement this. I made something like this, to handle both the locking and reusing of connections, but it only works for the first subscriber. I was thinking maybe a `ReplaySubject` would disposing work with it?

```
private ConcurrentMap<String, Observable<CoinmateStreamingService>> serviceMap;
  private static Lock connectionLock = new ReentrantLock();
  private Scheduler scheduler = Schedulers.single();
  public Observable<CoinmateStreamingService> createConnection(String endpoint, boolean needsAuth) {
    String url = baseUrl + "/" + endpoint;
    CoinmateStreamingService service;

    if (serviceMap.containsKey(url)) {
      return serviceMap.get(url);
    } else {
      String authUrl = url;
      // append auth parameters if necessary
      if (needsAuth && authParams != null) {
        authUrl += "?" + authParams.toParams();
      }
      service = new CoinmateStreamingService(authUrl);

      Observable<CoinmateStreamingService> observable = Observable.using(
          () -> service,
          s -> Completable.fromAction(() -> connectionLock.lock())
              .subscribeOn(scheduler)
              .andThen(s.connect())
              .observeOn(scheduler)
              .doFinally(() -> {
                connectionLock.unlock();
              }).andThen(Observable.just(s)),
          s -> s.disconnect()
      ).share();

      serviceMap.put(url, observable);

      return observable;
    }
  }
```

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
stachon: You are right, channel subscription with a fake channel name works as well.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingMarketDataService.java
badgerwithagun: It's not a case for `ReplaySubject`.  You're almost there. I;ll try and have a look at this when I get some time.

stachon: Eventually I used a slightly different approach

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
stachon: Actually, channel subscriptions only work for the first subscriber, so I reverted to using `PublishSubject`

badgerwithagun: As long as they are created with `share()` they should work for multiple subscribers. I have plenty of code in my application that does this.

I suspect the problem was something else.  Could you try putting up your code that uses normal channel subscriptions?  I may be able to help pinpoint the issue.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
```diff
@@ -0,0 +1,53 @@
...
+
+  CoinmateStreamingService(String url) {
+    super(url);
+    this.messages = PublishSubject.create();
+  }
+
+  @Override
+  protected String getChannelNameFromMessage(JsonNode message) throws IOException {
+    return null;
```
badgerwithagun, line 28: I think this should return whatever the channel this service represents (so pass in the channel name as a constructor argument).  That way `subscribeChannel` will work normally.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public String getSubscribeMessage(String channelName, Object... args) throws IOException {
+    return null;
```
badgerwithagun, line 33: Better to `throw new UnsupportedOperationException()`

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public String getUnsubscribeMessage(String channelName) throws IOException {
```
badgerwithagun, line 37: Better to `throw new UnsupportedOperationException()`

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingServiceFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import io.reactivex.Observable;
+import io.reactivex.Scheduler;
+import io.reactivex.schedulers.Schedulers;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+public class CoinmateStreamingServiceFactory {
```
badgerwithagun, line 14: Doesn't need to be public, I think

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
```diff
@@ -0,0 +1,53 @@
...
+import sun.rmi.runtime.Log;
+
+import java.io.IOException;
+
+/**
+ * The new API uses one WebSocket connection per channel, unlike other exchanges,
+ * which use channels over one WebSocket. This is a hack to use JsonNettyStreamingService as
+ * single channel source.
+ */
+public class CoinmateStreamingService extends JsonNettyStreamingService {
```
badgerwithagun, line 17: I don't think this needs to be public.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingServiceFactory.java
stachon: Thanks

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
stachon: Thanks

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
stachon: An exception thrown here would be sent down the Observable chain in subscribeChannel

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
stachon: This would generate a warning log when disposing.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
stachon: You are right. I changed it back to `subscribeChannel` and it works now.

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
nielsdraaisma: It would get caught and logged, but returning null from `getSubscribeMessage` eventually gets handled properly in `sendMessage`, binance follows the same pattern.

stachon: https://github.com/knowm/XChange/blob/develop/xchange-stream-service-netty/src/main/java/info/bitrich/xchangestream/service/netty/NettyStreamingService.java#L378

stachon:
@badgerwithagun Hello, could you please review my changes? Thanks

xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/CoinmateStreamingService.java
badgerwithagun: *bump*

PR #3540, opened 2020-05-27
[simulated] - Implemented additional cancel/getOrder functions 

nielsdraaisma:
Implemented additional functions on the Simulated exchange to allow getOder and cancel calls without passing in the currency pairs.


badgerwithagun:
Nice work, thanks @nielsdraaisma

PR #3541, opened 2020-05-28
[Coinjar] - Initial commit of coinjar exchange

nielsdraaisma:


nielsdraaisma:
@timmolter  or @badgerwithagun , any chance I could get a review on this ?

xchange-coinjar/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+    <!-- Parent provides default configuration for dependencies -->
+    <dependencies>
+
+        <dependency>
+            <groupId>org.knowm.xchange</groupId>
+            <artifactId>xchange-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        
+        <dependency>
```
timmolter, line 31: the test dependencies don't need to be defined here, as the parent pom takes care of it for you already.

nielsdraaisma: I'll keep that in mind for some other PR's ive got coming up, thanks for merging this in.

Issue #3542, opened 2020-05-28
[kraken] Failing to subscribe to websocket API - Unsupported field error


wvos:
The Kraken streaming API doesn't work for a number of subscriptions.  Using the below code works perfectly fine for other exchanges but gives a consistent error for Kraken

**Code** 
```
exchange = StreamingExchangeFactory.INSTANCE.createExchange(KrakenStreamingExchange.class.getName());

exchange.connect().blockingAwait();

Disposable sub = exchange.getStreamingMarketDataService()
                .getOrderBook(CurrencyPair.BTC_USD)
                .subscribe(orderBook -> LOG.info("Order book: {}", orderBook));
```
**Error/logs**
01:38:23.437 [main] INFO  info.bitrich.xchangestream.kraken.KrakenStreamingService - Subscribing to channel book-BTC/USD
01:38:23.475 [nioEventLoopGroup-2-1] INFO  info.bitrich.xchangestream.kraken.KrakenStreamingService - System status: KrakenSystemStatus{connectionID='18409444206880552937', event='systemStatus', status='online', version='1.0.1'}
01:38:23.668 [nioEventLoopGroup-2-1] ERROR info.bitrich.xchangestream.kraken.KrakenStreamingService - Channel book-BTC/USD has been failed: _Unsupported field: 'error' for the given msg type: subscribe_

badgerwithagun:
Hmm, looks like the socket's sending an error back from the subscribe request and `KrakenStreamingExchange` doesn't know how to interpret it.  That's hiding the real cause of the error.

wvos:
Indeed, the serialized subscription message has an 'error' field which is not part of the spec, and also a couple of null fields which causes it to fail

mdvx:
The Kraken Spec is unreliable, and you need to look at the Futures API as well as the spot API

nielsdraaisma:
Is this issue still relevant?, the code above seems to match to what's happening in `KrakenManualExample` and that example seems to work fine. 

If this is still happening it would be great if the actual message could be reported back here to it can be added to the testsuite.

earce:
Fix was merged in

PR #3544, opened 2020-05-29
[Streaming] issue #3542: Remove unused error field and ignore null fields when serializing

wvos:
removing null fields from json serialized subscription message

### DIFF
--- a/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/dto/KrakenEvent.java
+++ b/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/dto/KrakenEvent.java
@@ -9,18 +9,11 @@ public class KrakenEvent {
   @JsonProperty("event")
   private final KrakenEventType event;
 
-  @JsonProperty("error")
-  private String error;
-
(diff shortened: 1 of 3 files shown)


mdvx:
Was the error field recently removed by Kraken?  I was added because it was returned by the REST/Sockets calls

wvos:
there is no error field on the request messages - see: https://docs.kraken.com/websockets/#message-subscribe.  For an error response message https://docs.kraken.com/websockets/#errortypes you have an event 'error' and errorMessage with details which is being dealt with correctly so no changes there.  The change is fix the incorrect formatting of the json request msg. I validated this through the test classes you already have and by using this api in a project

mdvx:
ok, fair enough.  I now suspect it was in a beta API, and has since been cleaned up.  I checked 1.0.1 and 1.1.0

wvos:
@mdvx even though I think the changes I made is relevant, it looks like the code is working in the develop branch.  When I started this I pulled in the snapshot compiled version from https://oss.sonatype.org/content/groups/public/org/knowm/xchange/ which gave me the errors I mentioned.  The changes in this PR ensures the null fields doesn't get serialized although it seems to already happen in the latest version

PR #3555, opened 2020-06-11
[Streaming] Adding proper handlers to subscription for reconnection

earce:
Upon re-connection a subscribe without a proper onError observer is used `EmptyCompletableObserver.java` which has an onError method

`public void onError(Throwable e) {
        this.lazySet(DisposableHelper.DISPOSED);
        RxJavaPlugins.onError(new OnErrorNotImplementedException(e));
    }`

this triggers noisy stack traces in the logs so I have modified to log graceful error messages and to add a proper onError handler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ fi"
 dist: trusty
 
 jdk:
-  - oraclejdk8
   - openjdk11
 
 # whitelist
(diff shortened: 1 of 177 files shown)


earce:
@badgerwithagun @mdvx hey do you guys think you could  opine on this when you have some time?

earce:
Have also added the ability to subscribe to disconnects

badgerwithagun:
This disconnect publusher only seems to emit due a temporary disconnection, not a permanent one. Why is this?  Wouldn't it make sense to emit in both cases?

earce:
@badgerwithagun updated as per your comment

earce:
Have also added some improved logging to `WebSocketClientHandler.java`

earce:
@badgerwithagun hey wanted to bring this back up, want to make sure I get this in as part of this months cut :)

badgerwithagun:
Apologies for the delay - I've been busy with some other things.

Approving this for now, but @mdvx - we were discussing some options for a stream of status updates (i.e. a single stream of messages containing statuses like `CONNECTED, DISCONNECTED, PARTIALLY_CONNECTED` etc).  That would be a better solution here. Are we any closer to a conclusion on that?

earce:
thanks @badgerwithagun much appreciated, and just for my own understanding what you mean here is as opposed to having different methods to subscribe on you would subscribe on a single channel and get the state changes via that single channel?

PR #3568, opened 2020-06-18
Added metadata field

cymp:
orderbook can have metadata provided by an api call (example: bittrex provide a sequence number).


cymp:
done in another PR

PR #3571, opened 2020-06-23
Added fields on OrderBook and Balance

cymp:
- Metadata on OrderBook
- Timestamp on Balance

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
@@ -36,7 +36,7 @@ public final class Balance implements Comparable<Balance>, Serializable {
   private final BigDecimal borrowed;
   private final BigDecimal withdrawing;
   private final BigDecimal depositing;
-  private Date timestamp;
+  private final Date timestamp;
 
   /**
    * Constructs a balance, the {@link #available} will be the same as the <code>total</code>, and
@@ -44,8 +44,9 @@ public final class Balance implements Comparable<Balance>, Serializable {
    *
    * @param currency The underlying currency
    * @param total The total
+   * @param timestamp Time the balance was valid on the exchange server
    */
-  public Balance(Currency currency, BigDecimal total) {
+  public Balance(Currency currency, BigDecimal total, Date timestamp) {
 
     this(
         currency,
@@ -55,7 +56,8 @@ public final class Balance implements Comparable<Balance>, Serializable {
         BigDecimal.ZERO,
         BigDecimal.ZERO,
         BigDecimal.ZERO,
-        BigDecimal.ZERO);
+        BigDecimal.ZERO,
+        timestamp);
   }
 
   /**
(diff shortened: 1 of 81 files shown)


xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
```diff
@@ -34,6 +36,7 @@
   private final BigDecimal borrowed;
   private final BigDecimal withdrawing;
   private final BigDecimal depositing;
+  private Date timestamp;
```
walec51, line 39: Balance should be a value object

final fields, no setters

xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
```diff
@@ -275,26 +278,32 @@ public BigDecimal getDepositing() {
     return depositing;
   }
 
+  /**
+   * Returns the timestamp of the balance
```
walec51, line 282: the is no sense in repeating the word timestamp in the docs

should the client set this field when the data was downloaded or should this information be from the exchanges server?

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
```diff
@@ -131,6 +134,14 @@ public Date getTimeStamp() {
     return type == OrderType.ASK ? asks : bids;
   }
 
+  public Map<String, Serializable> getMetadata() {
+    return metadata;
+  }
+
+  public void setMetadata(Map<String, Serializable> metadata) {
+    this.metadata = metadata;
+  }
```
walec51, line 143: OrderBook should be a value object

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
```diff
@@ -25,6 +26,8 @@
   private final List<LimitOrder> bids;
   /** the timestamp of the orderbook according to the exchange's server, null if not provided */
   private Date timeStamp;
+  /** the metadata, possibly provided by the api */
+  private Map<String, Serializable> metadata;
```
walec51, line 30: this data structure is to loose

if you need exchange specific fields then use methods from *Raw classes

xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
cymp: fixed

xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
cymp: fixed

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
cymp: reverted

cymp:
Thx @walec51, please review again.

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
cymp: reverted

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/OrderBook.java
cymp: reverted

cymp:
forget about it, I'll make a builder

PR #3580, opened 2020-07-01
[bittrex] Added xchange-bittrexV3

cymp:
The current xchange-bittrex module is 90% implemented using the old v1.1 api, mixed with a limited number of calls to undocumented v2 and v3. The v1.1 api is reaching end of life in septembre 2020.
From [Bittrex](https://bittrex.github.io/api/v1-1):
`The v1.1 API will reach its planned end of life and no longer be supported as of 9/30/2020.`

Instead of fixing the old xchange-bittrex module, this new module is meant to have a parallel life until the v1.1 complete unplug, and will then be renamed xchange-bittrex and replace it. 
Reasons:
- Not all of the v1.1 calls have been migrated (candles, withdrawal and deposit histories for example), and applications using the library may be using these legacy calls, which are still valid for a little less than 3 months before nothing works anymore. 
- Not much of the xchange-bittrex source code has been conserved (new API calls, new DTOs, widely changed adapters...)

There is still some work which has to be done in further PR, here is an (incomplete) list:
- candles
- deposit + withdrawal history
- [Error codes](https://bittrex.github.io/api/v3#definition-Error)

### DIFF
--- a/xchange-bittrexV3/api-specification.txt
+++ b/xchange-bittrexV3/api-specification.txt
@@ -1,0 +1,6 @@
+Bittrex Exchange API specification
+================================
+
+Documentation
+-------------
+https://bittrex.com/home/api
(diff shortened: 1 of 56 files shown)


cymp:
@timmolter any idea why the build fails? Looks like a jdk error...

cymp:
I retriggered the build and it looks like the jdk problem is gone.

timmolter:
Integration test classes need to end with "Integration.java".

cymp:
done

PR #3595, opened 2020-07-06
[Gemini] Orderbook stream fix

mdvx:
updates sometimes arrive before the initial book state


xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
```diff
@@ -55,15 +56,21 @@ private boolean filterEventsByReason(JsonNode message, String type, String reaso
...
             .filter(
                 s ->
                     filterEventsByReason(s, "change", "initial")
                         || filterEventsByReason(s, "change", "place")
                         || filterEventsByReason(s, "change", "cancel")
                         || filterEventsByReason(s, "change", "trade"))
+            .filter(
+                s -> // filter out updates that arrive before initial book
+                orderbooks.get(currencyPair) != null
+                        || filterEventsByReason(s, "change", "initial"))
```
earce, line 73: Are these updates published again after the image of the book is received? How are these then added back into the book?

mdvx: Not dealing with collecting out of order updates, just ignores them until it get an initial snapshot of the book.  Its better than the previous way of handling it, than created crossed books.  also since there is no guarantee it will ever get an initial (I have seen that) keep all the deltas until then would likely lead to massive memory consumption.

earce: Makes sense, is it possible to have an inaccurate book if the snapshot came after some updates which were intended to be updates to the book?

mdvx: this may occur, however I have not seen it yet, it would manifest itself as a crossed book (where a delete was missed), which i check for in my code.

mdvx:
On Performance, my cpu usage fell from 50% to 14% with the limited book (3000+ to 100 max entries), using the treemap instead of the ConcurrentHashMap and removing the sort calls.  In that 14% I am running several other exchange books, so the performance improvement is even more drastic.

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
mdvx: Some more color:  I have been using the v2 sockets api now.  When a connection is made it subscribes to l2 data, which includes both trades and order book depth.  The initial orderbook is send out, and then trades and orderbook updates follow.  Have to watch for how we implement here, because if the trades subscription goes out first, on the socket, the "initial" update will be missed by a subsequent subscription (for the orderbook as the observer is shared).

the only 100% safe solution is to have two sockets open (trades, and orderbook), but that means the data is sent twice from the server.  I am pretty certain v1 is behaving in same way, since it explains the weird stuff i have seen in past.

On this update:  Well, it's better than the previous implementation, however i have moved on to v2 api now,  so we can keep it or scrap it.

earce: @mdvx I would say if you are not observing race conditions as you were before from within a single record type (orderbook and its respective updates) this is an improvement.

I would think even though it may redundant having two sockets for this may be better no? Alternatively you would have to keep the state of the world locally to give that to a new subscription as you say. Brainstorming here a bit but let me know if what I'm saying makes sense.

mdvx: So the messages handle both orderbook and trades, with one socket both states can be managed.
However, that means that if someone only subscribes to trades, the orderbook has to be managed in the background as well, just in case they later subscribe to the orderbook.  The opposite order also suffers but is less of a problem, as the newest trades only will be received (only the supplied history is lost), and that is surely acceptable, since we dont model trade history.

But we can not ensure, that the user will subscribe to orderBook before trades, that's when two sockets are needed.

so option a) manage the orderbook when trades are subscribed first, b) use two sockets, c) inform users to subscribe to the orderbook first (dont allow it to be second), d) others??

mdvx: Caveat:  I am basing this off my new experience with the v2 API, and assuming the same problem exist in V1.
but GeminiStreamingMarketDataService.getOrderBook & GeminiStreamingMarketDataService.getTrades could be combined into a single overarching method, to achieve this.

mdvx: NB: same problem will occur if we ever try to support candles

earce: @mdvx the only d) I can think of is to subscribe in the correct order if the user imports this with a flag to explicitly be able to turn this off and then just give the user an observable on the existing subscription (I don't think going this route is  a good idea)

I don't know what the burden of the extra socket really means but I think from a usability and maintainability it sounds easiest?

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/dto/GeminiOrderbook.java
```diff
@@ -1,50 +1,50 @@
...
 import org.knowm.xchange.dto.Order;
 import org.knowm.xchange.dto.marketdata.OrderBook;
 import org.knowm.xchange.dto.trade.LimitOrder;
 
 /** Created by Lukas Zaoralek on 15.11.17. */
 public class GeminiOrderbook {
-  private final Map<BigDecimal, GeminiLimitOrder> asks;
-  private final Map<BigDecimal, GeminiLimitOrder> bids;
+  private final Map<Double, GeminiLimitOrder> asks;
+  private final Map<Double, GeminiLimitOrder> bids;
```
walec51, line 18: why are you downgrading this do `Double`?

`level.getPrice().doubleValue()` can lead to lost precision

leave `BigDecimal`

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/dto/GeminiOrderbook.java
```diff
@@ -54,15 +54,35 @@ public void updateLevels(GeminiLimitOrder[] levels) {
     }
   }
 
-  public OrderBook toOrderbook() {
-    GeminiLimitOrder[] askLevels = asks.values().toArray(new GeminiLimitOrder[asks.size()]);
-    GeminiLimitOrder[] bidLevels = bids.values().toArray(new GeminiLimitOrder[bids.size()]);
+  public OrderBook toOrderbook(int maxLevels, Date timestamp) {
```
walec51, line 57: this class cannot make up its mind is it an adapter or an DTO

can't we move this to GeminiAdapters?

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/dto/GeminiOrderbook.java
mdvx: yes, I was dealing with the horrible performance and trying different things.  BigDecimal is better

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/dto/GeminiOrderbook.java
mdvx: created GeminiAdaptersX

mdvx:
At some point I will add my API V2 implementation

PR #3597, opened 2020-07-07
[core]Proposal for instrument enhancements and FuturesContract

Yampersand:
We are trying to implement the kraken futures and binance futures APIs. While implementing the *Raw services is straight forward, we are struggling with the non-*Raw services.
For some of those challenges this pull request offers solutions:

- some DTOs were still accepting only CurrencyPair, those now support Instrument

- some service methods accept only CurrencyPair, those now accept neutral Params

- There was no OpenPositions method: I added it to AccountInfo, because from our point of view a open positions is quite similar to a balance. 

- I added some interfaces and implementations for *InstrumentParam classes toi mirror existing *CurrencyPairParam classes

- I added a proposal for a FuturesContract, but I'm not very confided that it is already in a state to be used universally like CurrencyPair. As far as I can evaluate it, it is good enough for our use cases.

This proposal does not include any code for an equivalent of CurrencyPairMetaData for Instruments or FuturesContracts. I tried a couple of different approaches to make the Exchange and/ or CurrencyPairMetaData generic but I was not successful without breaking the current API-Interface.
So I' m very open to ideas how to deal with ExchangeMetaData. As far as I can see, that's the last building block preventing us to  implement the futures APIs for Kraken and Binance.

With this proposal I tried to stick closely to the discussion in #3171. My FuturesContract is inspired by #3461

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/Exchange.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/Exchange.java
@@ -7,6 +7,7 @@ import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.dto.meta.ExchangeMetaData;
 import org.knowm.xchange.exceptions.ExchangeException;
 import org.knowm.xchange.exceptions.NotYetImplementedForExchangeException;
+import org.knowm.xchange.instrument.Instrument;
 import org.knowm.xchange.service.account.AccountService;
 import org.knowm.xchange.service.marketdata.MarketDataService;
 import org.knowm.xchange.service.trade.TradeService;
@@ -41,6 +42,16 @@ public interface Exchange {
    * @return The exchange's symbols
    */
   List<CurrencyPair> getExchangeSymbols();
+
+  /**
+   * Returns a list of Instrument objects. This list can either come originally from a loaded json
+   * file or from a remote call if the implementation override's the `remoteInit` method.
+   *
+   * @return The exchange's instruments
+   */
+  default List<Instrument> getExchangeInstruments() {
+    throw new NotYetImplementedForExchangeException();
+  }
 
   /**
    * The nonce factory used to create a nonce value. Allows services to accept a placeholder that is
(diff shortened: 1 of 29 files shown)


xchange-core/src/main/java/org/knowm/xchange/service/trade/params/TradeHistoryParamsAll.java
```diff
@@ -30,6 +34,8 @@
   private Long offset;
   private CurrencyPair pair;
   private Collection<CurrencyPair> pairs = Collections.emptySet();
+  private Instrument instrument;
+  private Collection<Instrument> instruments = Collections.emptySet();
```
walec51, line 38: this should replace the `pair` / `pairs` fields rather then co-oexist with them

you can implement the *CurrencyPair* method using just `instument` / `instruments`

xchange-core/src/main/java/org/knowm/xchange/Exchange.java
```diff
@@ -42,6 +43,16 @@
...
 
+  /**
+   * Returns a list of Instrument objects. This list can either come originally from a loaded json
+   * file or from a remote call if the implementation override's the `remoteInit` method.
+   *
+   * @return The exchange's instruments
+   */
+  default List<Instrument> getExchangeInstruments() {
+    throw new NotYetImplementedForExchangeException();
+  }
```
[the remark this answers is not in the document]
Yampersand: Implemented in #e3c9159

xchange-core/src/main/java/org/knowm/xchange/service/trade/params/TradeHistoryParamsAll.java
Yampersand: Implemented in #cbfcdf7

Yampersand:
Hey,
I agree that an open position has some aspects which make it seem similar to an order, but I would still argue it shares more similarities with a balance:

- You build up a open position by buying futures contracts, either long or short positions. To build up a position you place actual orders, maker or taker, to buy contracts.

- Those contracts stay in your possession until you sell them, again with actual orders, they expire or you run out of liquidation to back them. In that time they behave like (leveraged) crypto balances as their market price rises and falls.

I hope that makes my point of view more clear.

walec51:
yep, everything is clear now, thank you

PR #3610, opened 2020-07-17
Added bittrexV3

cymp:


cymp:
@timmolter not sure how to solve this, any idea?

PR #3612, opened 2020-07-17
Added bittrexV3 module

cymp:


walec51:
looks ok, I just wanted to ask an question before we merge:

is the v3 module in its current form as feature complete as the old module?

if so then I would be in favor of just replacing the old module with your instead of keeping two modules for bittrex - ppl can always use the old code by just not incrementing the version of their xchange-bittrex dependency

I can help to make this modules error handling comply with best practices from: https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-7
as I see its a little lacking in this regard

cymp:
Please have a look at https://github.com/knowm/XChange/pull/3580#issue-442591688 for an exact answer to your concern.

cymp:
@walec51?

PR #3634, opened 2020-08-01
Fixed missing arguments for bitstamp, kraken and coinmate exchanges in tradehistory api

generalbytes:
Fixed missing arguments for bitstamp, kraken and coinmate exchanges in tradehistory api


timmolter:
I reverted the one dependency that was compiled with JDK 11 a while ago, so that issue should not be present anymore. What exact issue are you facing? Which dependency is the problem?

generalbytes:
jitpack.yml file has been removed - not needed any more.
conflicts with current dev branch resolved.

timmolter:
ERROR] /home/travis/build/knowm/XChange/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampAccountService.java:[181,83] incompatible types: java.lang.Long cannot be converted to java.lang.String

generalbytes:
I noticed. I made mistake during merge to current develop branch. Working on fix now.

xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenTradeService.java
```diff
@@ -68,32 +68,44 @@ public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
...
+      }
+      if (idSpan.getEndId() != null) {
+        end = idSpan.getEndId();
+      }
+    }
+
+    if (params instanceof TradeHistoryParamsTimeSpan) {
+      TradeHistoryParamsTimeSpan timeSpan = (TradeHistoryParamsTimeSpan) params;
+      if (timeSpan.getStartTime() != null) {
+        start = DateUtils.toUnixTimeNullSafe(timeSpan.getStartTime()) + "";
```
walec51, line 100: `+ ""` i unnecesary 

`getKrakenTradeHistory` should support null values

xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampTradeServiceRaw.java
```diff
@@ -20,20 +21,22 @@
   private final BitstampAuthenticated bitstampAuthenticated;
   private final BitstampAuthenticatedV2 bitstampAuthenticatedV2;
   private final BitstampDigest signatureCreator;
-  private final String apiKey;
-  private final SynchronizedValueFactory<Long> nonceFactory;
+  private String apiKey;
+  private SynchronizedValueFactory<Long> nonceFactory;
```
walec51, line 25: why is this not final any more?

xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenTradeService.java
```diff
@@ -68,32 +68,44 @@ public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
...
-    final Long offset;
     if (params instanceof TradeHistoryParamOffset) {
       offset = ((TradeHistoryParamOffset) params).getOffset();
-    } else {
-      offset = null;
+    }
+
+    if (params instanceof TradeHistoryParamsIdSpan) {
+      TradeHistoryParamsIdSpan idSpan = (TradeHistoryParamsIdSpan) params;
+      if (idSpan.getStartId() != null) {
```
walec51, line 89: these `if`s are unnecessary

just doing `start = idSpan.getStartId();` for a null `idSpan.getStartId()` will have the same result in `start` as it is initialized to `null`

walec51 (changes requested):

rpanak-generalbytes:
@timmolter @walec51 I think all the mentioned problems should be fixed now.

rpanak-generalbytes:
> I reverted the one dependency that was compiled with JDK 11 a while ago, so that issue should not be present anymore. What exact issue are you facing? Which dependency is the problem?

There is still some dependency issue, though. This is the JitPack log when building the PR (with merged upstream changes from today): https://jitpack.io/com/github/GENERALBYTESCOM/XChange/b88e82fab6/build.log

It seems like it's a maven plugin problem (fmt-maven-plugin). It depends on google-java-format lib whose published jar is built by JDK14.

walec51 (approved):

milczarekIT:
Here is example json trade from Bitstamp:
`{"fee": "0.00000", "order_id": 1274846773324107, "datetime": "2020-09-14 09:47:45.577000", "usd": 0.0, "btc": "5.61134825", "btc_eur": 8751.00000000, "type": "2", "id": 123193752, "eur": "-49100.00"}`

Bitstamp uses ids as numbers (see `order_id` and `id` properties). 



There are also long types in `org.knowm.xchange.bitstamp.dto.trade.BitstampUserTransaction` (for `id` and `order_id`).

Why in this PR in `org.knowm.xchange.bitstamp.BitstampAuthenticatedV2` types has been changes from `Long` to `String` (@generalbytes , @walec51)?
I know that `org.knowm.xchange.dto.trade.UserTrade` and `org.knowm.xchange.dto.trade.Trade` have `orderId` and `id` with String type, but `org.knowm.xchange.bitstamp.BitstampAuthenticatedV2` is operating on `org.knowm.xchange.bitstamp.dto.trade.BitstampUserTransaction`.

walec51:
@milczarekIT its returned as a string by the API and there is now guarantee in the docs what it will always by an integer

obraz

milczarekIT:
@walec51, I was not writting about integers, but long types.

Presented output from postman is for public api (`/api/v2/transactions/btcusd`). Changes in this PR has been introduced for `/api/v2/user_transactions/`. There are no strings in Bitstamp json for `order_id` and `id`. I am pasting sample json from Bitstamp once again:
```
{"fee": "0.00000", "order_id": 1274846773324107, "datetime": "2020-09-14 09:47:45.577000", "usd": 0.0, "btc": "5.61134825", "btc_eur": 8751.00000000, "type": "2", "id": 123193752, "eur": "-49100.00"}
```

walec51:
@milczarekIT long is just a short name for long integer

are you sure they won't change ids in /api/v2/user_transactions/ to match them up with those in /api/v2/transactions/... ?

they seam to be very loose on this matter, in the response you gave you have `"usd": 0.0` and `"eur": "-49100.00"`

interpreting id fields as string is safer in my opinion and more future proof when we have and exchange API that is very inconsistent in that matter, if one day they start returning UUID (like Bittrex has started to do) then deserialiaizing those ids as Longs will throw an exception

generalbytes:
I suggest to leave order_id as a String. Bitstamp in past used numbers longer than java "long" type allows and it broke multiple implementations. It is better to stay on the safe side with String type.

milczarekIT:
> long is just a short name for long integer

Yes, so integer and short integer can be used interchangeably, but long is a different type. I just wanted to clarify.

> are you sure they won't change

No, I can not (becasue I am not a creator of this API). 

Ok, I just wanted to clarify what was the purpose of change, let's leave id as string. @generalbytes, @walec51 are you going to change `long` ids for other exchanges (to be safe for API change)?

walec51:
@milczarekIT 

when I talk if the id will be always an integer I meant it in the mathematical sends https://en.wikipedia.org/wiki/Integer - stop nit picking which Java type we should use for integer ids - yes `long` is the proper Java type for ids if you are certain that they will be always integer number - but that's not the point of the discussion

I don't think we have to use `String` for ids for all exchange specific DTOs - only for those who do not state in their documentation explicitly that their id is an integer number

PR #3642, opened 2020-08-05
[CoinbasePro-Stream] increasing streaming to full channel and cleanup

earce:
- Adding ability to stream CoinbasePro full channel (l3 orderbook) and requisite code changes to permit for this

- Codebase cleanup (spelling fixes and code improvement)

- Adding `CoinbaseProStreamingService.java` constructor to configure `NettyStreamingService.java` arguments


earce:
@badgerwithagun @mdvx

earce:
@timmolter not sure when you have the next release planned but would love to get this merged before it's cut :)

timmolter:
@earce It won't be for a while, no worries...

earce:
wanted to bump this up and see if anyone has some time to review

badgerwithagun:
On it @earce

badgerwithagun:
@earce if you want the L3 order book, why not just include an option in the arguments to `getOrderBook()`? E.g.

```
streamingMarketDataService.getOrderBook(BTCUSD, FULL);
```
That's what those varargs are intended for.

earce:
CoinbasePro's implementation doesn't lend itself well to this because it uses `CoinbaseProWebSocketSubscriptionMessage` to subscribe to a given channel and upon construction of the `ProductSubscription` and the setting of it in `public void subscribeMultipleCurrencyPairs(ProductSubscription... products)`. Knowing this the main issue is that 'ProductSubscription' is a class which all exchanges use but the 'full' channel is unique to CoinbasePro.

badgerwithagun:
Right, I see.  In that case, I think wanting L3 order books should either:

 - Be a property of `ProductSubscription`
 - _or_ be an `ExchangeSpecification` [exchange specific parameter](https://github.com/knowm/XChange/blob/develop/xchange-core/src/main/java/org/knowm/xchange/ExchangeSpecification.java#L325)

I don't like the feeling of the first choice - it muddies the core API.  The latter seems to be lowest impact: basically just a flag that changes the behaviour of the `ORDER_BOOK` subscription for the connected exchange.  It can be implemented by any other exchanges that want to do the same thing too.

You can add the configuration key constant [here](https://github.com/knowm/XChange/blob/develop/xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingExchange.java#L12).

earce:
@badgerwithagun updated review based on your comments, looks much better

mattdornfeld:
Is there an example of how to subscribe to L3 messages? I only see the ability to get l2 and ticker updates.

earce:
@mattdornfeld you just need to set this ExchangeSpec param `exchangeSpecification.setExchangeSpecificParametersItem(StreamingExchange.L3_ORDERBOOK, true)` and create an instance of `CoinbaseProStreamingService` with the subscriptions you care about (in this case order book). L3 order books are not fully supported with XChange so this will get you the raw messages without any transformation on XChange's behalf.

PR #3657, opened 2020-08-17
[3656] allow overriding nonce factory

zdary:


walec51:
> The original implementation create separate instances of nonce factory for each service. That would mean a possibility of nonce collision.
> 
> I guess I’d be happy with a single instance of nonce factory for all services.

is this not fixed now?

in the current code I see

```
public class BTCMarketsBaseService extends BaseExchangeService implements BaseService {

  ...
  protected final SynchronizedValueFactory<Long> nonceFactory;

  public BTCMarketsBaseService(Exchange exchange) {
    super(exchange);
    ...
    this.nonceFactory = exchange.getNonceFactory();
```

walec51:
I'm closing this assuming it got fixed

PR #3666, opened 2020-08-23
[Streaming] Increasing configuration of various streaming exchanges, small improvements

earce:
Allowed for increased configuration in constructor of some streaming exchanges, made minor code improvements.

### DIFF
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
@@ -101,7 +101,8 @@ public class CoinbaseProAdapters {
   }
 
   public static CurrencyPair toCurrencyPair(final String productId) {
-    return new CurrencyPair(productId.replace('-', '/'));
+    final String[] parts = productId.split("-");
+    return new CurrencyPair(parts[0], parts[1]);
   }
 
   public static Ticker adaptTicker(
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
@@ -50,7 +50,7 @@ import org.slf4j.LoggerFactory;
 
 public class CoinbaseProAdapters {
 
-  private static Logger logger = LoggerFactory.getLogger(CoinbaseProAdapters.class);
+  private static final Logger logger = LoggerFactory.getLogger(CoinbaseProAdapters.class);
 
   private CoinbaseProAdapters() {}
 
@@ -98,6 +98,10 @@ public class CoinbaseProAdapters {
       logger.warn("unable to parse rawDate={} modified={}", rawDate, modified, e);
       return null;
     }
+  }
+
+  public static CurrencyPair toCurrencyPair(final String productId) {
+    return new CurrencyPair(productId.replace('-', '/'));
   }
 
   public static Ticker adaptTicker(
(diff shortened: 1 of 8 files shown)


earce:
@badgerwithagun another easy one for you :)

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
```diff
@@ -100,6 +100,10 @@ protected static Date parseDate(final String rawDate) {
     }
   }
 
+  public static CurrencyPair toCurrencyPair(final String productId) {
+    return new CurrencyPair(productId.replace('-', '/'));
```
badgerwithagun, line 104: nit: methods like this are called a _lot_ in high-volume code. This would be more efficient:

```
var split = productId.split('-');
return new CurrencyPair(split[0], split[1]);
```

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
```diff
@@ -29,7 +29,7 @@
 
   public KrakenStreamingExchange() {}
 
-  private static String pickUri(boolean isPrivate, boolean useBeta) {
+  public static String pickUri(boolean isPrivate, boolean useBeta) {
```
badgerwithagun, line 32: Why make this public?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
```diff
@@ -143,7 +143,7 @@ public KrakenStreamingMarketDataService(KrakenStreamingService service) {
             });
   }
 
-  private String getChannelName(
+  public String getChannelName(
```
badgerwithagun, line 146: Or this?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingExchange.java
earce: @badgerwithagun I use these urls directly in my own code and would like to expose them.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
```diff
@@ -14,6 +15,16 @@ public BinanceStreamingService(String baseUri, ProductSubscription productSubscr
     this.productSubscription = productSubscription;
   }
 
+  public BinanceStreamingService(String baseUri,
```
badgerwithagun, line 18: You've added these constructor overloads but no code that uses them. Are you planning to call them on construction in the `StreamingExchange` implementations in some follow-up PR?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
earce: same reason, I use these functions directly and would like a handle on them and receive updates for them if the underlying code changes.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
earce: I don't have immediate plans for them, I think being restricted to use the defaults is not ideal and I use these outside of this codebase (I think it makes sense to have them in place)

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
earce: updated

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
badgerwithagun: I'd be intrigued to hear how and why you are using the underlying streaming services directly (rather than just exposing nice entry points in the `StreamingExchange`s, but fair enough for now.

earce: @badgerwithagun in short I want raw json payloads, I'm not interested in the unified objects Xchange uses so I have found a few ways of accomplishing this and using the mature parts of Xchange's code.

earce: thanks for approving

PR #3681, opened 2020-09-02
Add kraken price scale

CayenneLow:


nielsdraaisma:
@timmolter just bringing to you attention that after this change this client now hits the kraken metadata API for each placeOrder request, i don't think this change is necesary at all if the remoteInit method is called properly?.

timmolter:
Oops, yeah good call. I'll try to revert it now...

matejsp:
We are also getting NPEs with latest 5.0.2.

```
Caused by: java.lang.NullPointerException: null
	at org.knowm.xchange.kraken.service.KrakenBaseService.getAssetPairScale(KrakenBaseService.java:172)
	at org.knowm.xchange.kraken.service.KrakenTradeServiceRaw.placeKrakenLimitOrder(KrakenTradeServiceRaw.java:212)
```

Why is the responsibility of the API to do the rounding? And rounding HALF_UP is wrong for at least one of bid or ask side of orderbook. This should be up to the calling party to get the correct price.

PR #3690, opened 2020-09-09
Coinbase Pro Funding Type addition

alwinc:
Coinbase pro require type as a parameter in the trade history parameters


timmolter:
closing because another PR covered this

Issue #3692, opened 2020-09-09
[Bitstamp] Bitstamp AccountService.getFundingHistory doesn't handle "subAccountTransfer" type user transactions.


mathvdh:
Hi all,

If you are trying to use the `AccountService.getFundingHistory` function it doesn't handle the user transactions which have a type 14 = `subAccountTransfer`.
The transactions are returned by the API but then just discarded by the function from Adapters : `adaptFundingHistory`.
I have a fix for it and it will be sent soon in a subsequent PR.

BTW thanks for this library it's awesome!

Cheers.

PR #3693, opened 2020-09-09
Fix for issue 3692

mathvdh:
Hi all.

This is a quite simple fix for issue : https://github.com/knowm/XChange/issues/3692

Basically I'm modifying the `BitstampUserTransaction` to add one function and then `BitstampAdapters` to handle this case.

I tested it locally (because I have Bitstamp subaccounts, which had sub account transfers and they were not showing up before the fix and are now showing up.

Best


mathvdh:
I have added a second commit, because the "type" of the FundingRecord returned by `getFundingHistory` was reported incorrectly for SubAccountTransfer types transactions. It's now fixed.

PR #3694, opened 2020-09-09
Poloniex - streaming account notifications & Kraken - adding/cancelling orders via websockets

marcinrabiej:


xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -102,6 +113,16 @@ protected void handleMessage(JsonNode message) {
                     "Channel {} has been failed: {}", channelName, statusMessage.getErrorMessage());
             }
             break;
+          case addOrderStatus:
```
earce, line 116: It may make more sense for these type of websocket transactions to keep as much of their logic within the service they reside in as possible. In this case keeping as much of the streaming trade logic inside `KrakenStreamingTradeService` so as to keep purpose of `KrakenStreamingService` clear and the footprint smaller.

I am thinking possibly use the `KrakenStreamingService` like a router where when raw messages that should be processed by say `KrakenStreamingTradeService` are send to that service by this switch statement so that `KrakenStreamingService`'s job really just becomes receiving messages to send and routing the result to the handling class.

@badgerwithagun usually has good ideas regarding implementation, not sure what you think?

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingService.java
```diff
@@ -75,6 +75,27 @@ public boolean processArrayMassageSeparately() {
...
+
+    // Parse incoming message to JSON
+    try {
+      jsonNode = objectMapper.readTree(message);
+    } catch (IOException e) {
+      LOG.error("Error parsing incoming message to JSON: {}", message);
+      return;
+    }
+
+    if (jsonNode.isArray() && jsonNode.size() < 3) {
```
earce, line 91: for these two equality comparisons can you switch the constant to the left side for NPE protetction?

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingService.java
```diff
@@ -124,21 +150,35 @@ protected String getChannelNameFromMessage(JsonNode message) {
 
   @Override
   public String getSubscribeMessage(String channelName, Object... args) throws IOException {
-    PoloniexWebSocketSubscriptionMessage subscribeMessage =
-        new PoloniexWebSocketSubscriptionMessage("subscribe", channelName);
+    Object subscribeMessage;
+    if (ACCOUNT_NOTIFICATIONS_CHANNEL.equals(channelName)) {
+      subscribeMessage =
+          getAccountNotificationsSubscription(
+              (String) args[0], (String) args[1], (SynchronizedValueFactory<Long>) args[2]);
```
earce, line 157: For these args casts, probably makes sense to check the array is the size you expect and each arg is of the type you expect and log when one of those checks is incorrect. This will make debugging incorrect params much simpler, `CoinbaseProMarketDataService.getOrderbook()` as an example.

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingUserDataService.java
```diff
@@ -0,0 +1,85 @@
...
+
+  public PoloniexStreamingUserDataService(PoloniexStreamingService service) {
+    this.service = service;
+    sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+    sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
+  }
+
+  public Observable<UserTrade> getUserTrades() {
+    return getStreamEvents()
+        .filter(s -> s.getEventType().equals("t"))
```
earce, line 39: `t` should be on the left side for NPE protection

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketAccountNotificationsTransaction.java
```diff
@@ -0,0 +1,78 @@
...
+                  new BigDecimal(jsonNode.get(5).asText()),
+                  jsonNode.get(6).asText());
+          events.add(new PoloniexWebSocketNewLimitOrderEvent(newLimitOrderEvent));
+          break;
+        default:
+          break;
+      }
+    }
+
+    return events.toArray(new PoloniexWebSocketEvent[events.size()]);
```
earce, line 68: leaving this array initialization with size 0 is more efficient for Java

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
marcinrabiej: I agree that the Kraken trading logic should reside in KrakenStreamingTradeService and KrakenStreamingService should serve generic Kraken message handling - and this is how this is implemented. The logic that you see int the KrakenStreamingService is for queueing the messages/dispatching the statuses to proper requests and I believe this part should stay there, while generating the messages and actually handling them is in the KrakenStreamingTradeService

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingService.java
marcinrabiej: fixed

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingService.java
marcinrabiej: added validation

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingService.java
```diff
@@ -124,21 +150,35 @@ protected String getChannelNameFromMessage(JsonNode message) {
...
 
+  private PoloniexWebSocketAccountNotificationsSubscriptionMessage
+      getAccountNotificationsSubscription(
+          String apiKey, String secretKey, SynchronizedValueFactory<Long> nonceFactory) {
+    String nonce = "nonce=" + nonceFactory.createValue();
+    String signature = new PoloniexSigner(secretKey).getSignature(nonce.getBytes());
+    PoloniexWebSocketAccountNotificationsSubscriptionMessage subscriptionMessage =
+        new PoloniexWebSocketAccountNotificationsSubscriptionMessage(
+            "subscribe", ACCOUNT_NOTIFICATIONS_CHANNEL, apiKey, nonce, signature);
+    LOG.info("Subscribing for account notifications: {}", subscriptionMessage);
```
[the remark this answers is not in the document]
marcinrabiej: removed the log statement - not needed at all

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/PoloniexStreamingUserDataService.java
marcinrabiej: fixed

xchange-stream-poloniex2/src/main/java/info/bitrich/xchangestream/poloniex2/dto/PoloniexWebSocketAccountNotificationsTransaction.java
marcinrabiej: fixed

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -35,6 +41,11 @@
   private final boolean isPrivate;
 
   private final Map<Integer, String> subscriptionRequestMap = new ConcurrentHashMap<>();
+  private final Map<Integer, ObservableEmitter<KrakenAddOrderStatusMessage>>
+          addOrderStatusEmitters = new HashMap<>();
+
+  private final Map<Integer, ObservableEmitter<KrakenCancelOrderStatusMessage>>
+          cancelOrderStatusEmitters = new HashMap<>();
```
earce, line 48: just noticed this, I believe the NettyEventLoop is run with multiple (2) threads which would imply this map should be threadsafe I think these should be concurrent hashmaps.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -224,6 +245,88 @@ protected WebSocketClientHandler getWebSocketClientHandler(
 
   private WebSocketClientHandler.WebSocketMessageHandler channelInactiveHandler = null;
 
+  String submitLimitOrder(KrakenStreamingLimitOrder krakenStreamingLimitOrder) {
```
earce, line 248: sorry missed these the first time through but the timeouts how did you decide on 60s for sending a limit order and 1s for canceling? Should the latter be a little longer?

earce, line 248: also, I see what you have effectively done here is make a non-blocking operation blocking, would it make more sense to have this method return the observable and let the caller deal with the async behavior, it may not, but if a user wants a blocking operation I would imagine they could just use REST?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
earce: sounds good I will defer to @badgerwithagun for this

earce (changes requested):

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
marcinrabiej: Kraken has some issues and sometimes you don't get any response for quite a long time. In case this is about placing a new order - no drama - at least in my case. In case it is about cancelling an order - I want to get the control quickly back to issue cancel via REST.

marcinrabiej: Well, the difference between sending the command via REST and via Websocket is that.... it is sent via Websocket. Just different channel, much faster though usually. The rest is exactly the same as in REST. Command sent, I want the response back I don't do anything until I get the response for my command therefore this is blocking.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
marcinrabiej: Fixed

timmolter:
Can this now be merged?

earce:
I would still like @badgerwithagun to take a look if possible as this will see a precedent for streaming based transactions for the library.

marcinrabiej:
> I would still like @badgerwithagun to take a look if possible as this will set a precedent for streaming based transactions for the library.

@earce are you worried about this blocking behaviour? Or is there any other thing that still worries you? Would exposing the Observable alongside the blocking behaviour solve this?

earce:
A few things. 

1.  First off yes the blocking behavior because I know @badgerwithagun was trying to move to a non-blocking approach for a lot of the streaming based functionality. I think for an async based library, implementing naturally async functions as blocking defeats the purpose. 

2. A lot of code is coming into the `KrakenStreamingService.java` which I am not totally sure that it should live here, the Exchange classes have natural separation of duties for different methods and avoid the monolithic classes. Additionally the cancel and add methods I would imagine to follow the XChange setup would be methods that are universal and other streaming services can implement. Effectively using inheritance to override and having a universal object for each type of transaction. I don't really see any of that here as well. 

I am not opposed to the changes but I am not as familiar with these design choices to make the call on these. I am simply raising them as considerations.

earce:
@badgerwithagun know you're busy but wanted to bump this up

Issue #3683, opened 2020-09-03
[Kraken-Streaming] websocket connection dead if Kraken returns invalid session message


earce:
When Kraken returns `ESession:Invalid session` the `KrakenStreamingService.java` class is unable to recover the websocket connectivity. Kraken describes what creates this error message on their website [Why am I getting Session Invalid](https://support.kraken.com/hc/en-us/articles/202966366-Why-am-I-getting-a-Session-Invalid-message-) and have answered this on reddit as well [Session Invalid](https://www.reddit.com/r/Kraken/comments/1z2kh3/session_invalid_immediately_after_first_login/).

This is the second time in about 2 months I've seen this present itself. Log line goes as follows `ERROR info.bitrich.xchangestream.kraken.KrakenStreamingService - Channel null has been failed: ESession:Invalid session`. I was thinking about throwing inside of the `case error:` to have it bubble back up and trigger a reconnect inside the `NettyStreamingService.java`. Not sure if this is the best approach or if this will help with a valid session.

Would love some input @badgerwithagun @mdvx.

mdvx:
Is the problem that the error is occuring when a new IP might have been assigned to the client, or is the problem that the error is not being handled correctly?

For my part, i do all my reconnects outside of knowm, because I maintain state on some structures, and silent reconnects can mess those up (unknown missing data), and also add undetected latency.

earce:
This is happening on AWS based deployment, I am not under the impression EC2 ips are ever reassigned if there is not host restart etc, in a normal running environment I don't think ips will ever change there.

Given that my initial suspicion was that this was a transient issue? Restarting the process seemed to fix it.

How do you do these reconnects outside of knowm? Like are you not using knowm's re-connection logic?

earce:
Does anyone have any other thoughts about this?

PR #3697, opened 2020-09-09
[Kraken-Streaming] fixing invalid session issues 

earce:
From an email I exchanged with Kraken support:

> Thank you for contacting us regarding the Kraken WebSocket API. In theory WebSocket authentication tokens can last indefinitely, but in reality they do sometimes expire causing an invalid session error. As an example, during a recent WebSocket API upgrade, many authentication tokens became invalid (for no apparent reason to the token owners), causing unexpected invalid session errors upon reconnecting/resubscribing after the upgrade. The solution to the invalid session error is simply to request a new authentication token via the REST API GetWebSocketsToken endpoint, and then use the new token for all new connections/subscriptions.

Based on this a few points come up that I attempt to address in this review:

1. In `KrakenStreamingTradeService.java` `subscribeChannel(channelName, renewToken().getToken())` is called with a token upon subscription so the instance of the token called here is reused for the duration of this process run (this sounds reasonable given that once a token is used to subscribe to a private feed a token will remain valid as long as a subscription remains active) [Kraken Documentation](https://support.kraken.com/hc/en-us/articles/360034664311-How-to-subscribe-to-the-Kraken-WebSocket-private-feeds). I see the code has `renewToken()` logic but given the fact that the token instance is passed directly into the subscription I am under the impression that this is never called during reconnects.

2. The scenario where the token becomes invalid due to something on their side and given our logic for token renewal we attempt to subscribe with a bad token before we our token lifespan has us consider the token invalid. This will cause a Session Invalid on our end and we won't attempt reconnect, rendering the websocket channel dead.

Edit:

3. I added an check for session invalid message coming back from Kraken which would indicate despite our efforts we still sent an invalid token, which would protect against the extremely unlikely situation where between us requesting a token and the submission the token became invalid. This removes all race conditions from our code.

Addresses what I investigated following https://github.com/knowm/XChange/issues/3683

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,12 +143,12 @@ Add the following dependencies in your pom.xml file. You will need at least xcha
 <dependency>
   <groupId>org.knowm.xchange</groupId>
   <artifactId>xchange-core</artifactId>
-  <version>5.0.1</version>
+  <version>5.0.2</version>
 </dependency>
 <dependency>
   <groupId>org.knowm.xchange</groupId>
   <artifactId>xchange-XYZ</artifactId>
-  <version>5.0.1</version>
+  <version>5.0.2</version>
 </dependency>
 ```
 
@@ -158,7 +158,7 @@ If it is available for your exchange, you may also want to use the streaming API
 <dependency>
   <groupId>org.knowm.xchange</groupId>
   <artifactId>xchange-stream-XYZ</artifactId>
-  <version>5.0.1</version>
+  <version>5.0.2</version>
 </dependency>
 ```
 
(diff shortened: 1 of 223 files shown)


earce:
@mdvx 
@badgerwithagun

earce:
hi wanted to bump this back up, was hoping to get it looked at

earce:
@timmolter how would would you feel about looking at this review?

badgerwithagun:
I'll look at this today. Been busy with work, sorry for the delay.

badgerwithagun (approved):

panieldark:
Nice job! @earce

PR #3704, opened 2020-09-12
[CoinbasePro] CoinbaseProException - return original exchange message for getMessage()

milczarekIT:
CoinbaseProException does not return original exchange message for `getMessage()`. It returns default `HttpStatusExceptionSupport.getMessage()`.
So `CoinbaseProException` currently returns:
`Invalid API Key  (HTTP status code: 401)`
instead of
`Invalid API Key`.

If we need to obtain http status code we can use: `HttpStatusExceptionSupport.getHttpStatusCode()`

This PR is fixing it.

### DIFF
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/dto/CoinbaseProException.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/dto/CoinbaseProException.java
@@ -5,7 +5,15 @@ import si.mazi.rescu.HttpStatusExceptionSupport;
 
 public class CoinbaseProException extends HttpStatusExceptionSupport {
 
-  public CoinbaseProException(@JsonProperty("message") String reason) {
-    super(reason);
+  private final String message;
+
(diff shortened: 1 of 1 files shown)


earce:
is the status code not present in the headers? if so you may just be able to implement HttpResponseAware and extract it from the headers similar to `BinanceException`

milczarekIT:
@earce  http status code is present in `CoinbaseProException`, because it extands `HttpStatusExceptionSupport`.

In this PR I am not adding 'HTTP status code' to message, I am removing it from default impl of `HttpStatusExceptionSupport.getMessage()`.
If someone needs status code it can be obtained it `HttpStatusExceptionSupport.getHttpStatusCode()`.

The purpose of this PR is to return original message from exchange not the default presented below:
```
@Override
    public String getMessage() {
        return String.format("%s (HTTP status code: %d)", super.getMessage(), __httpStatusCode);
    }
```

earce:
@milczarekIT got it, this looks good then, this may however may have adverse effects if someone is depending on that format of the message but I think your change is the right thing to do

PR #3710, opened 2020-09-16
[Bitstamp] Fix: getUserTransactions use correct method for currency pair null

milczarekIT:
[Bitstamp] Fix: getUserTransactions - use correct method for currency pair null

### DIFF
--- a/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampTradeService.java
+++ b/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampTradeService.java
@@ -24,7 +24,16 @@ import org.knowm.xchange.dto.trade.OpenOrders;
 import org.knowm.xchange.dto.trade.UserTrades;
 import org.knowm.xchange.exceptions.ExchangeException;
 import org.knowm.xchange.service.trade.TradeService;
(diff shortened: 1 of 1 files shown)


xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampTradeService.java
```diff
@@ -146,14 +155,22 @@ public UserTrades getTradeHistory(TradeHistoryParams params) throws IOException
...
-            currencyPair,
-            offset,
-            sort == null ? null : sort.toString(),
-            sinceTimestamp,
-            sinceId);
+    BitstampUserTransaction[] txs;
+    if (currencyPair == null) {
+      txs =
+          getBitstampUserTransactions(
+              limit, offset, sort == null ? null : sort.toString(), sinceTimestamp, sinceId);
```
walec51, line 162: just one small remark

prepare a `String sortStr = sort == null ? null : sort.toString();` before the if

milczarekIT: Sure, done.

walec51: thanks

PR #3718, opened 2020-09-17
[Bitmex-Stream] Relax restriction that currency string length must be >= 6

Dynamikey:
The current implementation assumes that the currency string is always at least 6 characters. However, some symbols are less than 6 characters e.g. https://www.bitmex.com/app/index/.BXBT


Dynamikey:
@earce @badgerwithagun @timmolter 
What's the workflow for approving PRs? thanks

earce:
Question, how will this work if both sides are > 3 in size?

badgerwithagun:
Bitmex only. Sure it'll be fine.

earce:
Even for something like LINK/USDT?

Dynamikey:
The Bitmex symbol for that is LINKUSDTZ20. That will be parsed by the new implementation without a problem. The problem arises when the symbol length is less than 6 as substring() throws an error.

Dynamikey:
Ah oops, should this have been pushed to knowm:develop?

earce:
@Dynamikey this has been merged you should be good to go on the next cut at the end of the month.

PR #3721, opened 2020-09-23
[Streaming-Core] code cleanup and improvement

earce:
- Removing unnecessarily overridden methods
- Fixed spelling
- Improving thread-safety in `PoloniexStreamingService.java`, `BitfinexStreamingService.java`, `BitmexStreamingService.java`, `PoloniexStreamingService.java`
- NPE protections
- Removed pusher code from library


earce:
@mdvx @badgerwithagun can you take a look when you get some time

@badgerwithagun the rate-controller was inspired by your class but I experienced some rate limiting that was not managed by XChange well and the library is needing of some form of rate control. Would love to get this looked at/modified/committed before 5.0.3 :)

badgerwithagun:
Hm, one of the reasons I'd not got any further with it is because I thought it better to line up with the existing resilience support using resilience4j that now exists in core XChange.

earce:
I looked at some other rate limiters but not resilience4j, and I guess I wanted to keep the rate limiting simple and without other deps. Is there something particularly valuable about resilience4j?

Do you want me to give using resilience4j for this a go?

walec51:
lets not reinvent the wheel and use and existing solution where possible

rate limiting look simple at first but becomes more complex when you get into the edge cases

take a look at:

https://github.com/knowm/XChange/blob/develop/xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeServiceRaw.java#L52

https://github.com/knowm/XChange/blob/9ef8de2f3e08a13588593ca5b3f87335892fa942/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceResilience.java#L7

plus resiliance4j also has a retry mechanism that integrates well with its rate limiter

resilience4j is also highly configurable and in the way it is used in normal XChange modules that configuraction can be overriten by our clients:

https://github.com/knowm/XChange/wiki/Implementing-resiliency-features

it would be a very bad direction for stream to have a totally different rate limit / retry mechanism then the normal libraries

walec51 (changes requested):
because of the above we should use resiliance4j in a similar manner as it is done in normal Xchange modules

earce:
@walec51 taking a look at this it appears that this library manages the request rate on the client side by decorating our methods with the resilience code. These rate limiters control the message rate based on a cycle and not exceeding the request count per cycle (at least how I understand it).

I am not quite capturing how this could be used to handle the case if a server throws back a 429 exception via a websocket channel and we either need to throttle the requests or halt all requests for a period of time. I am trying to react to a 429 event and basically turn on rate throttling or halting for a period of time after. The RateLimiter doesn't seem to be intended for a temporary throttle/halt of messages.

Something else I was curious about the way the library is setup now if we get back a 429 on a decorated function and the function had a retry on it we would treat that 429 as an error and trigger a retry after the configured retry interval correct? This doesn't feel correct, the retry interval and the amount of time we should wait after 429 will not/should not always be the same I imagine.

Let me know what you think.

walec51:
@earce we should avoid causing an 429 not react when it happens

In a multi threaded application if you start reacting after receiving an 429 error it may be to late as there may be already a dozen of requests initialed simultaneously and they will all fail before you event star to block them. Generate to much 429 errors and you get z temporary ban.

As for retrys in resiliance4j you can define:
- for which exceptions retrys should take place
- if a given retry loop should be under or above a rate limiter (for ex. retrys on ConncetException should no go through a rate limiters counter but other exception like a read timeout should)

earce:
@walec51 I agree to a certain extent but you absolutely need to handle both preventing 429s and reacting if and when they happen, expecting you will never have a 429 and building a core lib this way I don't think is prudent. Unintended setups, multiple processes on the same host, edge cases etc may bring you 429s and if they do you application should handle them accordingly and not be bound to the next release to be able to fix the issue. 

Having rate limiting which enforces a 15, 30 second wait etc could be what prevents your application from getting banned outright.

Also for certain code setups(mine included) where I may want to send messages via websocket with my own functions instead of Xchange functions, even if the right setup would be to wrap your own functions in resilience logic, I think the library itself should offer the protection if a 429 is triggered.

Let me know what you think.

walec51:
@earce agreed, both aspects are important, we should both prevent 429 and react if it happens

I have experience in contributing code to resiliance4j. I could add a configuration option to their rate limiter so that it will zero out the number of available permits in the current period if a given condition is met. From our side it would look like this:

```
public final class BinanceResilience {

  // ...

  public static ResilienceRegistries createRegistries() {
    ResilienceRegistries registries = new ResilienceRegistries();
    registries
        .rateLimiters()
        .rateLimiter(
            REQUEST_WEIGHT_RATE_LIMITER,
            RateLimiterConfig.from(registries.rateLimiters().getDefaultConfig())
                .limitRefreshPeriod(Duration.ofMinutes(1))
                .limitForPeriod(1200)
                // this is the new configuration option I'm talking about
                .closePeriodOnException(ClientExceptionContitions.statusCodeEquals(429))
                .build());
```  

if we agree that this solution is ok for both of us and would allow us to usage resiliance4j for streaming APIs then I can prepare a PR to resiliance4j during this weekend.

walec51:
@earce added a feature request to resiliance4j: https://github.com/resilience4j/resilience4j/issues/1228

please give it a like :) 

as for where to set up the limiter - I'm not familiar with our streaming API because I don't use it but

in normal xchange APIs resiliance4j components are always:
- configured in dedicated class, like for example `BinanceResilience`
- and applied in `*ServiceRaw` classes

this is because rate limiting, retry polices, etc. are very exchange specific - take a look at `xchange-binance` for a good example

RobWin:
Hi, I'm the creator of Resilience4j. I think I understood your feature request better after reading the comments.
Just out of curiosity, do you use CircuitBreaker do protect the remove services? You could also use a CircuitBreaker to stop further request for a certain period of time after a HTTP 429 has been catched.
But is just a workaround until the final solution is available.

earce:
@RobWin we do not currently. I think we will probably wait until the changes we want to make to resilience are in place to make the changes to the core lib.

earce:
@walec51 to keep it simple I have taken out all the rate limiting logic from this review and kept just code cleanup, improvement (things unrelated to the resilience code). Can you take a look and when it's merged I will open up another review once the ratelimiting code is ready from resilience?

earce:
@timmolter have removed pusher code from library

walec51:
@earce ok, merging

I should have a PR to resiliance4j by the end of this weekend

walec51 (approved):

Issue #3672, opened 2020-08-25
[Bittrex] Example for v3


appyhour770:
End of life for the v1 api is planned for 9/30/2020 and from this date, only v3 calls will be supported.
=> It will be ideal to have the bittrex-example covering v3 specific use/calls.

matejsp:
PR: https://github.com/knowm/XChange/pull/3722

PR #3722, opened 2020-09-23
[Bittrex v3] migrate examples to bittrex v3

matejsp:
migrate examples to bittrex v3 (closing #3672)

### DIFF
--- a/xchange-examples/pom.xml
+++ b/xchange-examples/pom.xml
@@ -77,7 +77,7 @@
 		<!-- XChange Bittrex exchange support -->
 		<dependency>
 			<groupId>${project.groupId}</groupId>
-			<artifactId>xchange-bittrex</artifactId>
+			<artifactId>xchange-bittrexV3</artifactId>
 			<version>${project.version}</version>
 		</dependency>
 		<!-- XChange CampBX exchange support -->
--- a/xchange-examples/src/main/java/org/knowm/xchange/examples/bittrex/account/BittrexAccountDemo.java
+++ b/xchange-examples/src/main/java/org/knowm/xchange/examples/bittrex/account/BittrexAccountDemo.java
@@ -1,9 +1,11 @@
 package org.knowm.xchange.examples.bittrex.account;
 
 import java.io.IOException;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import org.knowm.xchange.Exchange;
+import org.knowm.xchange.bittrex.dto.account.BittrexAddress;
 import org.knowm.xchange.bittrex.dto.account.BittrexBalance;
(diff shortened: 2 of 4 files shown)


appyhour770:
Thanks @matejsp for your pull request.
I noticed that the artifact "xchange-bittrexV3" can not be found in maven repositories.
Is it normal ?

matejsp:
Hm ... But it should be used from the parent (aka snapshot version)
https://mvnrepository.com/artifact/org.knowm.xchange/xchange-bittrexV3

matejsp:
[INFO] Building XChange BittrexV3 5.0.3-SNAPSHOT

But the issue is with:
[ERROR] Failed to execute goal on project xchange-btcmarkets: Could not resolve dependencies for project org.knowm.xchange:xchange-btcmarkets:jar:5.0.3-SNAPSHOT: Could not transfer artifact org.powermock:powermock-module-junit4-common:jar:2.0.7 from/to sonatype-oss-public (https://oss.sonatype.org/content/groups/public/): Connection reset -> [Help 1]

https://oss.sonatype.org/content/groups/public/org/powermock/powermock-module-junit4-common/2.0.7/

Seems like just a temp build error.

appyhour770:
Exact and thanks a lot for the support 👍

walec51:
module xchange-bittrexV3 will become the new module xchange-bittrex when its finished so I think that migrating examples now is unnecessary

matejsp:
This patch is exactly that ... Removing xchange-bittrex and migrating dependency to xchange-bittrexV3..

walec51:
you didn't get the point, in the future the current xchange-bittrex should be removed and xchange-bittrexV3 should be renamed to xchange-bittrex

matejsp:
So between now and some time in the future you would prefer wrong sample based on already obsolete v1. And do the same java changes in the future anyway.

Than the alternative to just migrate the sample to v3 (pom + java changes as well) with simple pom change and then in future in do the same rename (back) in pom.xml as well. It will be one line change.

Difference is that users can have a valid sample now ... or they can use a non working and already obsolete one, without wasting two changed lines of code in git ...

walec51:
if xchange-bittrex module is non working and already obsolete then why are we keeping it?

@cymp can you confirm that? 

if xchange-bittrex is unusable then its about time we removed it and renamed xchange-bittrexV3 to xchange-bittrex

cymp:
xchange-bittrex is still usable with v1.1 api, they announced that it will be unplugged on setpember 30th, but it will actually still be usable for at least a month. But on the other hand the v3 now fully works. It really depends on if we want to have the v1.1 be usable with latest xchange version, or leave the legacy users of v1.1 with the solution of using an old version of xchange.

PR #3726, opened 2020-09-26
[Deribit] Update DeribitTradeServiceRaw.java; added edit() for path: /private/edit

Ferdi-135:
Note: First time proposal.
The /private/edit works fine if you don't add the MMP Flag to the Parameters.
I added the other Params only as comments, because I don't need them for my work.
Tested it, works. Much easier than deleting the existing order and placing a new one.
Best regards


timmolter:
Can you see why the unit test is failing?

PR #3728, opened 2020-09-27
[Bitmex-Stream] Relax restriction that currency string must be >= 6 (#1)

Dynamikey:
Raising a PR to get this into the snapshot jar/develop branch. https://github.com/knowm/XChange/pull/3718


Dynamikey:
@badgerwithagun 
@earce

Issue #3729, opened 2020-09-28
Add BRZ on Currency and CurrencyPair class


mileschet:
Hi, 

I wanted to add the brazilian digital token that's a stable coin listed on some exchanges the project support, i've working on some integrations with Bittrex and it would be nice to have on the core project.

Thank you,

walec51:
just do `new Currency("ANY_OTHER_COIN", "BRZ")` in your project

mileschet:
Well the thing is that some exchanges are listing and i have some projects using the lib, including some implementation of connectors using it. do you still think doesn't worth to have on your parent project so i dont have to maintain my own fork ?

PR #3730, opened 2020-09-28
Add BRZ on Currency and CurrencyPair class #3729

mileschet:


walec51:
just do `new Currency("ANY_OTHER_COIN", "BRZ")` in your project

mileschet:
Hi, thank you for the clarification. I have some apps using this library, i really like the library and want to collaborate more, i have 2 "adapters" for exchanges that have not yet being implemented on your lib, both of then have BRZ listed, so it would be nice to me to not have to maintain my own fork of your parent, what can i do to proceed with it ?

mileschet:
Thank you!

Issue #3272, opened 2019-10-23
Passing through ResCU Interceptor's


Agrajag0:
Sometimes it would be nice to have the opportunity to test the api with real exchange data without the need to connect to the internet or to save the complete request/response data. 

For these use cases the ResCU client offers the so called Interceptor's. At the moment they are ignored when creating the REST Proxy, because they are optional. The question is, how to pass through interceptors if needed?

One solution that come to my mind is to add a method `Interceptor[] getInterceptors()` into the **BaseExchangeService.class**. Here you can read out some exchange specific parameter with the interceptors to pass through. It can then be used in the concrete service classes when creating the Proxy.

`RestProxyFactory.createProxy(restInterface, "some url", getClientConfig(), getInterceptors());
`

One drawback of this approach is, that you have to change a lot of sources and every new service has to use this pattern. Nethertheless i'am willing to make this effort if this approach is acceptable. Any opinions?

walec51:
You can accomplish this by just using wiremock and setting exchangeSpecification.setSslUri to localhost. Its a better test the using interceptors which are meant to decorate requests, not mock them entirely.

Agrajag0:
Well yes, that is an argument. Although it's a more heavier process with an extra server process involved. But okay.

What is with saving of request/response data. How would you accomplish that? Any suggestions appreciated.

milczarekIT:
@Agrajag0, if you are still interested in ResCU Interceptors, I added interceptors support for XChange in https://github.com/knowm/XChange/pull/3745

PR #3745, opened 2020-10-05
[core] Added rescu Interceptor(s) support

milczarekIT:
Added rescu Interceptor(s) support - added the possibility to use custom Interceptors.

Relates to #3272

Added rescu Interceptor(s) support to XChange library.
Based on Java ServiceLoader mechanism, so Interceptors can be easily added in client app by adding content to file:
```META-INF/services/si.mazi.rescu.Interceptor```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 *.iml
 *.ipr
 *.ips
-.idea/
 
 # Eclipse
 .settings/
@@ -24,7 +23,6 @@ javadoc/
 bin/
 log/
 classes/
-META-INF/
 
 # Misc.
 .DS_Store
--- a/xchange-core/src/main/java/org/knowm/xchange/client/ExchangeRestProxyBuilder.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/client/ExchangeRestProxyBuilder.java
@@ -1,9 +1,11 @@
 package org.knowm.xchange.client;
 
 import java.util.ArrayList;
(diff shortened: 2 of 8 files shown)


earce:
Is this something worth adding an example to the wiki/documentation for future reference?

milczarekIT:
@earce  I am going to add an example to wiki, after PR will be merged to `develop`

timmolter:
I have a few questions about this.

1. What are you planning on using this for?
1. Wouldn't this allow for a general tool for all exchange implementations for better automating collecting raw JSON responses, header values and possible errors compared to the "old way" of using logback as described on the wiki here: <https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices>?

For reference, here's the mmazi docs on Interceptors: <https://github.com/mmazi/rescu/wiki/Interceptors>

milczarekIT:
1. It can be used for various purposes. E.g. to record and replay HTTP requests and responses to create API mocks, for auditing purposes, custom logging requests and responses (because logging in si.mazi.rescu.HttpTemplate is not configurable, and you can log everything or nothing), to customize payload or headers - a lot of use cases for interceptors.
2. Sure, you can write an interceptor which collect some data and store it in the preferred format.

After https://github.com/knowm/XChange/pull/3703 all exchanges are build in the same way, so this PR is adding the possibility to use interceptors for each exchange, and it can be configured very easy.

PR #3761, opened 2020-10-12
Fix NPE, maxDepth default to zero instead of null when args is empty

dazito:
Fix a null pointer exception in `GeminiStreamingMarketDataService` class, method `getOrderBook` caused by when `Object... args` is empty it will set `maxDepth` to null which would throw a NPE when we call `GeminiAdaptersX.toOrderbook(geminiOrderbook, maxDepth, new Date())` at the return for `getOrderBook`.

`toOrderBook` method signature is: `public static OrderBook toOrderbook(GeminiOrderbook geminiOrderbook, int maxLevels, Date timestamp)`

`maxLevels` is a primitive int, passing in a null here throws a NPE.

In the fix, instead of setting `maxDepth` to null we set it to zero.

### DIFF
--- a/xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
+++ b/xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
@@ -57,7 +57,7 @@ public class GeminiStreamingMarketDataService implements StreamingMarketDataServ
   @Override
   public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
 
-    Integer maxDepth = args.length > 0 ? (Integer) args[0] : null;
+    int maxDepth = args.length > 0 ? (int) args[0] : 0;
 
     Observable<GeminiOrderbook> subscribedOrderbookSnapshot =
         service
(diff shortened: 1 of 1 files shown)


xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
```diff
@@ -57,7 +57,7 @@ private boolean filterEventsByReason(JsonNode message, String type, String reaso
   @Override
   public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
 
-    Integer maxDepth = args.length > 0 ? (Integer) args[0] : null;
+    int maxDepth = args.length > 0 ? (int) args[0] : 0;
```
badgerwithagun, line 60: args[0] could be null. How about something like

```
int maxDepth = (int) MoreObjects.firstNonNull(args.length > 0 ? args[0] : null, 0);
```

dazito: Yes, makes sense. I will push the changes in a minute.

badgerwithagun (approved):

Issue #2945, opened 2019-02-16
Java 8 EOL end of 2019


timmolter:
Let's plan on cutting off Java 8 support early 2020. It should be an easy transition. I've been using Java11 now for a while.

badgerwithagun:
Seems sensible. My litmus test for this for my own apps is when the top link on Google to download Java doesn't give you Java 8:

https://www.java.com/en/download/

And you don't have to install a plugin in major IDEs to support it:

https://marketplace.eclipse.org/content/java-11-support-eclipse-2018-09-49

In the meantime, everyone should have CI set up to make sure it works, so the switch to using language features can happen at any time.

cymp:
Java 11 now has an [HTTP Client library](https://openjdk.java.net/groups/net/httpclient/intro.html), do you think it means we can get rid of the rescu library? Not that I don't like it, but it would spare a dependency.

timmolter:
It would be a painful transition I believe, as rescu has built a lot of features for XChange as XChange grew. Without knowing the advantages and if it would even work, I'm happy with keeping rescu.

walec51:
rescu is a more abstract client proxy framework which currently uses the old Java HttpURLConnection

it would make more sense to add optional support for the new HTTP client API in rescu

BTW. despite the fact that Oracle announced EOF for Java 8, is Java 11 and its new HTTP client gona get any support on Android?

cymp:
No idea.
Thanks for your answers though.

EricAnderson1000:
Project builds and runs tests in Java 11 just fine. Any reason not to change the target compilation to 11 soon? Who changes the CI builds? 

Tested using 
java --version
openjdk 11.0.5 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
Eclipse OpenJ9 VM AdoptOpenJDK (build openj9-0.17.0, JRE 11 Mac OS X amd64-64-Bit Compressed References 20191016_371 (JIT enabled, AOT enabled)
OpenJ9   - 77c1cf708
OMR      - 20db4fbc
JCL      - 2a7af5674b based on jdk-11.0.5+10)

walec51:
because a lot of people are running this library on java 8 and android

earce:
Does it make sense to consider this for 5.0.0 given it's a major release + the inclusion of xchange-streaming exchange library?

walec51:
does Java 11 run on Android?

earce:
[...]

PR #3776, opened 2020-10-19
JDK11 for JitPack

rpanak-generalbytes:
jitpack.io build fails as it uses JDK8 to build repositiories by default. 
It seems like the JDK 8 incompatibility is a maven plugin problem (fmt-maven-plugin). It depends on google-java-format lib whose published jar is built by a newer JDK and thus can't be used by JDK8. One can specify a different JDK (by means of a docker image) via jitpack.yml file. By mistake this file was not removed (and left in an incorrect location) from PR #3634.
Moving to root project dir should solve the problem. 

JitPack error log for those interested: https://jitpack.io/com/github/GENERALBYTESCOM/XChange/b88e82fab6/build.log


walec51:
Java 8 is not supported any more by this library: https://github.com/knowm/XChange/issues/2945

rpanak-generalbytes:
All the more reason for merging this PR. You already use the same image for your Travis builds.

walec51:
ok

PR #3780, opened 2020-10-22
Update StreamingTradeService.java

cymp:
Not a duplicate of getOrderChanges(), which is not easy to use if we want to keep track of the active open orders.

### DIFF
--- a/xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingTradeService.java
+++ b/xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingTradeService.java
@@ -1,8 +1,11 @@
 package info.bitrich.xchangestream.core;
 
+import java.util.Collection;
+
 import io.reactivex.Observable;
 import org.knowm.xchange.currency.CurrencyPair;
 import org.knowm.xchange.dto.Order;
+import org.knowm.xchange.dto.trade.OpenOrders;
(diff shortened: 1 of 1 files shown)


cymp:
@badgerwithagun I think you know well the streaming topic, maybe you can give your opinion on this. Of course any comments from other people are welcome.

nielsdraaisma:
Wouldn't it be easier to create a fixed method that uses the normal getOpenOrders and getOrderChanges to produce the same?

cymp:
If you want to track the active orders with getOrderChanges(), you'll have some problems and coding needed: you would have to call in your client code TradeService.getOpenOrders() and synchronize the sequences (thus instead a call to a raw method giving the sequence), an information you don't have with StreamingTradeService.getOrderChanges() making the synchronization a problem. 
In my opinion, getOrderChanges() does not have much use compared to a getActiveOrders(). If it were me, I would not have put getOrderChanges() in the streaming interface in the first place, and I would have put getOpenOrders() instead.

earce:
Chiming in here, from a streaming perspective I would imagine the right way to do this is sow & subscribe. These type of methods/endpoints typical send a snapshot of the state of the world and then all deltas, I think that would be the right way to support this method. Otherwise if you want open orders that's a single synchronous request, requesting a snapshot of the sow.

I would look at https://docs.pro.coinbase.com/#the-level2-channel for inspiration.

PR #3791, opened 2020-10-30
code reformats

mdvx:
from the google plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,17 +70,16 @@
   </issueManagement>
 
   <distributionManagement>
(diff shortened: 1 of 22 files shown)


pom.xml
```diff
@@ -70,17 +70,16 @@
   </issueManagement>
 
   <distributionManagement>
```
walec51, line 72: code formatting change that changes our distribution repositories?

timmolter:
I usually do this when I do a release. If there was ONLY formatting in this PR, it wouldn't be a problem though.

PR #3794, opened 2020-11-03
[Streaming] more dependency cleanup

earce:
Cleaning up
- commons-io
- fluent-hc
- fastjson

@timmolter


earce:
@timmolter wanted to follow up on this

PR #3800, opened 2020-11-11
[BX] removing defunct exchange

earce:
https://bitcoinmagazine.com/articles/thai-crypto-exchange-bx-in-th-shuts-down


earce:
@timmolter

PR #3811, opened 2020-11-16
[Upbit] Implements getTickers()

aaron-jang:
- Implements getTickers() 
- Added Upbit util class

### DIFF
--- a/xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
+++ b/xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
@@ -139,7 +139,7 @@ public final class UpbitAdapters {
       status = Order.OrderStatus.FILLED;
     }
     return new LimitOrder(
-        upbitOrderResponse.getSide().equals("ask") ? OrderType.ASK : OrderType.BID,
+        UpbitUtils.fromSide(upbitOrderResponse.getSide()),
         upbitOrderResponse.getVolume(),
         UpbitUtils.toCurrencyPair(upbitOrderResponse.getMarket()),
         upbitOrderResponse.getUuid(),
--- a/xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
+++ b/xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
@@ -128,7 +128,7 @@ public final class UpbitAdapters {
   }
 
   public static Order adaptOrderInfo(UpbitOrderResponse upbitOrderResponse) {
-    Order.OrderStatus status = Order.OrderStatus.NEW;
+    Order.OrderStatus status;
     if (upbitOrderResponse.getState().equalsIgnoreCase("cancel")) {
       status = Order.OrderStatus.CANCELED;
     } else if (upbitOrderResponse.getExecutedVolume().compareTo(BigDecimal.ZERO) == 0) {
@@ -137,6 +137,8 @@ public final class UpbitAdapters {
       status = Order.OrderStatus.PARTIALLY_FILLED;
     } else if (upbitOrderResponse.getState().equalsIgnoreCase("done")) {
       status = Order.OrderStatus.FILLED;
+    } else {
+      status = Order.OrderStatus.NEW;
     }
     return new LimitOrder(
         UpbitUtils.fromSide(upbitOrderResponse.getSide()),
--- a/xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
+++ b/xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
@@ -46,17 +46,16 @@ public final class UpbitAdapters {
     Arrays.stream(upbitOrders)
         .forEach(
             upbitOrder -> {
-              OrderType orderType = OrderType.ASK;
-              BigDecimal price = upbitOrder.getAskPrice();
-              BigDecimal amount = upbitOrder.getAskSize();
-              LimitOrder limitOrder =
-                  new LimitOrder(orderType, amount, currencyPair, null, null, price);
-              asks.add(limitOrder);
-              orderType = OrderType.BID;
-              price = upbitOrder.getBidPrice();
-              amount = upbitOrder.getBidSize();
-              limitOrder = new LimitOrder(orderType, amount, currencyPair, null, null, price);
-              bids.add(limitOrder);
+              asks.add(
+                  new LimitOrder.Builder(OrderType.ASK, currencyPair)
+                      .originalAmount(upbitOrder.getAskSize())
+                      .limitPrice(upbitOrder.getAskPrice())
+                      .build());
+              bids.add(
+                  new LimitOrder.Builder(OrderType.BID, currencyPair)
+                      .originalAmount(upbitOrder.getBidSize())
+                      .limitPrice(upbitOrder.getBidPrice())
+                      .build());
             });
     Map<OrderType, List<LimitOrder>> map = new HashMap<>();
     map.put(OrderType.ASK, asks);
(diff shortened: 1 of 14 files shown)


xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitUtils.java
```diff
@@ -0,0 +1,26 @@
+package org.knowm.xchange.upbit;
+
+import org.knowm.xchange.currency.CurrencyPair;
+
+public class UpbitUtils {
+
+  public static final String MARKET_NAME_SEPARATOR = "-";
+
+  private UpbitUtils() {
```
earce, line 9: This isn't needed

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitUtils.java
```diff
@@ -0,0 +1,26 @@
...
+    return currencyPair.counter.getCurrencyCode().toUpperCase()
+        + MARKET_NAME_SEPARATOR
+        + currencyPair.base.getCurrencyCode().toUpperCase();
+  }
+
+  public static CurrencyPair toCurrencyPair(String pairString) {
+    if (pairString == null) return null;
+    String[] pairStringSplit = pairString.split(MARKET_NAME_SEPARATOR);
+    if (pairStringSplit.length != 2) return null;
+    return new CurrencyPair(pairStringSplit[1], pairStringSplit[0]);
```
earce, line 24: is this index order correct (1 then 0)?

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
```diff
@@ -131,27 +138,27 @@ public static Order adaptOrderInfo(UpbitOrderResponse upbitOrderResponse) {
...
-            orderId,
-            Date.from(dateTime.toInstant()),
-            price,
-            price,
-            cumulativeAmount,
-            upbitOrderResponse.getPaidFee(),
-            status);
-    return order;
+    return new LimitOrder(
+        upbitOrderResponse.getSide().equals("ask") ? OrderType.ASK : OrderType.BID,
```
earce, line 142: put the ask on the left side for NPE protection

aaron-jang: fixed.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitUtils.java
aaron-jang: There was a bug in the previous code.

Check out the api below.
* https://api.upbit.com/v1/trades/ticks?market=BTC-ETH (success)
* https://api.upbit.com/v1/trades/ticks?market=ETH-BTC (fail)

base: ETH (index 1)
counter: BTH(index 0)

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitUtils.java
aaron-jang: do you mean that private constructor is not neseccery?

earce: Yes, it's really about enforcing so that static classes can't be sub classed, final on the class should accomplish the same thing, but neither is necessary.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
```diff
@@ -130,28 +137,30 @@ public static Order adaptOrderInfo(UpbitOrderResponse upbitOrderResponse) {
...
-            currencyPair,
-            orderId,
-            Date.from(dateTime.toInstant()),
-            price,
-            price,
-            cumulativeAmount,
-            upbitOrderResponse.getPaidFee(),
-            status);
-    return order;
+    return new LimitOrder(
```
earce, line 143: also sorry missed this but can you use the builder class that limit order has to do this

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitUtils.java
aaron-jang: Removed.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
aaron-jang: Fixed.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
```diff
@@ -130,28 +136,32 @@ public static Order adaptOrderInfo(UpbitOrderResponse upbitOrderResponse) {
...
+  }
+
+  public static ExchangeMetaData adaptMetadata(List<UpbitMarket> markets) {
+    Map<CurrencyPair, CurrencyPairMetaData> pairMeta =
+        markets.stream()
+            .map(UpbitMarket::getMarket)
+            .map(UpbitUtils::toCurrencyPair)
+            .collect(
+                Collectors.toMap(
+                    Function.identity(), cp -> new CurrencyPairMetaData.Builder().build()));
```
earce, line 164: is there a reason why you aren't using cp here?

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
```diff
@@ -121,7 +127,7 @@ public static Wallet adaptWallet(UpbitBalances wallets) {
   }
 
   public static Order adaptOrderInfo(UpbitOrderResponse upbitOrderResponse) {
-    Order.OrderStatus status = Order.OrderStatus.NEW;
+    Order.OrderStatus status;
     if (upbitOrderResponse.getState().equalsIgnoreCase("cancel")) {
       status = Order.OrderStatus.CANCELED;
     } else if (upbitOrderResponse.getExecutedVolume().compareTo(BigDecimal.ZERO) == 0) {
```
earce, line 133: the compare to == 0 implies executed volume == 0 which would be unfilled, wouldn't partially filled be a number > 0?

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
aaron-jang: The second argument of tomap is a function parameter, so I used cp only to receive an argument. 
Is there any way to recommend it?

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
aaron-jang: Does unfilled mean that the executed value can be null?
In the demo tests, there was no case where the executed value was null.
It was 0.0 when the transaction was not executed.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
earce: I was wondering more like why did you not pass in anything to the CurrencyPairMetaData Builders like .set(currencyPair) before building it, otherwise it doesn't have any data. Is that intentional? If yet that is fine leave it as is.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
earce: Unfilled means executed value == 0, but in this case you are saying if executedVolume == 0 that it's partially filled, which I don't think is right?

   ` } else if (upbitOrderResponse.getExecutedVolume().compareTo(BigDecimal.ZERO) == 0) {
      status = Order.OrderStatus.PARTIALLY_FILLED;`

I don't think executed volume of 0 is partially filled?

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
aaron-jang: Yes, it was intended.
CurrencyPairMetaData has no CurrencyPair field and no more values to fill in.

xchange-upbit/src/main/java/org/knowm/xchange/upbit/UpbitAdapters.java
aaron-jang: Didn't you see the folded code?

Full code

```java
     } else if (upbitOrderResponse.getExecutedVolume().compareTo(BigDecimal.ZERO) == 0) {
       status = Order.OrderStatus.NEW;
     } else if (upbitOrderResponse.getRemainingVolume().compareTo(BigDecimal.ZERO)> 0) {
       status = Order.OrderStatus.PARTIALLY_FILLED;
```

https://github.com/knowm/XChange/pull/3811/files/3ad817e7ccd567c372ab8670af978d9b44f11214#diff-a07acc5ce0d340e36689164bcc4d69b8382ff6f6b174210b9d6a11bdcdd977c1R133-L131

earce: Ah, that's what I was missing, thanks for clarifying.

aaron-jang:
@earce 

Thanks for reviewing the code. 😄

PR #3814, opened 2020-11-19
[OkEx] Fix websocket url for OkExStreamingExchange

dazito:


xchange-stream-okcoin/src/main/java/info/bitrich/xchangestream/okcoin/OkExStreamingExchange.java
```diff
@@ -2,7 +2,7 @@
 
 /** Created by Lukas Zaoralek on 17.11.17. */
 public class OkExStreamingExchange extends OkCoinStreamingExchange {
-  private static final String API_URI = "wss://real.okex.com:10441/websocket?compress=true";
+  private static final String API_URI = "wss://real.okcoin.com:8443/ws/v3";
```
earce, line 5: based on https://www.okex.com/docs/en/#spot_ws-general should it not be wss://real.okex.com:8443/ws/v3 ?

dazito: Yes, you are right. OkCoin, OkEx being the same entity in the end confused me. Fixed it.

PR #3817, opened 2020-11-23
[Bitflyer] Implemented getNonceFactory on BitflyerStreamingExchange

dazito:


xchange-stream-bitflyer/src/main/java/info/bitrich/xchangestream/bitflyer/BitflyerStreamingExchange.java
```diff
@@ -7,6 +7,7 @@
 import io.reactivex.Completable;
 import org.knowm.xchange.ExchangeSpecification;
 import org.knowm.xchange.bitflyer.BitflyerExchange;
+import org.knowm.xchange.utils.nonce.AtomicLongIncrementalTime2014NonceFactory;
```
earce, line 10: get rid of this

PR #3818, opened 2020-11-23
[Core] fixing race conditions in nonce factories and improving naming

earce:
I am fixing some race conditions present in these classes and and renaming them to something a bit cleaner/more consistent.

Moved `CurrentTime1000NonceFactory` and `CurrentNanosecondTimeIncrementalNonceFactory` into `CurrentTimeIncrementalNonceFactory`

- Fixing thread safety
- Minimal arithmetic which is more efficient then while loop
- Using built-in nano time to calculate nano time
- Adding parameter to specify `TimeUnit` to nonce generation.

I would like to submit subsequent reviews with some more cleanup of these classes names. 

@walec51 
@nielsdraaisma


nielsdraaisma:
Looks good!

timmolter:
Go for it. Just know that it was originally designed as one thread per API key. I'm interested in what you come up with.

earce:
Rolled up all nonce factories into a single one with constructor args.

earce:
@walec51 wanted to see if I could bump this up

PR #3826, opened 2020-12-05
[Core] using consistent nonce generator as base exchange default

earce:
- Using a consistent nonce generator as the default for BaseExchange
- Moving over to a thread-safe version of the nonce generator
- Removing old nonce generators which shouldn't be used


earce:
@timmolter

andre77:
@earce 
the old class `org.knowm.xchange.utils.nonce.AtomicLongCurrentTimeIncrementalNonceFactory` is removed. Could you please show, how to replace it, if i need the exact same functionality ?

earce:
@andre77 which exchange needs this exact functionality? Are you unable to use CurrentTimeIncrementalNonceFactory with millis?

PR #3828, opened 2020-12-06
[HitBtc]OrderBook userReference

mdvx:
OrderBook entries do not have userReference, so pass null to LimitOrder constructor


earce:
Can you change this to use a builder?

mdvx:
> Can you change this to use a builder?

Yes, that might even be faster

PR #3831, opened 2020-12-08
[Streaming] removing jaxb dependency and other code cleanup

earce:
- Removing jaxb
- Digest related code cleanup
- Consolidating some code to make simpler and reusing existing code


earce:
@timmolter

PR #3836, opened 2020-12-11
Coinbasepro ratelimiting - WIP

nibarulabs:
Putting this up for any initial comments. Still WIP - will fill in more info when done.

### DIFF
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbasePro.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbasePro.java
@@ -218,19 +218,6 @@ public interface CoinbasePro {
       @QueryParam("after") Integer startingOrderId)
       throws CoinbaseProException, IOException;
 
-//  @GET
-//  @Path("accounts/{account_id}/transfers")
-//  @Consumes(MediaType.APPLICATION_JSON)
-//  CoinbaseProTransfers transfers(
-//      @HeaderParam("CB-ACCESS-KEY") String apiKey,
-//      @HeaderParam("CB-ACCESS-SIGN") ParamsDigest signer,
-//      @HeaderParam("CB-ACCESS-TIMESTAMP") SynchronizedValueFactory<Long> nonce,
-//      @HeaderParam("CB-ACCESS-PASSPHRASE") String passphrase,
-//      @PathParam("account_id") String accountId,
-//      @QueryParam("profile_id") String profileId,
-//      @QueryParam("limit") Integer limit,
-//      @QueryParam("after") String createdAtDate);
-
   @GET
   @Path("transfers")
   @Consumes(MediaType.APPLICATION_JSON)
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbasePro.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbasePro.java
@@ -218,18 +218,18 @@ public interface CoinbasePro {
       @QueryParam("after") Integer startingOrderId)
       throws CoinbaseProException, IOException;
 
-  @GET
-  @Path("accounts/{account_id}/transfers")
-  @Consumes(MediaType.APPLICATION_JSON)
-  CoinbaseProTransfers transfers(
-      @HeaderParam("CB-ACCESS-KEY") String apiKey,
-      @HeaderParam("CB-ACCESS-SIGN") ParamsDigest signer,
-      @HeaderParam("CB-ACCESS-TIMESTAMP") SynchronizedValueFactory<Long> nonce,
-      @HeaderParam("CB-ACCESS-PASSPHRASE") String passphrase,
-      @PathParam("account_id") String accountId,
-      @QueryParam("profile_id") String profileId,
-      @QueryParam("limit") Integer limit,
-      @QueryParam("after") String createdAtDate);
+//  @GET
+//  @Path("accounts/{account_id}/transfers")
+//  @Consumes(MediaType.APPLICATION_JSON)
+//  CoinbaseProTransfers transfers(
+//      @HeaderParam("CB-ACCESS-KEY") String apiKey,
+//      @HeaderParam("CB-ACCESS-SIGN") ParamsDigest signer,
+//      @HeaderParam("CB-ACCESS-TIMESTAMP") SynchronizedValueFactory<Long> nonce,
+//      @HeaderParam("CB-ACCESS-PASSPHRASE") String passphrase,
+//      @PathParam("account_id") String accountId,
+//      @QueryParam("profile_id") String profileId,
+//      @QueryParam("limit") Integer limit,
+//      @QueryParam("after") String createdAtDate);
 
   @GET
   @Path("transfers")
--- a/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
+++ b/xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
@@ -84,7 +84,7 @@ public class CoinbaseProExchange extends BaseExchange {
         concludeHostParams(exchangeSpecification);
 
         this.marketDataService = new CoinbaseProMarketDataService(this);
-        this.accountService = new CoinbaseProAccountService(this);
+        this.accountService = new CoinbaseProAccountService(this, getResilienceRegistries());
         this.tradeService = new CoinbaseProTradeService(this);
     }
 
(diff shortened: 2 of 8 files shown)


xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProAccountServiceRaw.java
```diff
@@ -54,56 +70,102 @@ public CoinbaseProWithdrawCryptoResponse withdrawCrypto(
...
+                    apiKey,
+                    digest,
+                    nonceFactory,
+                    passphrase,
+                    type,
+                    profileId,
+                    before,
+                    after,
+                    limit))
+        .withRetry(retry("transfers"))
```
earce, line 138: why have you added a retry on this one specifically?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProAccountServiceRaw.java
```diff
@@ -54,56 +70,102 @@ public CoinbaseProWithdrawCryptoResponse withdrawCrypto(
...
-    return coinbasePro.getReport(apiKey, digest, nonceFactory, passphrase, reportId);
-  }
-
-  public CoinbaseProTransfers transfers(
-      String accountId, String profileId, int limit, String after) {
-    return coinbasePro.transfers(
-        apiKey, digest, nonceFactory, passphrase, accountId, profileId, limit, after);
+    return decorateApiCall(
+            () -> coinbasePro.getReport(apiKey, digest, nonceFactory, passphrase, reportId))
+        .withRetry(retry("report"))
```
earce, line 119: why have you added a retry on this one specifically?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProMarketDataServiceRaw.java
```diff
@@ -72,10 +96,14 @@ public CoinbaseProProductBook getCoinbaseProProductOrderBook(CurrencyPair curren
...
-              currencyPair.base.getCurrencyCode(),
-              currencyPair.counter.getCurrencyCode(),
-              String.valueOf(level));
+          decorateApiCall(
+                  () ->
+                      coinbasePro.getProductOrderBook(
+                          currencyPair.base.getCurrencyCode(),
+                          currencyPair.counter.getCurrencyCode(),
+                          String.valueOf(level)))
+              .withRateLimiter(rateLimiter(PUBLIC_PER_SECOND_RATE_LIMITER))
```
earce, line 105: why do some of these not have retry logic?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProResilience.java
```diff
@@ -0,0 +1,48 @@
...
+                                .limitForPeriod(3)
+                                .build());
+        registries
+                .rateLimiters()
+                .rateLimiter(
+                        ORDERS_PER_DAY_RATE_LIMITER,
+                        RateLimiterConfig.from(registries.rateLimiters().getDefaultConfig())
+                                .timeoutDuration(Duration.ZERO)
+                                .limitRefreshPeriod(Duration.ofDays(1))
+                                .limitForPeriod(500)
```
earce, line 44: How has this limit been decided?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
```diff
@@ -16,6 +16,7 @@
 import lombok.NoArgsConstructor;
```
earce, line 16: why did the entirety of this class get highlighted as changes?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbasePro.java
```diff
@@ -218,18 +218,18 @@ CoinbaseProSendMoneyResponse sendMoney(
...
+//  @Consumes(MediaType.APPLICATION_JSON)
+//  CoinbaseProTransfers transfers(
+//      @HeaderParam("CB-ACCESS-KEY") String apiKey,
+//      @HeaderParam("CB-ACCESS-SIGN") ParamsDigest signer,
+//      @HeaderParam("CB-ACCESS-TIMESTAMP") SynchronizedValueFactory<Long> nonce,
+//      @HeaderParam("CB-ACCESS-PASSPHRASE") String passphrase,
+//      @PathParam("account_id") String accountId,
+//      @QueryParam("profile_id") String profileId,
+//      @QueryParam("limit") Integer limit,
+//      @QueryParam("after") String createdAtDate);
```
earce, line 232: I assume this will be removed completely right?

earce:
This looks pretty good, impressed with the quick turn around. One bit to consider @walec51 is working on https://github.com/resilience4j/resilience4j/pull/1264 which will allow us to react to 429s which we currently do in that long function I pasted earlier in our conversation. 

By moving to this coinbase in particular will loose its ability to react to 429s. 

Once @walec51 introduces this additional functionality into the resilience lib we will be able to react to them properly.

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbasePro.java
nibarulabs: Yes. I think this was before I pinged you guys about this.

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
nibarulabs: Ah. I think I edited this file before I added the google code format plugin. I will fix this up as expected.

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProResilience.java
nibarulabs: From the coinbase api docs:


PUBLIC ENDPOINTS
We throttle public endpoints by IP: 3 requests per second, up to 6 requests per second in bursts. Some endpoints may have custom rate limits.

PRIVATE ENDPOINTS
We throttle private endpoints by profile ID: 5 requests per second, up to 10 requests per second in bursts. Some endpoints may have custom rate limits.

Orders

You can place two types of orders: limit and market. Orders can only be placed if your account has sufficient funds. Each profile can have a maximum of 500 open orders on a product. Once reached, the profile will not be able to place any new orders until the total number of open orders is below 500. Once an order is placed, your account funds will be put on hold for the duration of the order. How much and which funds are put on hold depends on the order type and parameters specified. See the Holds details below.


Currently the rate limiter doesn't support the concept of having a max open order per product. I put it in there to see what others would say about that.

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProMarketDataServiceRaw.java
nibarulabs: These are all pretty much first pass. I think since you raised this it makes sense - all ```GET``` endpoints should have retry's while the ```POST/DELETE``` ones should not.

nibarulabs: These should be as expected now.

nibarulabs:
@earce Thanks again for the initial review(s). I still need to go through all the tests and whatever else is on the best practices list.

I just looked at the resilience4j pr - so we're basically waiting on a pr for that side to do the circuit breaker type functionality on the 429's? Or somewhere along those lines? I think I get it. Sounds good.

I also have some dependencies on my project coming up. One is Binance.us (I will have to add this - tbh, I think it will mostly be a clone of existing binance code since the api's are pretty close) and the other is possibly Kraken or another U.S. based exchange. I'll need the rate limiting in these as well, so I can add necessary changes there.

After that, I'll see what I can commit to for bringing rate limiting into some of the other exchanges.

earce:
@nibarulabs sounds good, kraken is already on here and for Binance.US I would just make the existing url changeable with a configuration because I believe the APIs are the same

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProResilience.java
earce: Hey so thinking about this more I think we should get rid of this order limiter unless we can control it by product id. If you are trading a lot of symbols you could hit this limit and it would be an artificial limit.

nibarulabs: Ok, yah that makes sense. Will do.

earce:
Closing this because it was already implemented for coinbase

PR #3841, opened 2020-12-17
[Serum-Exchange] initial scaffold commit

earce:
This is some scaffolding and parts required for using [Serum](https://projectserum.com/) 

I am currently in the process of translating typescript and JavaScript logic to be able to subscribe to Serum in Java. There are several parts to getting this working correctly but I think it makes sense to commit this in parts given it's a lot of logic. 

I had to inevitably bring one dependency [Java Binary Block Parser (JBBP)](https://github.com/raydac/java-binary-block-parser) because without it deserializing the raw bytes is a lot of code/hard to read. 

I ended up deciding to just write all the logic to do this directly into XChange itself because I do not want to maintain another library for the time being.


earce:
@timmolter

timmolter:
Looks awesome except for the test failures!!

earce:
Should be good now

PR #3844, opened 2020-12-18
fix surefire dependency in surefire-fix branch

raph431:
Fixed a surefire-plugin dependency in xchange-core

### DIFF
--- a/xchange-core/pom.xml
+++ b/xchange-core/pom.xml
@@ -7,6 +7,33 @@
 	</parent>
 
 	<artifactId>xchange-core</artifactId>
+    <profiles>
+        <profile>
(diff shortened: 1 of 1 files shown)


timmolter:
What does this fix?

raph431:
It allows it to build. Without it, I get 
`[ERROR] java.lang.ClassNotFoundException: org.apache.maven.surefire.junit4.JUnit4Provider`
during mvn install

walec51:
no one else seams to have this problem

surefire dependancy is defined here: https://github.com/knowm/XChange/blob/develop/pom.xml#L511

in which directory are you running `mvn install` and which version of java and maven are you using?

Issue #3845, opened 2020-12-21
Cexio streaming orderbook return an empty snapshot as first response


katamarano:
Hello everyone,
as stated in cexio docs, when requesting an orderbook via WebSocket, we should receive first a snapshot of the book, and then a single message of "update", to edit the local book.
There is a "depth" parameter that according to docs if set to 0 should return all the orderbook as a snapshot in the first message while setting to a value N > 0 return only the first N bids and asks.
In xchange code this parameter is hard coded to 0 in CexioWebSocketOrderBookSubscriptionData.
```java
 public CexioWebSocketOrderBookSubscriptionData(CurrencyPair currencyPair, boolean subscribe) {
    pair =
        new ArrayList<String>(
            Arrays.asList(currencyPair.base.toString(), currencyPair.counter.toString()));
    this.subscribe = subscribe;
    this.depth = 0; // 0 for full depth
  }
```
Well, for me the first snapshot is always empty (empty bids and asks)  so it's impossible to maintain the local orderbook in sync.
If instead, I use a depth  N > 0, the snapshot contains the first N bids and docs as stated in the doc.

So would be nice to have the possibility to specify the depth during the request, and propagate this parameter till CexioWebSocketOrderBookSubscriptionData.
For instance, during the getOrderBook request:
```java
exchange.getStreamingMarketDataService().getOrderBook(currencyPair, 100 /* depth value */).subscribe( //other code
```
so to have it propagated to:
```java
 public CexioWebSocketOrderBookSubscriptionData(CurrencyPair currencyPair, boolean subscribe, int depth) 
```
I already made this change, using the varargs parameter provided by the existent implementation, if needed.

earce:
Submit a PR if this fixes your issue we can review

katamarano:
> Submit a PR if this fixes your issue we can review

Just created a PR for it

earce:
Closed

PR #3848, opened 2020-12-21
Cexio streaming orderbook return an empty snapshot as first response …

katamarano:
…#3845

### DIFF
--- a/xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
+++ b/xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
@@ -74,15 +74,7 @@ public class CexioStreamingRawService extends JsonNettyStreamingService {
       case ORDERBOOK:
         {
           CurrencyPair currencyPair = (CurrencyPair) args[0];
-          int depth = 0;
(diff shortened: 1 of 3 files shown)


xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingRawService.java
```diff
@@ -73,7 +74,16 @@ private Object GetEventSubscriptionData(String channelName, boolean isSubscribe,
       case ORDERBOOK:
         {
           CurrencyPair currencyPair = (CurrencyPair) args[0];
-          return new CexioWebSocketOrderBookSubscriptionData(currencyPair, isSubscribe);
+          int depth = 0;
+          if (args[1] != null && args[1] instanceof Object[] && ((Object[]) args[1]).length > 0) {
+            Object[] objArgs = (Object[]) args[1];
```
earce, line 79: Why is this second arg an array?

katamarano: It was the result of passing a vararg parameter, inside another varargs parameter. Yes, you are right, not needed and I already pushed a change to make things easier and clearer.

PR #3874, opened 2021-01-09
[FTX] Adding new exchange

makarid:


xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
```diff
@@ -431,6 +431,10 @@ public OrderType getOpposite() {
     REJECTED,
     /** Order has expired it's time to live or trading session and been removed from order book */
     EXPIRED,
+    /** Order is open and waiting to be filled */
+    OPEN,
+    /** Order has been either filled or cancelled */
+    CLOSED,
```
earce, line 437: Some of these states overlap each other, a partially filled order is technically opened, same a filled order can be closed, I am not sure if these states have not been added here before for a reason?

@timmolter

xchange-ftx/src/main/java/org/knowm/xchange/ftx/FtxAdapters.java
```diff
@@ -0,0 +1,275 @@
...
+            ftxMarketDto -> {
+              CurrencyPairMetaData currencyPairMetaData =
+                  new CurrencyPairMetaData.Builder()
+                      .amountStepSize(ftxMarketDto.getPriceIncrement())
+                      .minimumAmount(ftxMarketDto.getSizeIncrement())
+                      .priceScale(ftxMarketDto.getPriceIncrement().scale())
+                      .baseScale(ftxMarketDto.getSizeIncrement().scale())
+                      .build();
+
+              if (ftxMarketDto.getType().equals("spot")) {
```
earce, line 146: move string to left side for NPE protection

xchange-ftx/src/main/java/org/knowm/xchange/ftx/FtxExchange.java
```diff
@@ -0,0 +1,74 @@
...
+import org.knowm.xchange.ftx.service.*;
+import org.knowm.xchange.service.account.AccountService;
+import org.knowm.xchange.service.marketdata.MarketDataService;
+import org.knowm.xchange.service.trade.TradeService;
+import org.knowm.xchange.utils.nonce.CurrentTimeIncrementalNonceFactory;
+import si.mazi.rescu.SynchronizedValueFactory;
+
+public class FtxExchange extends BaseExchange implements Exchange {
+
+  private final SynchronizedValueFactory<Long> nonceFactory =
```
earce, line 19: This and the overridden method isn't necessary as it's the default.

xchange-core/src/main/java/org/knowm/xchange/dto/Order.java
makarid: Hello, these are the exact words that FTX sends with their response. They have only these 3

Issue #3863, opened 2021-01-01
Gemini WSS v2 Support


9-9-0:
Is anyone currently working on this?
https://docs.gemini.com/websocket-api/#level-2-data
The /v2 stream allows subscribing to multiple pairs over a single connection.

I can take a look at it if not.

earce:
I don't think anyone is currently working on this would be great if you wanted to!

earce:
Closed as merged

PR #3875, opened 2021-01-10
Separate Gemini v1 and v2 websocket integrations, Setup Gemini v2 Str…

9-9-0:


9-9-0:
https://github.com/knowm/XChange/issues/3863

A few open items:
- I did not implement candlestick streaming
- When you subscribe to the L2 price feed, trade and auction(although I didn't see any of these during my testing) messages are streamed in addition to l2_update messages. I didn't implement any behavior to react to non-l2 messages.

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/v2/GeminiStreamingMarketDataService.java
```diff
@@ -0,0 +1,72 @@
...
+    private final Map<CurrencyPair, SortedMap<BigDecimal, BigDecimal>> bids =
+            new ConcurrentHashMap<>();
+    private final Map<CurrencyPair, SortedMap<BigDecimal, BigDecimal>> asks =
+            new ConcurrentHashMap<>();
+
+    public GeminiStreamingMarketDataService(GeminiStreamingService service) {
+        this.service = service;
+    }
+
+    private boolean containsPair(List<CurrencyPair> pairs, CurrencyPair pair) {
```
earce, line 35: you can probably remove this method and use single line stream code

`pairs.stream().anyMatch(item -> item.compareTo(pair) == 0)`

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/v2/dto/GeminiOrderbook.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  public Collection<GeminiLimitOrder> getAsks() {
+    return asks.values();
+  }
+
+  public Collection<GeminiLimitOrder> getBids() {
+    return bids.values();
+  }
```
earce, line 62: Not sure if we intend the orderbook data structure to be threadsafe here but I imagine if you call update on a side and also on a separate thread iterate over that side I suspect you will get a concurrent modification exception.

earce:
Two comments
1). Is there anyway to subscribe to just the l2 messages? At the least I imagine non-l2 messages are filtered out?
2). I tend to prefer if you are building a v2 of the exchange to create a separate module xchange-gemini-v2 or something along those lines so that when it comes time to remove v1 it is a clean removal @timmolter not sure if you have a preference of how to do this.

timmolter:
@earce no preference, I've seen both options.

9-9-0:
In that case, I'll go ahead and create a separate module. @earce I'll take a look at the issues you raised in your comments sometime later this week.

xchange-stream-gemini-v2/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+        <version>5.0.5-SNAPSHOT</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.knowm.xchange</groupId>
+            <artifactId>xchange-stream-core</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+        <dependency>
```
earce, line 23: do you need this I believe its inherited

xchange-stream-gemini-v2/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
```diff
@@ -0,0 +1,72 @@
...
+    @Override
+    public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
+        if (!containsPair(service.getProduct().getOrderBook(), currencyPair))
+            throw new UnsupportedOperationException(
+                    String.format("The currency pair %s is not subscribed for orderbook", currencyPair));
+
+        int maxDepth = (int) MoreObjects.firstNonNull(args.length > 0 ? args[0] : null, 1);
+
+        return service.getRawWebSocketTransactions(currencyPair, false)
+                .filter(message -> message.getType().equals(L2_UPDATES))
```
earce, line 53: set constant on left for NPE

xchange-stream-gemini-v2/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
```diff
@@ -0,0 +1,72 @@
...
+            new ConcurrentHashMap<>();
+    private final Map<CurrencyPair, SortedMap<BigDecimal, BigDecimal>> asks =
+            new ConcurrentHashMap<>();
+
+    public GeminiStreamingMarketDataService(GeminiStreamingService service) {
+        this.service = service;
+    }
+
+    private boolean containsPair(List<CurrencyPair> pairs, CurrencyPair pair) {
+        for (CurrencyPair item : pairs) {
```
earce, line 36: can we remove this method as per previous comment

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/v2/GeminiStreamingMarketDataService.java
9-9-0: Done!

xchange-stream-gemini-v2/pom.xml
9-9-0: You're correct, removed the dependency.

xchange-stream-gemini-v2/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
9-9-0: Done!

xchange-stream-gemini-v2/src/main/java/info/bitrich/xchangestream/gemini/GeminiStreamingMarketDataService.java
9-9-0: Done!

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/v2/dto/GeminiOrderbook.java
9-9-0: I'll take a deeper look into this and get back to you.

9-9-0: Went ahead and removed GeminiOrderBook.java and GeminiLimitOrder.java from V2 since they're not being used.

I stitched the logic from coinbasepro's stream for processing changes in Gemini's MarketDataService when I worked on this. The bids and asks are ConcurrentHashMaps which should preserve thread safety, though they're being modified by the .toOrderBook() method, which doesn't seem all that clean. All of this should run on a single thread though if I understand it correctly (the netty nio thread?).

9-9-0:
> Two comments
> 1). Is there anyway to subscribe to just the l2 messages? At the least I imagine non-l2 messages are filtered out?

I don't think so, the trade, auction, and heartbeat events are part and parcel to the L2 channel. They are being filtered out, at least no logic for processing them is currently in place.

xchange-stream-gemini/src/main/java/info/bitrich/xchangestream/gemini/v2/dto/GeminiOrderbook.java
earce: This should run on a single thread, but if you pass around that data structure while it's getting updated by the netty thread another thread will most likely be traversing it.

PR #3878, opened 2021-01-12
Binance trade fee inclusion in order

CayenneLow:


xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -49,6 +50,8 @@
 import org.knowm.xchange.service.trade.params.orders.OrderQueryParams;
 import org.knowm.xchange.utils.Assert;
 
+import lombok.Value;
```
earce, line 53: is this necessary/used?

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -321,12 +332,39 @@ public OpenOrdersParams createOpenOrdersParams() {
...
-                    null)));
+        Order order = BinanceAdapters.adaptOrder(
+                        super.orderStatus(
+                          orderQueryParamCurrencyPair.getCurrencyPair(),
+                          BinanceAdapters.id(orderQueryParamCurrencyPair.getOrderId()),
+                          null
+                        )
+                      );
+
+        // Calculate fees by collecting trades associated with this order
```
earce, line 343: should this logic reside inside of adaptOrder?

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -321,12 +332,39 @@ public OpenOrdersParams createOpenOrdersParams() {
...
+        BinanceTradeHistoryParams tradeHistParams = new BinanceTradeHistoryParams(orderQueryParamCurrencyPair.getCurrencyPair());
+        UserTrades tradeHist = getTradeHistory(tradeHistParams);
```
walec51, line 345: scanning the entire trade history for each order will not we acceptable for performance reasons

walec51:
if we cannot pass the orders id to getTradeHistory then this feature is not doable in a performant manner

Issue #3866, opened 2021-01-03
Source Kraken CurrencyPair Metadata from API


gmax0:
https://api.kraken.com/0/public/AssetPairs

@earce @timmolter is there a field in the metadata map I can use for volume precision similar to priceScale? amountStepSize?

PR #3903, opened 2021-01-31
Source CurrencyPairMetaData from api.kraken.com/0/public/AssetPairs

9-9-0:
https://github.com/knowm/XChange/issues/3866


xchange-kraken/src/main/resources/kraken.json
```diff
@@ -1,269 +1,4 @@
 {
```
earce, line 1: is this no longer needed?

9-9-0:
@earce Removed the currency metadata from the .json file.
Withdrawal fees and minimums aren't included in the metadata since Kraken's API requires authentication and a withdrawal key name to return that data.

earce:
@timmolter can you do a once-over on this and if it looks good I or you can merge

Issue #3907, opened 2021-02-08
Binance "invalid WebSocket Extension handshake"


feelus:
Hello,

I'm suddenly experiencing issues connecting to Binance streaming API, since 13:20 UTC today. I checked their changelog and there seem to have been some changes today to the API, see https://binance-docs.github.io/apidocs/spot/en/

It can be reproduced with this example

```java
    StreamingExchange exchange = StreamingExchangeFactory.INSTANCE.createExchange(BinanceStreamingExchange.class);

    ProductSubscription streamSubscription = ProductSubscription.create()
    .addOrderbook(CurrencyPair.BTC_EUR)
    .build();

    exchange.connect(streamSubscription).blockingAwait();

    Disposable subscription1 = exchange.getStreamingMarketDataService()
        .getOrderBook(CurrencyPair.BTC_EUR)
        .subscribe(
            book -> System.out.println(book),
            throwable -> System.out.println(throwable));
 ```

The error I'm getting is

```
Exception in thread "main" io.netty.handler.codec.CodecException: invalid WebSocket Extension handshake for "permessage-deflate; server_no_context_takeover; client_max_window_bits=15"
        at io.netty.handler.codec.http.websocketx.extensions.WebSocketClientExtensionHandler.channelRead(WebSocketClientExtensionHandler.java:110)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.handler.timeout.IdleStateHandler.channelRead(IdleStateHandler.java:286)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:103)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
[...]

PR #3908, opened 2021-02-08
#3907 Quick fix for binance stream due to invalid websocket exception

paolo-rendano:
Fix to solve the issue reported.
Warn: this is based on magic test, and not on Binance changelog since there is no reference at all to the change required.

### DIFF
--- a/xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
+++ b/xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingService.java
@@ -1,9 +1,9 @@
 package info.bitrich.xchangestream.binance;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import info.bitrich.xchangestream.binance.netty.WebSocketClientCompressionAllowClientNoContextHandler;
 import info.bitrich.xchangestream.core.ProductSubscription;
 import info.bitrich.xchangestream.service.netty.JsonNettyStreamingService;
+import info.bitrich.xchangestream.service.netty.WebSocketClientCompressionAllowClientNoContextAndServerNoContextHandler;
 import io.netty.handler.codec.http.websocketx.extensions.WebSocketClientExtensionHandler;
 
 import java.io.IOException;
@@ -48,7 +48,7 @@ public class BinanceStreamingService extends JsonNettyStreamingService {
 
   @Override
   protected WebSocketClientExtensionHandler getWebSocketClientExtensionHandler() {
-    return WebSocketClientCompressionAllowClientNoContextHandler.INSTANCE;
+    return WebSocketClientCompressionAllowClientNoContextAndServerNoContextHandler.INSTANCE;
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


earce:
Can you move this to the same package as all the other context handlers for consistency?

paolo-rendano:
done. Put in “netty” package as the coinbasepro project

Il giorno 8 feb 2021, alle ore 19:34, Erick Arce <notifications@github.com <mailto:notifications@github.com> > ha scritto:


Can you move this to the same package as all the other context handlers for consistency?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or unsubscribe.

<http://schema.org>

earce:
no it should not be inside of binance at all, just the netty package

paolo-rendano:
> no it should not be inside of binance at all, just the netty package

I think it should stay in Binance project, since each exchange may require different settings on this (or eventually not needing it). Basically this is a customized context handler for the Binance case. Not a general change for any stream exchange.

earce:
We have done this in the past and we end up including it in the common package as more then one exchange ended up using it. This makes it easier to find and reference.  I won't approve it in current form, maybe another approver can if they are comfortable with it.

earce:
Is there a reason you thing having them for general usage is an error? The usage of an **ExtensionHandler** class is something that is very deliberate. The default class is for `getWebSocketClientExtensionHandler()` is `WebSocketClientCompressionHandler.java`, the only time one would override them is if the default class doesn't work. 

In the past we had a similar situation with `WebSocketClientCompressionAllowClientNoContextHandler.java` where it was only available to CoinbasePro but we ended up needing the same configuration in Kraken and making it available required bringing the class into the common classes. I am of the opinion that having known configurations available as common class is better and just more practical if a service provider switches to using a known configuration.

Your proposal is fine but I would prefer that happening in a separate review in the spirit of merging this change. As well as following an already used pattern (having `WebSocketClientCompressionAllowClientNoContextHandler.java` in the common lib).

Let me hear your thoughts

paolo-rendano:
> Is there a reason you thing having them for general usage is an error? The usage of an **ExtensionHandler** class is something that is very deliberate. The default class is for `getWebSocketClientExtensionHandler()` is `WebSocketClientCompressionHandler.java`, the only time one would override them is if the default class doesn't work.

PR: Exactly, infact what I'm saying is that once the general purpose implementation (the default), is not enough, you are specializing it for a specific purpose. In this case of Binance, this class will set several parameters. E.g. :
- compressionLevel: 6
- max window size: 15
- allow client no context: true
- allow server no context: true
- the other two handshaker 
if this is not generalized, for example, you are duplicating the code in coinbasepro, while if we want to generalize it then we can reuse across different exchanges. 

> In the past we had a similar situation with `WebSocketClientCompressionAllowClientNoContextHandler.java` where it was only available to CoinbasePro but we ended up needing the same configuration in Kraken and making it available required bringing the class into the common classes. I am of the opinion that having known configurations available as common class is better and just more practical if a service provider switches to using a known configuration.
> 
> Your proposal is fine but I would prefer that happening in a separate review in the spirit of merging this change. As well as following an already used pattern (having `WebSocketClientCompressionAllowClientNoContextHandler.java` in the common lib).
> 
> Let me hear your thoughts

PR: I agree on the spirit of merging this quick fix, so the last thing that we can do is putting them as you told, changing the name of the class describing it with the parameters that are specific of this implementation. To be clear the class would be named WebSocketClientCompressionAllowClientNoContextAndServerNoContextHandler and put in the package info.bitrich.xchangestream.service.netty of project xchange-stream-service-netty

sounds good?

earce:
This sounds good to me!

PR #3909, opened 2021-02-08
Allow to override web socket client extention handler

0xfaulty:
Today I faced with issue that binance exchange start to force 'server_no_context_takeover' flag for handshaker and i received
`io.netty.handler.codec.CodecException: invalid WebSocket Extension handshake for "permessage-deflate; server_no_context_takeover; client_max_window_bits=15"`
I checked sources and there is no way how I can change that behaviour without copy and edit all sources for binance, so I suggest to add method for set custom handler if that needed.


earce:
This was handled https://github.com/knowm/XChange/pull/3908

Issue #3932, opened 2021-02-22
[Deribit] NumberFormatException while parsing stop market orders from json


treerz:
Stop-market orders do not have a numeric `price` attribute, they have the placeholder string `"market"` instead.

When the class `org.knowm.xchange.deribit.v2.dto.trade.Order` tries to parse the `price` attribute as a `BigDecimal` it throws an unhandled `NumberFormatException`.

Since the `price` attribute is practically "empty" in this case (no meaningful value), just catching the exception and using `null` as the value would solve the issue.

PR #3933, opened 2021-02-22
Fix: [Deribit] NumberFormatException while parsing stop market orders from json #3932

treerz:
Fixes issue #3932

* deribit fix: handle non-numeric marker values in order.price


timmolter:
I'd rather see `price` be defined as a String and then have a custom getter. Otherwise you're throwing away info that may be useful like if price was set to `market`.

treerz:
Sure, i've changed it to a custom getter instead. It makes sense to keep the original value.

PR #3943, opened 2021-02-24
Issue 3914/binance live subscription unsubscription

dpageau2021:
Add the Binance Live Subscribing/Unsubscribing to streams feature.
https://github.com/binance/binance-spot-api-docs/blob/master/web-socket-streams.md#live-subscribingunsubscribing-to-streams

### DIFF
--- a/xchange-stream-binance/README.md
+++ b/xchange-stream-binance/README.md
@@ -2,7 +2,7 @@
 ## Binance
 
 ### Live Subscription/Unsubscription
-The feature to support Live Subscribe/Unsubcribe to streams has been added in version 5.0.7.
+The feature to support Live Subscribe/Unsubcribe to streams has been added on 2021-02-24.
 This allow to subscribe new currency pairs without disconnecting the streams.
 
 To use this feature, follow these steps:
--- a/xchange-stream-binance/README.md
+++ b/xchange-stream-binance/README.md
@@ -1,0 +1,52 @@
+
+## Binance
+
+### Live Subscription/Unsubscription
+The feature to support Live Subscribe/Unsubcribe to streams has been added in version 5.0.6.
+This allow to subscribe new currency pairs without disconnecting the streams.
+
+To use this feature, follow these steps:
+```java
+ExchangeSpecification spec = StreamingExchangeFactory.INSTANCE.createExchange(BinanceStreamingExchange.class)
+    .getDefaultExchangeSpecification();
+BinanceStreamingExchange exchange = (BinanceStreamingExchange) StreamingExchangeFactory.INSTANCE.createExchange(spec);
+
+// First, we need to subscribe to at least one currency pair at connection time
+// Note: at connection time, the live subscription is disabled
+ProductSubscription subscription =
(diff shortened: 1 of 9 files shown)


xchange-stream-binance/README.md
```diff
@@ -0,0 +1,52 @@
+
+## Binance
+
+### Live Subscription/Unsubscription
+The feature to support Live Subscribe/Unsubcribe to streams has been added in version 5.0.7.
```
earce, line 5: would probably eliminate this version `5.0.7` in case version released differs

earce:
@dpageau2021 

The way I understand this, we are making it so that Binance changes from requiring all explicit subscriptions at connection time to connecting and sending individual subscriptions (and being able to unsubscribe individually). My commentary that follows draws on this assumption.

If we move to this approach existing users of this module will most likely break correct? They will not explicit be subscribing in the way they have been in the past and expect to?

Does it make sense to leave the default functionality as it was and make your code something you can enable via a configuration? So library continues to work as it does and folks who want to use the library as you are planning to can just enable that functionality?

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
```diff
@@ -110,6 +110,12 @@ public Completable connect(ProductSubscription... args) {
     streamingAccountService = new BinanceStreamingAccountService(userDataStreamingService);
     streamingTradeService = new BinanceStreamingTradeService(userDataStreamingService);
 
+    // We disable live subscription at connection time because we don't want to send a subscribe message through the websocket
+    // (not required), and the websocket is limited to 5 incoming messages per second (cannot subscribe to many streams
+    // at the same time via live Subscribing/Unsubscribing feature).
+    // See https://github.com/binance/binance-spot-api-docs/blob/master/web-socket-streams.md#websocket-limits for more details
+    disableLiveSubscription();
```
earce, line 117: why is this required? do we not default to false?

xchange-stream-binance/README.md
dpageau2021: Done

dpageau2021:
> @dpageau2021
> 
> The way I understand this, we are making it so that Binance changes from requiring all explicit subscriptions at connection time to connecting and sending individual subscriptions (and being able to unsubscribe individually).
> 
> Just to confirm if users upgrade to this version of this library they should continue to work as it does now, this change is completely backward compatible?

The explicit subscriptions at connection time doesn't change and if the live subscription is not enabled the behaviour will continue as it does now (live subscription is off by default). 

I confirm that there are no breaking change with this fix and it will continue to work as before. It only allow the users to use live subscription/unsubscription on demand (by enabling the live subscription).

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
dpageau2021: You are right it's totally useless, I removed this block (the **isLiveSubscriptionEnabled** variable is false by default in BinanceStreamingService).

FYI in this commit I also added a validation if streamingService is null in case a user try to enable live subscription before calling connect.

PR #3944, opened 2021-02-25
implement Bitmex orderbook10 ws connection, implement dataservice.getRawOrderBook

akuloff:
implement Bitmex orderbook10 websocket channel connection
add BitmexStreamingMarketDataService.getRawOrderBook method

### DIFF
--- a/xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
+++ b/xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
@@ -70,8 +70,8 @@ public class BitmexStreamingMarketDataService implements StreamingMarketDataServ
         .map(s -> {
           RawOrderBook orderBook = s.toRawOrderBook();
           if (orderBook != null) {
-            List<LimitOrder> asks = new ArrayList<>(10);
-            List<LimitOrder> bids = new ArrayList<>(10);
+            List<LimitOrder> asks = new ArrayList<>(orderBook.getAsks().size());
+            List<LimitOrder> bids = new ArrayList<>(orderBook.getBids().size());
             orderBook.getAsks().forEach(r -> {
               LimitOrder order = new LimitOrder.Builder(ASK, currencyPair)
                 .originalAmount(r.get(1))
--- a/xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
+++ b/xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
@@ -68,10 +68,10 @@ public class BitmexStreamingMarketDataService implements StreamingMarketDataServ
       return streamingService
         .subscribeBitmexChannel(channelName)
         .map(s -> {
-          List<LimitOrder> asks = new ArrayList<>();
-          List<LimitOrder> bids = new ArrayList<>();
           RawOrderBook orderBook = s.toRawOrderBook();
           if (orderBook != null) {
+            List<LimitOrder> asks = new ArrayList<>(10);
+            List<LimitOrder> bids = new ArrayList<>(10);
             orderBook.getAsks().forEach(r -> {
               LimitOrder order = new LimitOrder.Builder(ASK, currencyPair)
                 .originalAmount(r.get(1))
@@ -86,8 +86,9 @@ public class BitmexStreamingMarketDataService implements StreamingMarketDataServ
                 .build();
               bids.add(order);
             });
+            return new OrderBook(new Date(), asks, bids);
           }
-          return new OrderBook(new Date(), asks, bids);
+          return new OrderBook(new Date(), Collections.emptyList(), Collections.emptyList());
         });
     } else {
       return streamingService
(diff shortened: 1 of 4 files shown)


xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
```diff
@@ -48,33 +53,72 @@ private String getBitmexSymbol(CurrencyPair currencyPair) {
...
+      } else {
+        channelName = String.format("orderBookL2_25:%s", instrument);
+      }
+    }
+
+    if (fullBook) {
+      return streamingService
+        .subscribeBitmexChannel(channelName)
+        .map(s -> {
+          List<LimitOrder> asks = new ArrayList<>();
```
earce, line 71: can you preallocate with orderbook.getBids()/getAsks() .size()? will reduce garbage generation

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
```diff
@@ -48,33 +53,72 @@ private String getBitmexSymbol(CurrencyPair currencyPair) {
...
-                if (orderbook == null) {
-                  return new OrderBook(
-                      new Date(), Collections.emptyList(), Collections.emptyList());
-                }
-                BitmexLimitOrder[] levels = s.toBitmexOrderbookLevels();
-                orderbook.updateLevels(levels, action);
+          s -> {
+            BitmexOrderbook orderbook;
+            String action = s.getAction();
+            if (action.equals("partial")) {
```
earce, line 99: constant on left side for NPE protection

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
```diff
@@ -48,33 +53,72 @@ private String getBitmexSymbol(CurrencyPair currencyPair) {
...
+              orderbook = s.toBitmexOrderbook();
+              orderbooks.put(instrument, orderbook);
+            } else {
+              orderbook = orderbooks.get(instrument);
+              // ignore updates until first "partial"
+              if (orderbook == null) {
+                return new OrderBook(
```
earce, line 106: will we throw is we just return null instead of allocating empty orderbook structure? if not then just return null

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/dto/BitmexWebSocketTransaction.java
```diff
@@ -90,6 +92,15 @@ public BitmexFunding toBitmexFunding() {
     return funding;
   }
 
+  public RawOrderBook toRawOrderBook() {
+    try {
+      return mapper.treeToValue(this.data.get(0), RawOrderBook.class);
+    } catch (JsonProcessingException e) {
+      e.printStackTrace();
```
earce, line 99: would prefer to log using logging framework, log line will be explicit with location and what is triggering this error no need for stack trace

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
akuloff: fixed (only if orderBook is not null, if null then "old" logic implemented - return empty lists

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
akuloff: fixed

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
akuloff: I think it may affects external systems logic (actually it is not my code, may touch it with formatting), so i think this should be not changed

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/dto/BitmexWebSocketTransaction.java
akuloff: fixed in all exceptions handling

xchange-stream-bitmex/src/main/java/info/bitrich/xchangestream/bitmex/BitmexStreamingMarketDataService.java
```diff
@@ -48,33 +53,72 @@ private String getBitmexSymbol(CurrencyPair currencyPair) {
...
+      }
+    }
+
+    if (fullBook) {
+      return streamingService
+        .subscribeBitmexChannel(channelName)
+        .map(s -> {
+          RawOrderBook orderBook = s.toRawOrderBook();
+          if (orderBook != null) {
+            List<LimitOrder> asks = new ArrayList<>(10);
```
earce, line 73: this might actually be worse, given the default is 16, why don't you set it to orderBook.getAsks().size() and orderBook.getBids().size() ?

akuloff: fixed, my fault (actually it will be 10  due to channel description: "orderBook10")

PR #3947, opened 2021-02-26
[btcmarkets] - Restored service unit tests, added post only support

nielsdraaisma:
- Added support for POST_ONLY order flag
- Restored tests disabled by @timmolter earlier this year


timmolter:
thanks!

Issue #3946, opened 2021-02-26
BTC Markets - withdrawFunds should return ID


zdary:
BTCMarket now provides ID (fundTransferId ) when creating withdrawal. 
Add this id to the BTCMarketsAccountService

See 
https://api.btcmarkets.net/doc/v3#tag/Fund-Management-APIs/paths/~1v3~1withdrawals/post

This commit is no longer needed
https://github.com/knowm/XChange/pull/2687/commits/02e66820c17aad90e2a500a9af0b378b180ef8b1

The fundTransferId can be used to obtain blockchain hash later.

PR #3948, opened 2021-02-26
#3946 BTC Markets - withdrawCrypto should return fundTransferId. that id c…

zdary:
…an then be used to obtain blockchain hash.


earce:
This may have broken a test BTCMarketsAccountServiceTest can you take a look?

earce:
you can ignore i fixed it

PR #3954, opened 2021-02-28
[btcmarkets] - Added support for userrefs / clientOrderId

nielsdraaisma:


xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
```diff
@@ -6,9 +6,9 @@
 
 public class BTCMarketsBaseResponse {
 
-  private final Boolean success;
-  private final String errorMessage;
-  private final Integer errorCode;
+  public final Boolean success;
+  public final String errorMessage;
+  public final Integer errorCode;
```
walec51, line 11: restore the getters in all DTOs

by hand or lombok pls

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsException.java
```diff
@@ -6,22 +6,31 @@
...
 
   public BTCMarketsException(
       @JsonProperty("success") Boolean success,
       @JsonProperty("errorMessage") String errorMessage,
       @JsonProperty("errorCode") Integer errorCode,
       @JsonProperty("clientRequestId") String clientRequestId,
       @JsonProperty("id") Long id,
-      @JsonProperty("responses") List<BTCMarketsException> responses) {
+      @JsonProperty("responses") List<BTCMarketsException> responses,
+      // V3
```
walec51, line 23: use different exception class for that pls, rename this one to BTCMarketsLegacyException or add BTCMarketsExceptionV3

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
```diff
@@ -6,9 +6,9 @@
 
 public class BTCMarketsBaseResponse {
 
-  private final Boolean success;
-  private final String errorMessage;
-  private final Integer errorCode;
+  public final Boolean success;
```
walec51, line 9: lets use a primitive boolean - is much easier to use then a three state Boolean

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/BTCMarketsAuthenticated.java
```diff
@@ -43,16 +43,6 @@ BTCMarketsOrders getOpenOrders(
...
 
-  @POST
-  @Path("order/history")
-  @Consumes(MediaType.APPLICATION_JSON)
-  BTCMarketsOrders getOrderHistory(
-      @HeaderParam("apikey") String publicKey,
-      @HeaderParam("timestamp") SynchronizedValueFactory<Long> nonceFactory,
-      @HeaderParam("signature") BTCMarketsDigest signer,
-      BTCMarketsOpenOrdersRequest request)
-      throws BTCMarketsException, IOException;
```
walec51, line 54: if its DTOs are not incorrect I think its worth leaving and exposing it at least via the *Raw service

walec51:
besides 3 small remarks loots good

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/BTCMarketsAuthenticated.java
nielsdraaisma: Why would we leave in an old implementation that's no longer in use?, that API will eventually be deprecated. clearing out the old client code gives a good overview of how far along the migration is.

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
nielsdraaisma: Most PR's ive done and seen gone through have opted for public finals instead of getters ?

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
nielsdraaisma: I'd have to check if it's present in all responses, null is desireable if it's not present in the response.

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
walec51: most DTOs have getters, core DTOs have getters with one exception which will be corrected in a future release

its best t ostick with one convetion

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/BTCMarketsAuthenticated.java
walec51: if its getting deprecated then ok, lets remove it

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
walec51: if its not present when we could assume its false but ok, it might be better to do it in service code

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
walec51: and in the case of core DTO we should definitely stick to getters as without them it will be hard to refine the API in a backward compatible manner

nielsdraaisma:
> besides 3 small remarks everything looks good

updated.

xchange-btcmarkets/src/main/java/org/knowm/xchange/btcmarkets/dto/BTCMarketsBaseResponse.java
timmolter: DTO members should be private final and have getters or if using Lombok they can be droppped.

nielsdraaisma: updated in last commits.

nielsdraaisma:
@walec51 @earce All review comments done, are we able to get this merged in ?

earce:
@nielsdraaisma good by me, @walec51 what do you think?

walec51:
everything look ok now

there is only a small conflict that need resolving

PR #3962, opened 2021-03-02
[CoinbasePro] incorporating rate limiting into CoinbasePro

earce:
Incorporating rate limiting into CoinbasePro. Additional code cleanup


earce:
@walec51 I wanted to get your opinion in how I setup `public ResilienceRegistries getResilienceRegistries()` which just sets up a singleton. I think this is better setup as a singleton from the `CoinbaseProResilience` layer, want to get your thoughts.

timmolter:
Looks good @earce . Thanks for all your work as usual!

earce:
@walec51 are you good with this?

earce:
@timmolter if you are happy with this I can merge it?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
```diff
@@ -255,35 +238,37 @@ public static Order adaptOrder(CoinbaseProOrder order) {
...
+          case "canceled":
+            return OrderStatus.CANCELED;
+        }
+        return OrderStatus.UNKNOWN;
     }
 
     if (order.getFilledSize().signum() == 0) {
 
-      if (order.getStatus().equals("open") && order.getStop() != null) {
+      if ("open".equals(order.getStatus()) && order.getStop() != null) {
```
walec51, line 264: I don't see why we are stating to prefer this type of Yoda conditions?

it looks strange like this would: `null != order.getStop()`

timmolter:
yes, looks good!

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
```diff
@@ -110,6 +111,11 @@ public ExchangeSpecification getDefaultExchangeSpecification() {
     throw new UnsupportedOperationException("CoinbasePro uses timestamp rather than a nonce");
   }
 
+  @Override
+  public ResilienceRegistries getResilienceRegistries() {
+    return CoinbaseProResilience.INSTANCE;
+  }
```
walec51, line 117: it would be better to do lazy initialization as suggested in: https://github.com/knowm/XChange/wiki/Implementing-resiliency-features#step-1

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProAccountServiceRaw.java
```diff
@@ -16,116 +17,178 @@
...
-        accountId);
+    return decorateApiCall(() ->
+            coinbasePro.sendMoney(
+                    new CoinbaseProSendMoneyRequest(to, amount, currency.getCurrencyCode()),
+                    apiKey,
+                    digest,
+                    UnixTimestampFactory.INSTANCE.createValue(),
+                    passphrase,
+                    accountId))
+            .withRateLimiter(rateLimiter(PRIVATE_REST_ENDPOINT_RATE_LIMITER))
```
walec51, line 61: if this an idempotent operation?

if two request get to the server because of a retry will it not send money twice?

earce: I don't have any retries here? If there is no retry logic why would it send it again it's just a rate limiter?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
earce: My only issue with the lazy initialization is that people can create multiple instances of the resilience class, is that a property we want, I figured for the rate limiters we want a singleton?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
earce: I am not following, what do you mean, what is a yoda condition?

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/service/CoinbaseProAccountServiceRaw.java
walec51: a sorry I misunderstood the code

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
walec51: doing `"open".equals(order.getStatus())` instead of `order.getStatus().equals("open")`

https://en.wikipedia.org/wiki/Yoda_conditions

earce: Ah! that is funny I have never heard the term yoda condition. In this case we have added protection, if order.getStatus() returned null, we would do null.equals which would NPE, however by doing constant left side we will not NPE, the way I have written the function is effectively equivilant to order.getStatus() != null && order.getStatus().equals("open)

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
walec51: doing lazy initialization in the way suggested in https://github.com/knowm/XChange/wiki/Implementing-resiliency-features#step-1

```
public class BinanceExchange extends BaseExchange {

  private static ResilienceRegistries RESILIENCE_REGISTRIES;

  // ...

  @Override
  public ResilienceRegistries getResilienceRegistries() {
    if (RESILIENCE_REGISTRIES == null) {
      RESILIENCE_REGISTRIES = BinanceResilience.createRegistries();
    }
    return RESILIENCE_REGISTRIES;
  }
```

will result in a singelton - ` BinanceResilience.createRegistries` will be called only once, when the first exchange instance gets created - any more instances of the exchange will result in using the previously created ResilienceRegistries instance

walec51:
looks good, just two small remarks however they are not critical

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProExchange.java
earce: It will be a singleton from the BinanceExchange perspective, but it will not be a singleton from the JVM perspective. But this is less likely so I can change it if you prefer.

walec51: ResilienceRegistries will not be a singelton from the JVM perspective in either case 

CoinbaseProResilience is also not a singleton, its an utility class

because of this `CoinbaseProResilience.INSTANCE` just looks confusing to me

we want all instances of `XExchange` to use a shared `ResilienceRegistries` - if think that this should be expressed in the implementation of `XExchange`

xchange-coinbasepro/src/main/java/org/knowm/xchange/coinbasepro/CoinbaseProAdapters.java
walec51: if `order.getStatus()` will return null then the earlier `switch (order.getStatus()) {` will throw an NPE either way

however we can leave it as is if its not confusing to any one else

PR #3968, opened 2021-03-06
[FTX]Adding FTX fundingPayments and borrowingHistory support

makarid:
Also, adding a starting point for ftx-stream


xchange-ftx/src/main/java/org/knowm/xchange/ftx/service/FtxAccountServiceRaw.java
```diff
@@ -95,4 +95,33 @@ public FtxAccountServiceRaw(Exchange exchange) {
...
+      return ftx.getFundingPayments(
+              exchange.getExchangeSpecification().getApiKey(),
+              exchange.getNonceFactory().createValue(),
+              signatureCreator,
+              subaccount,
+              startTime,
+              endTime,
+              future);
+    } catch (FtxException e) {
+      throw new FtxException(e.getMessage());
```
earce, line 124: why don't you just throw e back?

makarid:
Because i thought that we must a create an ExchangeException for every
exchange, everytime

Στις Κυρ, 7 Μαρ 2021 στις 9:50 μ.μ., ο/η Erick Arce <
notifications@github.com> έγραψε:

xchange-ftx/src/main/java/org/knowm/xchange/ftx/service/FtxAccountServiceRaw.java
makarid: For readability mostly

xchange-stream-ftx/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+    <parent>
+        <groupId>org.knowm.xchange</groupId>
+        <artifactId>xchange-parent</artifactId>
+        <version>5.0.7-SNAPSHOT</version>
+    </parent>
+
+    <name>XChange Ftx Stream</name>
+    <artifactId>xchange-stream-ftx</artifactId>
+
+    <properties>
```
earce, line 15: I don't think we want to target java11 right now @timmolter ?

makarid: i am sorry my bad, i use Java 11 and i didn't change it. I will change it and push again

Issue #3969, opened 2021-03-08
Independent reserve implement getDynamicFee


zdary:
add account getDynamicFee method for independent reserve

zdary:
https://github.com/knowm/XChange/pull/3970

PR #3970, opened 2021-03-08
#3969 Independent Reserve add getDynamicFees

zdary:


xchange-independentreserve/src/main/java/org/knowm/xchange/independentreserve/IndependentReserveAuthenticated.java
```diff
@@ -34,10 +31,11 @@
...
 public interface IndependentReserveAuthenticated {
 
-  public static final String SynchDigitalCurrencyDepositAddressWithBlockchain =
+  String SynchDigitalCurrencyDepositAddressWithBlockchain =
       "SynchDigitalCurrencyDepositAddressWithBlockchain";
-  public static final String WithdrawDigitalCurrency = "WithdrawDigitalCurrency";
-  public static final String GetDigitalCurrencyDepositAddress = "GetDigitalCurrencyDepositAddress";
+  String WithdrawDigitalCurrency = "WithdrawDigitalCurrency";
+  String GetDigitalCurrencyDepositAddress = "GetDigitalCurrencyDepositAddress";
+  String GetBrokerageFees = "GetBrokerageFees";
```
earce, line 38: why are we getting rid of static final?

zdary: It is an interface, static final is redundant

PR #3974, opened 2021-03-09
[Ftx] Fixing authentication fail check

makarid:


xchange-ftx/src/main/java/org/knowm/xchange/ftx/service/FtxDigest.java
```diff
@@ -30,7 +30,7 @@ public String digestParams(RestInvocation restInvocation) {
             + restInvocation.getHttpMethod().toUpperCase()
             + restInvocation.getPath();
 
-    if(restInvocation.getQueryString() != null){
+    if(!restInvocation.getQueryString().isEmpty()){
```
earce, line 33: can you check if ! null in front as well

makarid: it is never null

PR #3976, opened 2021-03-11
[bitstamp] do not hardcode currencies

matejsp:
Bitstamp had hardcoded set of currencies. 
Refactored using JsonAnySetter and list of markets form remoteInit.

### DIFF
--- a/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAdapters.java
+++ b/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAdapters.java
@@ -10,6 +10,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+
 import org.knowm.xchange.bitstamp.dto.account.BitstampBalance;
 import org.knowm.xchange.bitstamp.dto.marketdata.BitstampOrderBook;
 import org.knowm.xchange.bitstamp.dto.marketdata.BitstampPairInfo;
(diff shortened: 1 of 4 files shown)


xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAdapters.java
```diff
@@ -300,80 +301,24 @@ public static UserTrades adaptTradeHistory(BitstampUserTransaction[] bitstampUse
...
-
-    if (transaction.getLink() != null && transaction.getEth() != null) return CurrencyPair.LINK_ETH;
+    String[] keys = transaction.getAmounts().keySet().toArray(new String[0]);
+    if (keys.length != 2) {
+      throw new IllegalArgumentException("Amount size is not 2. Unable to calculate currency pair.");
+    }
 
-    throw new NotYetImplementedForExchangeException();
+    if (exchangeSymbols.contains(new CurrencyPair(keys[0], keys[1]))) {
+      return new CurrencyPair(keys[0], keys[1]);
```
earce, line 312: minor memory optimization, rather then return `new CurrencyPair(keys[0], keys[1])` can you just return `exchangeSymbols.get(new CurrencyPair(keys[0], keys[1]))`

matejsp: List.get() does not accept object but index. But I have created a variable for the same purpose.

PR #3986, opened 2021-03-23
Upgrade rescu to 2.1

mmazi:
See https://github.com/mmazi/rescu/pull/135.

PR #3989, opened 2021-03-25
Remove hacky reflection based digest in binance and huobi

tonykwok1992:
rescu upgrade support Query Param digest now, so the original code which use reflection becomes a bug mentioned here
https://github.com/knowm/XChange/pull/3986#issuecomment-805707311

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceHmacDigest.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceHmacDigest.java
@@ -69,23 +69,6 @@ public class BinanceHmacDigest extends BaseParamsDigest {
       Mac mac = getMac();
       mac.update(input.getBytes("UTF-8"));
       String printBase64Binary = bytesToHex(mac.doFinal());
-
-      // https://github.com/mmazi/rescu/issues/62
-      // Seems rescu does not support ParamsDigest in QueryParam.
-      // hack to replace the signature in the invocation URL.
-      String invocationUrl = restInvocation.getInvocationUrl();
-      // String newInvocationUrl = UriBuilder.fromUri(invocationUrl).replaceQueryParam("signature",
-      // printBase64Binary).build().toString();
-
-      final String sig = "signature=";
-      int idx = invocationUrl.indexOf(sig);
-      String newInvocationUrl = invocationUrl.substring(0, idx + sig.length()) + printBase64Binary;
-      try {
-        invocationUrlField.set(restInvocation, newInvocationUrl);
-      } catch (IllegalArgumentException | IllegalAccessException e) {
-        throw new RuntimeException(e);
-      }
-
       return printBase64Binary;
     } catch (UnsupportedEncodingException e) {
       throw new RuntimeException("Illegal encoding, check the code.", e);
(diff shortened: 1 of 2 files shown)


xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceHmacDigest.java
```diff
@@ -69,23 +60,6 @@ public String digestParams(RestInvocation restInvocation) {
...
       String printBase64Binary = bytesToHex(mac.doFinal());
-
-      // https://github.com/mmazi/rescu/issues/62
-      // Seems rescu does not support ParamsDigest in QueryParam.
-      // hack to replace the signature in the invocation URL.
-      String invocationUrl = restInvocation.getInvocationUrl();
-      // String newInvocationUrl = UriBuilder.fromUri(invocationUrl).replaceQueryParam("signature",
-      // printBase64Binary).build().toString();
-
-      final String sig = "signature=";
```
walec51, line 80: how are we passing this signature now without this code?

tonykwok1992: Just like digesting other type of params, when the digested value (signature) is returned from this method it will become part of the query params. This is how the original digest interface is designed but historically it does not support Query Params as digest, as the rescu upgrade to support it, this can be simplified to just return the digested value

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceHmacDigest.java
```diff
@@ -69,23 +60,6 @@ public String digestParams(RestInvocation restInvocation) {
       Mac mac = getMac();
       mac.update(input.getBytes("UTF-8"));
```
earce, line 61: Use standardcharsets.utf8 instead of this string

xchange-huobi/src/main/java/org/knowm/xchange/huobi/service/HuobiDigest.java
```diff
@@ -54,9 +46,7 @@ public String digestParams(RestInvocation restInvocation) {
             .collect(Collectors.joining("&"));
     String toSign = String.format("%s\n%s\n%s\n%s", httpMethod, host, method, query);
     Mac mac = getMac();
-    String signature =
-        encodeValue(Base64.getEncoder().encodeToString(mac.doFinal(toSign.getBytes())).trim());
-    replaceSignatureUrl(restInvocation, signature);
+    String signature = Base64.getEncoder().encodeToString(mac.doFinal(toSign.getBytes())).trim();
```
earce, line 49: getBytes(StandardCharsets.UTF8)

tonykwok1992: done

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceHmacDigest.java
tonykwok1992: done

earce:
@mmazi do you agree with this change?

mmazi:
> @mmazi do you agree with this change?

Seems less hacky than before, indeed. I don't think I'm the one to judge this though.

tonyconv:
I think without this change all binance and huobi digest will not work after rescu upgrade, so this is not just a clean up but a bug fix

earce:
Approving

PR #3996, opened 2021-03-31
[cryptofacilities] fix for new rescu version

andre77:
the code did not work with the new rescu version 2.1
see https://github.com/knowm/XChange/pull/3986

### DIFF
--- a/xchange-cryptofacilities/src/main/java/org/knowm/xchange/cryptofacilities/service/CryptoFacilitiesDigest.java
+++ b/xchange-cryptofacilities/src/main/java/org/knowm/xchange/cryptofacilities/service/CryptoFacilitiesDigest.java
@@ -46,14 +46,7 @@ public class CryptoFacilitiesDigest extends BaseParamsDigest {
       throw new RuntimeException(
           "Illegal algorithm for post body digest. Check the implementation.");
     }
-
-    String decodedQuery = null;
-    try {
-      decodedQuery = URLDecoder.decode(restInvocation.getParamsMap().get(QueryParam.class).asQueryString(), "UTF-8");
-    } catch (UnsupportedEncodingException e) {
-      throw new IllegalArgumentException("Unsupported query encoding", e);
-    }
-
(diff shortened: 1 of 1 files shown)


xchange-cryptofacilities/src/main/java/org/knowm/xchange/cryptofacilities/service/CryptoFacilitiesDigest.java
```diff
@@ -48,7 +49,7 @@ public String digestParams(RestInvocation restInvocation) {
 
     String decodedQuery = null;
     try {
-      decodedQuery = URLDecoder.decode(restInvocation.getQueryString(), "UTF-8");
+      decodedQuery = URLDecoder.decode(restInvocation.getParamsMap().get(QueryParam.class).asQueryString(), "UTF-8");
```
earce, line 52: use standardcharsets.utf8 instead of the string

sizmestiev:
> 
> 
> This fix is based on Java 11 and broke digest creation under Java 8. Is it correct to introduce in XChange Java 11 - based code?
> 
> #2945

Sorry, Java 11 was introduced later: https://github.com/knowm/XChange/commit/6ddd2e3bd568fbf64bb80dc8c39a5c5c17a80ad7#diff-6ec4afdd46cac2fa3bfee1282876504ced1c0c5fd6258a7fc778786568c4e8db

andre77:
is it about the usage of class StandardCharsets ?
actually it was an explicit wish, see earlier comments
https://github.com/knowm/XChange/pull/3996#discussion_r604902682

sizmestiev:
> is it about the usage of class StandardCharsets ?

Yes, actually `URLDecoder.decode` with `StandardCharsets` argument was introduced in Java 11, and that point is the only one (for the whole project) incompatible with Java 8 .

andre77:
@sizmestiev sorry about that, like i said, it was an explicit wish to use it
my first commit did not contain the StandardCharsets class
feel free to create a simple PR to change it back

timmolter:
Yes, let's keep the project compatible with Java 8 for Android.

PR #3998, opened 2021-03-31
[FTX] Adding Ftx StreamOrderbook support

makarid:
I am just not sure if i have correctly dealt with error throwing on the FtxStreamingAdapter when the checksum is not correct. Any feedback would be appreciated. Thanks


xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+                    try{
+                        return mapper.readValue(res.toString(),FtxOrderbookResponse.class);
+                    }catch (IOException e){
+                        throw new RuntimeException(e);
+                    }
+                })
+                .forEach(
+                        message -> {
+
+                            if(message.getAction().equals("partial")){
```
earce, line 38: constant on left side

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+                                message.getAsks().forEach(ask->{
```
earce, line 39: lambda here and below doesn't need {} since it's single line

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+                        .append(asks.get(i).getOriginalAmount().doubleValue());
+            }
+            if(i != 99){
+                data.append(":");
+            }
+
+        }
+
+        CRC32 crc32 = new CRC32();
+        byte[] toBytes = data.toString().getBytes();
```
earce, line 89: getBytes(StandardCharsets.UTF8)

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingMarketDataService.java
```diff
@@ -1,18 +1,33 @@
...
+    public FtxStreamingMarketDataService(FtxStreamingService service) {
+        this.service = service;
+    }
 
     @Override
     public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
-        return null;
+        OrderBook orderBook = new OrderBook(null, Lists.newArrayList(), Lists.newArrayList());
+
+    return service
```
earce, line 28: tab indent return

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingMarketDataService.java
```diff
@@ -1,18 +1,33 @@
...
+    private final FtxStreamingService service;
+
+    public FtxStreamingMarketDataService(FtxStreamingService service) {
+        this.service = service;
+    }
 
     @Override
     public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
-        return null;
+        OrderBook orderBook = new OrderBook(null, Lists.newArrayList(), Lists.newArrayList());
```
earce, line 26: please just new new ArrayList<>() no need to introduce google deps here

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+                            if(orderBook.getAsks().size() > 0 && orderBook.getBids().size() > 0){
+                                Long calculatedChecksum = getOrderbookChecksum(orderBook.getAsks(),orderBook.getBids());
+
+                                if(!calculatedChecksum.equals(message.getChecksum())){
+                                    throw new RuntimeException("Checksum is not correct!");
+                                }
+                            }
+                        });
+
+        return new OrderBook(Date.from(Instant.now()), Lists.newArrayList(orderBook.getAsks()),Lists.newArrayList(orderBook.getBids()),true);
```
earce, line 64: new ArrayList<>(orderbook.getasks()) no need for extra google dep

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+    private final static ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
+
+    public static OrderBook adaptOrderbookMessage(OrderBook orderBook, Instrument instrument, JsonNode jsonNode){
+
+        Streams.stream(jsonNode)
+                .filter(JsonNode::isObject)
+                .map(res->{
+                    try{
+                        return mapper.readValue(res.toString(),FtxOrderbookResponse.class);
+                    }catch (IOException e){
```
earce, line 31: space after try { bracket and before catch bracket

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+
+                            if(message.getAction().equals("partial")){
+                                message.getAsks().forEach(ask->{
+                                    orderBook.getAsks().add(new LimitOrder.Builder(Order.OrderType.ASK,instrument).limitPrice(ask.get(0)).originalAmount(ask.get(1)).build());
+                                });
+
+                                message.getBids().forEach(bid ->{
+                                    orderBook.getBids().add(new LimitOrder.Builder(Order.OrderType.BID,instrument).limitPrice(bid.get(0)).originalAmount(bid.get(1)).build());
+                                });
+                            }else{
```
earce, line 46: space before else bracket

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+                            }
+                        });
+
+        return new OrderBook(Date.from(Instant.now()), Lists.newArrayList(orderBook.getAsks()),Lists.newArrayList(orderBook.getBids()),true);
+    }
+
+    public static Long getOrderbookChecksum(List<LimitOrder> asks,List<LimitOrder> bids){
+        StringBuilder data = new StringBuilder();
+
+        for (int i=0;i<100;i++){
```
earce, line 70: spaces between for conditionals

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -0,0 +1,94 @@
...
+                                    throw new RuntimeException("Checksum is not correct!");
+                                }
+                            }
+                        });
+
+        return new OrderBook(Date.from(Instant.now()), Lists.newArrayList(orderBook.getAsks()),Lists.newArrayList(orderBook.getBids()),true);
+    }
+
+    public static Long getOrderbookChecksum(List<LimitOrder> asks,List<LimitOrder> bids){
+        StringBuilder data = new StringBuilder();
```
earce, line 68: can you get an estimation of the string size when complete, and find an approximation and make the default size that, a single large allocation is more efficient then starting at 16 growing, allocating another array, gc'ing and copying over to new array etc.

makarid: Do you mean the StringBuilder to have a fixed size?

earce: stringbuilder will never have a fixed size, it's just about its initial size, you want to have an initial size that it won't have to create a new backing array during the creation process, if you see the default starting size is 16

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingMarketDataService.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingMarketDataService.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: done

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
makarid: you mean to do the following: 

StringBuilder data = new StringBuilder(16);

earce: No, that is what happens by default, 16 is the default size.

The way this works under the hood the stringbuilder is backed by an array of size 16, so if you say, sb.append("test") the array of 16 now has 4 spaces used, if you sb.append("111222333444555") it will not fit all those characters in this array since total qty is 19. 

What happens now? That array is thrown away and a new one of size 32 is allocated and all the contents are copied over. This means everytime the stringbuilder has enough contents to exceed the backing array's size, it 1) creates a new array of size 2*prev_size 2) copies the contents over 3) gc's the old contents.

This is an expensive thing to do if we know what the final array size or at least a better sense of the final size of the array will be so rather then do this you can just say something like new StringBuilder(2048) or a number that is close to the final size of the string and only allocate one time.

Does this make sense?

makarid: Thank you very much for the explanation! The thing is that every currencyPair has different prices and even the volume of every currencyPair has different scale settings. So, i should either put an extreme number like 5000000 or just leave it as it is. The checksum gets the first 100 prices on both sides(ask and bid) and adds them separated by a colon. So, it is a very big String. What do you thing?

earce: You can probably do a reasonable estimation of size that will give you a good starting point:

100*(8 + 1 + 6)
8 = len("60000.00") (price)
1 = ":"
5 = len("12.01") (qty) = 2400

So to be slightly aggressive = 3072 works? if it uses less nbd it will get collected if it uses more same thing it will only require maybe one more copy

makarid: Great, thanks!

PR #4018, opened 2021-04-18
[Bittrex] Place Market Order

garciapd:
Implement place market order for Bittrex


xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/BittrexAdapters.java
```diff
@@ -62,26 +75,48 @@ public static CurrencyPair adaptCurrencyPair(BittrexSymbol bittrexSymbol) {
...
-    return new LimitOrder.Builder(type, pair)
-        .originalAmount(order.getQuantity())
-        .id(order.getId())
-        .timestamp(order.getUpdatedAt() != null ? order.getUpdatedAt() : order.getCreatedAt())
-        .limitPrice(order.getLimit())
-        .remainingAmount(order.getQuantity().subtract(order.getFillQuantity()))
-        .fee(order.getCommission())
-        .orderStatus(status)
-        .build();
+    if (order.getType().equalsIgnoreCase(BittrexConstants.MARKET)) {
```
earce, line 97: put constant on left side

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/BittrexAdapters.java
```diff
@@ -62,26 +75,48 @@ public static CurrencyPair adaptCurrencyPair(BittrexSymbol bittrexSymbol) {
...
+  }
+
+  public static LimitOrder adaptLimitOrder(BittrexOrder order, OrderStatus status) {
+    return (LimitOrder) adaptOrder(order, status);
+  }
+
+  public static Order adaptOrder(BittrexOrder order, OrderStatus status) {
 
     OrderType type =
         order.getDirection().equalsIgnoreCase(BittrexConstants.SELL)
```
earce, line 93: constant on left side

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
```diff
@@ -106,10 +123,25 @@ public OpenOrdersParams createOpenOrdersParams() {
     try {
       List<Order> orders = new ArrayList<>();
       for (String orderId : orderIds) {
-        BittrexOrder order = getBittrexOrder(orderId);
-        if (order != null) {
-          LimitOrder limitOrder = BittrexAdapters.adaptOrder(order);
-          orders.add(limitOrder);
+        BittrexOrder bittrexOrder = getBittrexOrder(orderId);
+        if (bittrexOrder != null) {
```
earce, line 127: rather then nest this so much can you just return on the negation. e.g. 
```
if (bittrexOrder != null) {
    continue;
}
if (!(order instanceof MarketOrder)) {
    continue;
}
```

This makes following the code a bit easier to read.

Also instead of executions.size() use !isEmpty();

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
```diff
@@ -106,10 +123,25 @@ public OpenOrdersParams createOpenOrdersParams() {
...
+          if (order instanceof MarketOrder) {
+            BigDecimal rate =
+                decorateApiCall(
+                        () -> {
+                          List<BittrexExecution> executions =
+                              getBittrexOrderExecutions(order.getId());
+                          if (executions != null && executions.size() > 0) {
+                            return executions.get(0).getRate();
+                          }
+                          throw new BittrexException();
```
earce, line 138: do you want to trigger this exception and cause this to use the 'withRetry' this way?

earce:
You didn't address 2/3 of my comments?

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
```diff
@@ -132,7 +132,7 @@ public OpenOrdersParams createOpenOrdersParams() {
                         () -> {
                           List<BittrexExecution> executions =
                               getBittrexOrderExecutions(order.getId());
-                          if (executions != null && executions.size() > 0) {
+                          if (executions != null && !executions.isEmpty()) {
```
garciapd, line 135: did that one

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/BittrexAdapters.java
```diff
@@ -90,11 +90,11 @@ public static LimitOrder adaptLimitOrder(BittrexOrder order, OrderStatus status)
   public static Order adaptOrder(BittrexOrder order, OrderStatus status) {
 
     OrderType type =
-        order.getDirection().equalsIgnoreCase(BittrexConstants.SELL)
+            BittrexConstants.SELL.equalsIgnoreCase(order.getDirection())
```
garciapd, line 93: and add constants at the beginning

garciapd (commented):
> You didn't address 2/3 of my comments?

the only thing I haven't changed is the loop and the exception flow. I put the constant at the beginning and change the .isEmpty().

For the exception, I think it should work, it will retry 3 times and stop, it is working for me at the moment.

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
earce: do you see what I'm asking there I pasted code here? Instead of nesting the if statements, just continue so you don't need to indent so much, makes the code more readable.

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
```diff
@@ -126,15 +127,19 @@ public OpenOrdersParams createOpenOrdersParams() {
...
-            }
+            BigDecimal rate =
+                decorateApiCall(
+                        () -> {
+                          List<BittrexExecution> executions =
+                              getBittrexOrderExecutions(order.getId());
+                          if (executions != null && executions.size() > 0) {
+                            return executions.get(0).getRate();
+                          }
+                          throw new BittrexException();
```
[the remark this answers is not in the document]
earce: I think you may probable have a do while type of statement or a for-each with you retry count, otherwise what you are trying to do is a bit obfuscated and unclear. Throwing exceptions to try and make something happen isn't really the point of exceptions, your looping logic should encompass that.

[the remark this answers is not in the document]
garciapd: what about now? when a market order is performed, the price is not coming from the getBittrexOrder, the retry is because if you just placed the order, it can take a few milliseconds for the execution to be created, in that case, the loop won't find anything, it will break and the retry will fetch again.

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
garciapd: I have added the continue when checking if the bittrexOrder is null, I agree it is more readable, for checking if the order is a MarketOrder, I think it is good like that, because fetching the price only has to be done for MarketOrders, otherwise just add to the list

garciapd:
@earce is it ok now? Thanks for reviewing

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeService.java
```diff
@@ -106,11 +125,29 @@ public OpenOrdersParams createOpenOrdersParams() {
...
+                        return executions.stream()
+                            .map(BittrexExecution::getRate)
+                            .reduce(BigDecimal.ZERO, BigDecimal::add)
+                            .divide(
+                                new BigDecimal(executions.size()),
+                                new MathContext(8, RoundingMode.FLOOR));
+                      })
+                  .withRetry(retry("orderExecutionRate"))
+                  .call();
+          order.setAveragePrice(rate);
```
earce, line 148: Sorry for being pedantic, trying to keep the code very readable here, do you think you could do something like 

```
          List<BittrexExecution> executions = 
              decorateApiCall(() -> getBittrexOrderExecutions(order.getId()))
              .withRetry(retry("orderExecutionRate"))
              .call();
          BigDecimal rate = executions.stream()
              .map(BittrexExecution::getRate)
              .reduce(BigDecimal.ZERO, BigDecimal::add)
              .divide(
                      new BigDecimal(executions.size()),
                      new MathContext(8, RoundingMode.FLOOR));
          order.setAveragePrice(rate);
```
To help with separating what is the actual API call vs Xchange business logic

garciapd:
I had a fix for bitfinex too mixed in the branch, could you please check if the code as it is good enough? i separated the api call from the logic. Bittrex sometimes takes a few millis to get you the order with the price, that is why the retries...

timmolter:
I'd merge this expect for the retries. That can be done outside of this lib.

walec51:
it would be best for now to ommit those retries or implement them in this fashion https://github.com/knowm/XChange/wiki/Implementing-resiliency-features

however does the originam author have the time to polish this PR?

Issue #4021, opened 2021-04-21
Bitfinex init exception 


IhorBr:
Caused by: java.lang.IllegalStateException: Duplicate key DOGE/USD (attempted merging values 0.094716 and 0.30776)
        at java.base/java.util.stream.Collectors.duplicateKeyException(Collectors.java:133)
        at java.base/java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:180)
        at java.base/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
        at org.knowm.xchange.bitfinex.BitfinexExchange.remoteInit(BitfinexExchange.java:85)
        at org.knowm.xchange.BaseExchange.applySpecification(BaseExchange.java:113)
        at org.knowm.xchange.ExchangeFactory.createExchange(ExchangeFactory.java:80)
        at org.knowm.xchange.ExchangeFactory.createExchange(ExchangeFactory.java:38)

cyrus13:
This is because Bitfinex added DOGE/USD and it also has a pair: XDC/USD which appears to be a different coin (https://coinmarketcap.com/currencies/xinfin-network/).

The problem is in this part of code in Currency.java:

`  public static final Currency DOGE = createCurrency("DOGE", "Dogecoin", null, "XDC", "XDG");`

that maps both DOGEUSD and XDCUSD to the same pair.

Apparently XDCUSD is the DOGEUSD pair for Kraken. So, it appears that we need to make a different mapping for Bitfinex. Question to the dev team: Have you come across a similar problem in another exchange so that we can follow the same solution?

IhorBr:
Thanks for the answer.
[...]

PR #4023, opened 2021-04-22
Fix for Issue #4021 Bitfinex init exception and #4024 [Bitstamp] Some Currency Metadata and CurrencyPair Metadata are null

mathvdh:
As discussed in the issue, it's a simple fix that

Defines Doge as DOGE or XDG (not XDC)
And add a Currency for XinFin Network called XDC

As you can see on CMC : https://coinmarketcap.com/currencies/xinfin-network/
XDC is the official symbol of XinFin Network

And you can see on https://www.kraken.com/prices?quote=USD that the Kraken symbol for Doge is XDG (not XDC)

### DIFF
--- a/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAdapters.java
+++ b/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAdapters.java
@@ -10,6 +10,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


mathvdh:
This also fixes #4024 

The reason I have deleted all the currency_pairs from bitstamp.json is that this information was just **incorrect** the minimum amounts where false, because bitstamp check the minimal amounts using the quote or counter currency, which means that putting base amounts as minimal amounts is just false as the price varies.

Moreover the correct information is now taken from Bitstamp APIs during the remote init.

mathvdh:
If it's really a problem I can do 2 different PRs for the fixes ... let me know

PR #4031, opened 2021-04-25
[ftx] remove broken order book checksum

### LINKED ISSUE #4014: Stream FTX : Checksum is not correct!
Hi,

I try the stream with FTX (code is like the example) and this part cause an issue : Checksum is not correct!

class FtxStreamingAdapters

                if (!calculatedChecksum.equals(message.getChecksum())) {
                  throw new RuntimeException("Checksum is not correct!");
                }
If i comment this lines (88/89/90), it's worked. I receive the orderbook.

phraktle:
Checksumming the FTX order book is currently broken. Fixing it would require storing the original values verbatim to retain number formatting. It's also very wasteful to do this on every change, so I suggest removing this completely. Fixes  #4014.

### DIFF
--- a/xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
+++ b/xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
@@ -6,12 +6,9 @@ import com.google.common.collect.Streams;
 import info.bitrich.xchangestream.ftx.dto.FtxOrderbookResponse;
 import info.bitrich.xchangestream.service.netty.StreamingObjectMapperHelper;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Date;
-import java.util.List;
-import java.util.zip.CRC32;
 import org.knowm.xchange.dto.Order;
 import org.knowm.xchange.dto.marketdata.OrderBook;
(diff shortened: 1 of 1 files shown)


bitcoinmat-dev:
Ok, but how will we know that the orderbook is correct? Thanks

phraktle:
> Ok, but how will we know that the orderbook is correct? Thanks

how do you know in any of the other XChange streaming client implementations? :)

if you feel like taking a crack at a correct implementation (which is also optional for performance reasons), i'm in favor. in the meantime, the implementation removed by this PR is fundamentally flawed and stops streaming from working at all.

earce:
@phraktle can you make get orderbook checksum just extract the current checksum that is part of the message, then when you try to compare it, it will pass, and add a TODO on the method describing what needs to be done so it's understood as something to potentially implement later

phraktle:
@earce I wouldn’t consider this a “to do” item.

earce:
I would disagree https://docs.ftx.com/#orderbooks, ensuring your client is in sync with server should be part of implementation if the exchange has it.

earce:
I will not approve this without a TODO, you will have to find someone else to merge it.

mdvx:
can we merge this in?

mdvx:
https://github.com/knowm/XChange/pull/4151/commits/20154ce6504cd077d219888deb1f8e5e7737f254

makarid:
Hello, please do not merge this. Right now the checksum is working without any issue.

PR #4032, opened 2021-04-26
[TradeOgre] more endpoints implemented

cymp:
private: buy, sell, getOrders
public: getOrderBook

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -340,7 +340,7 @@
       <dependency>
         <groupId>com.pubnub</groupId>
         <artifactId>pubnub-gson</artifactId>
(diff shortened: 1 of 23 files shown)


pom.xml
```diff
@@ -9,7 +9,7 @@
 
   <groupId>org.knowm.xchange</groupId>
   <artifactId>xchange-parent</artifactId>
-  <version>5.0.8-SNAPSHOT</version>
+  <version>5.0.2-SNAPSHOT</version>
```
earce, line 12: you are downgrading snapshot number and versions of dependencies?

Issue #3991, opened 2021-03-26
bitstamp v2 API should be taking UUID nonces


matt-nydig:
It appears that methods that call the Bitstamp v2 api are sending `Long` nonces rather than UUID nonces. This breaks with a nonce error much of the time.

Some background: I just upgraded some code that fetches balances from Bitstamp to xchange 5.0.6, which includes a change to use the v2 API for fetching balances: https://github.com/knowm/XChange/commit/dd134e31dd50663b30bb61ae19150e7b6d387e0f

Shortly after the change, we started getting 403's with an "Invalid nonce" message. I've taken a (newbie, not-too-detailed) look at your code, and it _looks_ like the nonce generation algorithm used for Bitstamp's v2 api still relies on sending a `Long` nonce, rather than a UUID.

Falling back to xchange 5.0.4 immediately fixed this error.

earce:
Hi Matt,

Taking a glance at the Bitstamp API and I see that the V2 API does in fact have some significant changes from the V1 as far as expected nonce. We will probably need to create a custom Bitstamp nonce generator because the documentation has some pretty specific requirements namely:

- lowercase,
- 36 char string,
- each nonce can be used only once within a timeframe of 150 seconds.

I would think using Java's UUID.randomUUID().toString() would satisfy what this API wants. It returns 36 length string. If not writing a custom random string generator isn't too difficult. You do have the extremely unlikely event of a collision but a strong guarantee could be to create a map using something like guava or caffeine to evict nonces after they exceed 150 seconds in age. I think in most cases Java's UUID is probably good enough.

I don't know who else is using this right now but if you'd like you can definitely submit a PR for this and I can review it or offer further guidance.

matt-nydig:
`UUID.randomUUID()` uses, according to Java's docs, a "a cryptographically strong pseudo random number generator." This implies 128 bits of high entropy data. There is no remotely realistic possibility, under these assumptions, and given the rate limits of the Bitstamp API, of generating a collision within 150 seconds. We should have a look at that... I'm a bit time constrained myself right now, though.

earce:
Makes sense, sounds good to me!

joakjo:
Bitstamp will also stop supporting HTTP API v1 later this year, which will require using the new v2 authentication (header based as opposed to sending key, signature and nonce as request parameters which both `BitstampAuthenticated` and `BitstampAuthenticatedV2` do atm)

[DEPRECATION WARNING
[...]

PR #4038, opened 2021-04-28
Fix for #3991 bitstamp v2 API should be taking UUID nonces

mathvdh:
Hi All,

So this is a fix following the discussions in #3991 

It was a bit more complex than expected, here are some of the subtilities.

- We need both a UUID Nonce Factory and a millisecond time stamp (as string) factory because both are required in the requests.
- I had to write a new `BitstampDigestV2` because it changed completely from V1.
- All the authentication stuff is now in HeaderParams and not FormParams
- The content-type CANNOT be present if there is no body to the requests, hence the move of `@Consumes(MediaType.APPLICATION_FORM_URLENCODED)` from the top to the individual functions (when necessary). It's also taken into account into the `BitstampDigestV2` which includes it or doesn't include it for the signing process.
- Introduction of `apiKeyForV2Requests` as the api key for requests (and signing) must now be `"BITSTAMP " + apiKey`.

I also cleaned/standardized some stuff when different calls where using `this.apiKey` or `exchange.getExchangeSpecification().getApiKey()`

Otherwise it passes all my tests (with a real bitstamp account behind) namely :

- get funding history
- get open orders
- get order books
- get balances
- place and cancel limit order
- get pairs list (and meta)
- get currencies list (and meta)

(Which is all the functionalities I'm using in my app)

All the xchange standard tests are also passing (with some deprecation notices for bitstamp v1 obviously)

Eventually we will have to get rid of all the V1 stuff and this would bring some nice cleaning (I didn't find when exactly the V1 will stop working on bitstamp thou .. I'm not exactly sure what functionality is still missing from V2 in comparison to V1 and if it will be possible to emulate it or just remove it.

You can find documentation about the new authentication method [here](https://www.bitstamp.net/api/#api-authentication
)

Let me know if it's all OK for you ?

### DIFF
--- a/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampDigestV2.java
+++ b/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampDigestV2.java
@@ -1,0 +1,46 @@
+package org.knowm.xchange.bitstamp.service;
+
+import java.math.BigInteger;
+import javax.crypto.Mac;
+import javax.ws.rs.FormParam;
+import org.knowm.xchange.service.BaseParamsDigest;
+import si.mazi.rescu.RestInvocation;
+
+/** @author Benedikt Bünz */
+public class BitstampDigest extends BaseParamsDigest {
+
+  private final String clientId;
+  private final String apiKey;
+
+  /**
+   * Constructor
+   *
+   * @param secretKeyBase64
+   * @param clientId
+   * @param apiKey @throws IllegalArgumentException if key is invalid (cannot be base-64-decoded or
+   *     the decoded key is invalid).
+   */
+  private BitstampDigest(String secretKeyBase64, String clientId, String apiKey) {
+
+    super(secretKeyBase64, HMAC_SHA_256);
+    this.clientId = clientId;
+    this.apiKey = apiKey;
+  }
+
+  public static BitstampDigest createInstance(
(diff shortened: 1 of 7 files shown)


xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampDigestV2.java
```diff
@@ -0,0 +1,76 @@
...
+    return secretKeyBase64 == null ? null : new BitstampDigestV2(secretKeyBase64, apiKey);
+  }
+
+  @Override
+  public String digestParams(RestInvocation restInvocation) {
+
+    Mac mac256 = getMac();
+    mac256.update(apiKey.getBytes());
+
+    System.out.println("restInvocation.getHttpMethod() " + restInvocation.getHttpMethod());
```
earce, line 41: please remove all these System.out.println

earce:
Thank you for the effort on this, this is awesome, I have just one comment.

xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampDigestV2.java
mathvdh: Oups sorry forgot about that :)
Will do now ...

mathvdh:
Voilà I removed all the debug prints ...
I'm happy I can help, this project is great and has helped me a lot, so it's cool if I can contribute back ..

PR #4042, opened 2021-04-29
[Binance] Added sandbox support 

mrmx:
Added sandbox support for Binance exchange/streaming via exchangeSpecification.setExchangeSpecificParametersItem("Use_Sandbox", true)

The Binance spot testnet is located at https://testnet.binance.vision

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
@@ -78,7 +78,25 @@ public class BinanceExchange extends BaseExchange {
     AuthUtils.setApiAndSecretKey(spec, "binance");
     return spec;
   }
+  
+  @Override
+  public void applySpecification(ExchangeSpecification exchangeSpecification) {
+    concludeHostParams(exchangeSpecification);
+    super.applySpecification(exchangeSpecification);
+  }
 
+  /**
+   * Adjust host parameters depending on exchange specific parameters
+   */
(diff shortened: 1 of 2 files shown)


vaseok10:
Hi!
Thanks for update!
When we create standard Binance connection we use getDefaultExchangeSpecification(), setApiKey and setSecretKey to insatiate.
Do we need some other settings beside setExchangeSpecificParametersItem when we are instantiating testnet? For now it's throwing 404

earce:
Not sure, you will need to see if the api tells you more and make you upgrade to the latest build version

vaseok10:
I think the problem is in signature creator(org.knowm.xchange.binance.service.BinanceHmacDigest#getQuery)
Binance requires signature in path, and here I see it get's filtered   .filter(e -> !BinanceAuthenticated.SIGNATURE.equals(e.getKey())) 

When initializing the bean he is making some calls to get some data, which are failing

vaseok10:
```
updated to latest build, using this to instantiate and failing :(
        ExchangeSpecification exchangeSpecification = new BinanceExchange().getDefaultExchangeSpecification();
exchangeSpecification.setApiKey("XXX");
        exchangeSpecification.setSecretKey("XXX");
exchangeSpecification.setExchangeSpecificParametersItem("Use_Sandbox", true);
return ExchangeFactory.INSTANCE.createExchange(exchangeSpecification);

```

mrmx:
That error 404 is because in the testnet there is no wapi/v3/assetDetail.html endpoint present when the remoteInit method is called (from createExchange). At the moment there is no workaround.

Issue #4028, opened 2021-04-24
Coinbasepro: MarketOrder incorrect (currency pair and type)


straumat:
_The first listed currency of a currency pair is called the base currency, and the second currency is called the quote currency. When you buy a currency pair, you buy the base currency and sell the quote currency so when you are making a buy on BTC/EUR, you should be getting BTC for euros._

It works this way with other exchanges (like kucoin for example) but, with Coinbasepro things don't work this way.

This is the test I'm running on Coinbase sandbox:
- Buying 10 BTC with Euro.
- Selling 20 BTC for euro

The result is:
- 10 euro bought.
- 1 BTC sold.

This is my code:
```
        final CurrencyPair cp = new CurrencyPair(Currency.BTC, Currency.EUR);

        // Creates a buy market order.
        MarketOrder buyingOrder = new MarketOrder(Order.OrderType.BID,    // Buying order.
                new BigDecimal("10"), // "10".
                cp);    // BTC / EUR.
        final String buyingOrderId = tradeService.getXChangeService().placeMarketOrder(buyingOrder);
        System.out.println("Buying 10 BTC with Euro, order number " + buyingOrderId);

        // Creates a buy market order.
        MarketOrder sellingOrder = new MarketOrder(Order.OrderType.ASK,    // Selling order.
                new BigDecimal("1"), // "1".
                cp);    // BTC / EUR.
        final String sellingOrderId = tradeService.getXChangeService().placeMarketOrder(sellingOrder);
        System.out.println("Selling 1 BTC for Euro, order number " + sellingOrderId);
```

The result:
```
Buying 10 BTC with Euro, order number b1d0b278-659e-4f96-83bc-583222e35f94
Selling 1 BTC for Euro, order number 7a41c749-0fb9-4063-8d72-ee3eb5017f6a
```

And this is the result in Coinbasepro Sandbox:
![coinbase](https://user-images.githubusercontent.com/536498/115972277-adb02880-a54d-11eb-881c-1b7a411ecdd3.png)

How come 10 euros was bought? instead of buying BTC ?

straumat:
In the coinbase [place a new order documentation](https://docs.pro.coinbase.com/#place-a-new-order) I read this:
_FUNDS
[...]

PR #4052, opened 2021-05-05
Fix for issue https://github.com/knowm/XChange/issues/4028

straumat:
Details of the issue here [4028](https://github.com/knowm/XChange/issues/4028)


earce:
We will need to add this into release notes

PR #4055, opened 2021-05-08
add GateIo function "get_klines"

petrovMA:
This small PR adds "getKline" function for GateIo exchange.

### DIFF
--- a/xchange-gateio/src/main/java/org/knowm/xchange/gateio/dto/marketdata/GateioCandlestickHistory.java
+++ b/xchange-gateio/src/main/java/org/knowm/xchange/gateio/dto/marketdata/GateioCandlestickHistory.java
@@ -20,18 +20,15 @@ public class GateioCandlestickHistory extends GateioBaseResponse {
   }
 
   public List<List<String>>  getCandlesticks() {
-
     return candlesticks;
   }
 
   public String getElapsed() {
-
     return elapsed;
   }
 
   @Override
   public String toString() {
-
     return "BTERPublicTrades [candlesticks=" + candlesticks + ", elapsed=" + elapsed + "]";
   }
 }
--- a/xchange-gateio/src/main/java/org/knowm/xchange/gateio/dto/marketdata/GateioCandlestickHistory.java
+++ b/xchange-gateio/src/main/java/org/knowm/xchange/gateio/dto/marketdata/GateioCandlestickHistory.java
@@ -1,0 +1,37 @@
+package org.knowm.xchange.gateio.dto.marketdata;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.knowm.xchange.gateio.dto.GateioBaseResponse;
+import java.util.List;
+
+public class GateioCandlestickHistory extends GateioBaseResponse {
+
+  private final List<List<String>> candlesticks;
+  private final String elapsed;
+
+  private GateioCandlestickHistory(
+      @JsonProperty("data") List<List<String>> candlesticks,
+      @JsonProperty("result") boolean result,
+      @JsonProperty("elapsed") String elapsed) {
+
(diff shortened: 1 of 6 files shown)


xchange-gateio/src/main/java/org/knowm/xchange/gateio/dto/marketdata/GateioCandlestickHistory.java
```diff
@@ -0,0 +1,37 @@
...
+      @JsonProperty("result") boolean result,
+      @JsonProperty("elapsed") String elapsed) {
+
+    super(result, null);
+    this.candlesticks = candlesticks;
+    this.elapsed = elapsed;
+  }
+
+  public List<List<String>>  getCandlesticks() {
+
```
earce, line 23: get rid of these extra spaces here and below

xchange-gateio/src/main/java/org/knowm/xchange/gateio/service/GateioMarketDataServiceRaw.java
```diff
@@ -106,4 +105,35 @@ public GateioTradeHistory getBTERTradeHistorySince(
     List<CurrencyPair> currencyPairs = new ArrayList<>(bter.getPairs().getPairs());
     return currencyPairs;
   }
+
+  public List<GateioKline> getKlines(CurrencyPair pair, GateioKlineInterval interval, int limit) throws IOException {
+
```
earce, line 110: i think if the limit is a negative you should just throw, the user should exception if a value like that is passed in not absorb it and turn it to 1

xchange-gateio/src/main/java/org/knowm/xchange/gateio/service/GateioMarketDataServiceRaw.java
```diff
@@ -106,4 +105,35 @@ public GateioTradeHistory getBTERTradeHistorySince(
...
+    limit = limit < 0 ? 1 : limit;
+
+    long limitHours = (interval.getSeconds() * limit) / HOURS.toSeconds(1);
```
earce, line 113: if what you need is always hours just pass in limit hours instead of doing all this?

xchange-gateio/src/main/java/org/knowm/xchange/gateio/dto/marketdata/GateioCandlestickHistory.java
petrovMA: Done

xchange-gateio/src/main/java/org/knowm/xchange/gateio/service/GateioMarketDataServiceRaw.java
petrovMA: Ok, exception added.

xchange-gateio/src/main/java/org/knowm/xchange/gateio/service/GateioMarketDataServiceRaw.java
petrovMA: Ok, 'limit' parameter replaced by 'hours' parameter.
Also added method without 'hours' parameter.

earce: what is the point of that second method?

petrovMA: Just for info that it's not necessary to put parameter 'hours'.

earce: can we just eliminate it? I can approve once you do, not really worth having the function will exception always?

PR #4063, opened 2021-05-12
Added checksum checking of kraken streaming orderbooks

nielsdraaisma:
Fixes https://github.com/knowm/XChange/issues/4080


xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -155,7 +185,7 @@ private static BigDecimal nextNodeAsDecimal(Iterator<JsonNode> iterator) {
     if (iterator == null || !iterator.hasNext()) {
       return null;
     }
-    return new BigDecimal(iterator.next().textValue()).stripTrailingZeros();
+    return new BigDecimal(iterator.next().textValue());
```
earce, line 188: why are we removing this striptrailingzeros?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -19,58 +14,93 @@
...
-            currentNode -> {
-              for (String key : BID_KEYS) {
-                if (currentNode.has(key)) {
-                  adaptLimitOrders(instrument, Order.OrderType.BID, currentNode.get(key))
-                      .forEach(orderBook::update);
+    public static OrderBook adaptOrderbookMessage(int depth, AtomicBoolean awaitingSnapshot, TreeSet<LimitOrder> bids, TreeSet<LimitOrder> asks, Instrument instrument, ArrayNode arrayNode) {
+        final AtomicLong expectedChecksum = new AtomicLong(0);
+        final AtomicReference<Date> lastTime = new AtomicReference<>(Date.from(Instant.EPOCH));
+        arrayNode.elements().forEachRemaining(currentNode -> {
+            if (awaitingSnapshot.get()) {
```
earce, line 61: do we expect a scenario where we get updates before the snapshot? why do we need this await

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingChecksum.java
```diff
@@ -0,0 +1,35 @@
...
+    public static String createCrcString(TreeSet<LimitOrder> asks, TreeSet<LimitOrder> bids) {
+        StringBuilder stringBuilder = new StringBuilder();
+        asks.stream().limit(10).forEach(o -> addOrderToCrcString(stringBuilder, o));
+        bids.stream().limit(10).forEach(o -> addOrderToCrcString(stringBuilder, o));
+        return stringBuilder.toString();
+    }
+
+    public static long createCrcLong(String crcString) {
+        CRC32 crc = new CRC32();
+        crc.update(crcString.getBytes());
```
earce, line 28: getBytes(StandardCharsets.UTF8)?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingChecksum.java
```diff
@@ -0,0 +1,35 @@
...
+
+    static void addOrderToCrcString(StringBuilder stringBuilder, LimitOrder order) {
+        addBigDecimalToCrcString(stringBuilder, order.getLimitPrice());
+        addBigDecimalToCrcString(stringBuilder, order.getOriginalAmount());
+    }
+
+    public static String createCrcString(TreeSet<LimitOrder> asks, TreeSet<LimitOrder> bids) {
+        StringBuilder stringBuilder = new StringBuilder();
+        asks.stream().limit(10).forEach(o -> addOrderToCrcString(stringBuilder, o));
+        bids.stream().limit(10).forEach(o -> addOrderToCrcString(stringBuilder, o));
```
earce, line 22: is the checksum only on top 10

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
nielsdraaisma: Removing the trailing zero's modifies the bigdecimal's internals, making it impossible to use them directly in the checksum code.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
nielsdraaisma: that logic is in place to ensure it doesnt keep applying snapshots when the state is cleared, that would result in a subscribe loop.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingChecksum.java
nielsdraaisma: Correct, as described on https://docs.kraken.com/websockets/#book-checksum .

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -19,58 +14,93 @@
...
+            }
+            if (!awaitingSnapshot.get() && currentNode.has(CHECKSUM)) {
+                expectedChecksum.set(currentNode.get(CHECKSUM).asLong());
+            }
+        });
+
+        OrderBook result = new OrderBook(lastTime.get(), Lists.newArrayList(asks), Lists.newArrayList(bids), true);
+        long localChecksum = createCrcChecksum(asks, bids);
+        if (expectedChecksum.get() > 0 && expectedChecksum.get() != localChecksum) {
+            LOG.warn("Checksum does not match, expected {} but local checksum is {}", expectedChecksum.get(), localChecksum);
```
wlk, line 86: Is it possible to log information which order book has checksum mismatch? IMO would be helpful

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingChecksum.java
earce: can you update this?

nielsdraaisma: Done.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
nielsdraaisma: This has been simplified a bit by just using empty bid/ask maps as 'awaiting snapshot' indicator.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -19,58 +15,96 @@
...
+            if (!awaitingSnapshot && currentNode.has(CHECKSUM)) {
+                expectedChecksum.set(currentNode.get(CHECKSUM).asLong());
+            }
+        });
+        if ( awaitingSnapshot && bids.isEmpty() && asks.isEmpty()){
+            LOG.info("Ignoring {} message {}, awaiting snapshot", instrument, arrayNode);
+        }
+//        if ( (Math.random() > 0.99)) {
+//            throw new IllegalStateException("Testing");
+//        }
```
earce, line 86: remove this

xchange-stream-bankera/src/main/java/info/bitrich/xchangestream/bankera/BankeraStreamingService.java
```diff
@@ -26,7 +26,7 @@ public String getSubscribeMessage(String channelName, Object... args) throws IOE
   }
 
   @Override
-  public String getUnsubscribeMessage(String channelName) throws IOException {
+  public String getUnsubscribeMessage(String channelName, Object... args) throws IOException {
```
earce, line 29: can you break out changing this signature into a separate review?

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -19,58 +15,96 @@
...
-  }
+                if (currentNode.has(ASK_UPDATE)) {
+                    updateInBook(depth, instrument, Order.OrderType.ASK, currentNode, ASK_UPDATE, asks);
+                }
+            }
+            if (!awaitingSnapshot && currentNode.has(CHECKSUM)) {
+                expectedChecksum.set(currentNode.get(CHECKSUM).asLong());
+            }
+        });
+        if ( awaitingSnapshot && bids.isEmpty() && asks.isEmpty()){
```
earce, line 81: remove spacing on left

earce: also why do you need the additional awaiting snapshot boolean check

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingMarketDataService.java
```diff
@@ -30,16 +31,32 @@ public KrakenStreamingMarketDataService(KrakenStreamingService service) {
...
-                KrakenStreamingAdapters.adaptOrderbookMessage(orderBook, currencyPair, arrayNode));
-  }
+    @Override
+    public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
+      String channelName = getChannelName(KrakenSubscriptionName.book, currencyPair);
+      TreeSet<LimitOrder> bids = Sets.newTreeSet();
+      TreeSet<LimitOrder> asks = Sets.newTreeSet();
+      Object lock = new Object();
+      int depth = KrakenStreamingService.parseOrderBookSize(args);
+      return subscribe(channelName, MIN_DATA_ARRAY_SIZE, args).map(arrayNode -> {
```
earce, line 41: is there not a way we can just create a new orderbook from scratch when needed (bids/asks) rather then use synchronization? seems super heavy handed and more dangerous for devs who may not be as familiar with this

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
```diff
@@ -265,4 +266,34 @@ public void channelInactive(ChannelHandlerContext ctx) {
       }
     }
   }
+
+  static int parseOrderBookSize(Object[] args) {
+    if (args != null && args.length > 0) {
```
earce, line 271: can you move this logic outside to the methods that use it, we shouldn't have a method like this having to inspect an args[] object array.

also just maybe make it simpler since its one arg 
```
if (!(args != null && args.length > 0 && args[0] instanceof Number)) {
      LOG.error(
              "Order book size param type {} is invalid. Expected: {}. Default order book size has been used {}",
              obSizeParam.getClass().getName(),
              Number.class,
              ORDER_BOOK_SIZE_DEFAULT);
              return;
}
```

xchange-stream-kraken/src/test/java/info/bitrich/xchangestream/kraken/KrakenOrderbookExample.java
```diff
@@ -22,21 +46,17 @@ public static void main(String[] args) throws InterruptedException {
...
-                },
-                throwable -> {
-                  LOG.error("Fail to get ticker {}", throwable.getMessage(), throwable);
-                });
-
-    TimeUnit.SECONDS.sleep(60);
-
-    tickerDis.dispose();
+    subscribe(krakenExchange, BTC_USD);
+//    subscribe(krakenExchange, ETH_USD);
```
earce, line 50: please remove this

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
nielsdraaisma: We probably don't, i'll remove that, low impact as it's logging only anyway.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingService.java
nielsdraaisma: this is the original implementation that was used to determine the levels, I've opted to leave that as is as much as possible since change is already quite large.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
nielsdraaisma: done

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -19,58 +15,93 @@
...
+        }
+        OrderBook result = new OrderBook(lastTime.get(), Lists.newArrayList(asks), Lists.newArrayList(bids), true);
+        long localChecksum = createCrcChecksum(asks, bids);
+        if (expectedChecksum.get() > 0 && expectedChecksum.get() != localChecksum) {
+            LOG.warn("{} checksum does not match, expected {} but local checksum is {}", instrument, expectedChecksum.get(), localChecksum);
+            throw new IllegalStateException("Checksum did not match");
+        } else if (expectedChecksum.get() == 0) {
+            LOG.debug("Skipping {} checksum validation, no expected checksum in message", instrument);
+        }
+        return result;
```
wlk, line 92: you can just
```
return new OrderBook(lastTime.get(), Lists.newArrayList(asks), Lists.newArrayList(bids), true);
```
instead of creating it on line 84

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingChecksum.java
wlk: I suggest to make it more obvious and extract `10` into constant with descriptive name

gewure:
i'd take this as is.

xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
walec51: it would be better to fix the check sum code then making it dependant from this line of code which can easly get changed in the future

earce:
if you breakout the   `public String getUnsubscribeMessage(String channelName, Object... args) throws IOException {` changes into a separate review I can approve that quickly to make this review smaller/isolated

wlk:
Hi, what is the status of the PR?

If the only problem is with `public String getUnsubscribeMessage(String channelName, Object... args) throws IOException {` I'm happy to create new PR without this change.

According to [Kraken specification](https://docs.kraken.com/websockets/#message-unsubscribe), those parameters are optional when unsubscribing, therefore this change isn't really needed

If you connect to Kraken WS via any WS client you can replay the example payloads shown in the spec, you can see how you are successfully disconnected (only `channelID` is imporant). Here's how it looks:


```
{
  "connectionID": 17255188203992185000,
  "event": "systemStatus",
  "status": "online",
  "version": "1.8.6"
}

{
  "event": "subscribe",
  "pair": [
    "XBT/GBP"
  ],
  "subscription": {
    "name": "book",
    "depth": 25
  }
}

{
  "channelID": 352,
  "channelName": "book-25",
  "event": "subscriptionStatus",
  "pair": "XBT/GBP",
  "status": "subscribed",
  "subscription": {
    "depth": 25,
    "name": "book"
  }
}

...
...
updates are coming in
...

{
  "channelID": 352,
  "event": "unsubscribe"
}
{
  "channelID": 352,
  "event": "subscriptionStatus",
  "pair": "XBT/GBP",
  "status": "unsubscribed",
  "subscription": {
    "depth": 25,
    "name": "book"
  }
}
```

nielsdraaisma:
@wlk that change is required to ensure the correct book depth is still subscribed after reconnecting. If you connect to 100 initially but get subscribed to  the default depth 25 after reconnecting automatically that's not ideal and also breaks the checksumming logic at some point.

wlk:
@nielsdraaisma please take a look, the resubscription logic is here (taken from this PR):

KrakenStreamingMarketDataService lines 52-54
```
                                // Resubscribe to the channel, triggering a new snapshot
                                this.service.sendMessage(service.getUnsubscribeMessage(channelName, args));
                                this.service.sendMessage(service.getSubscribeMessage(channelName, args));
```

we still have `args` in order to subscribe in when the CRC fails, they are not needed in order to unsubscribe

nielsdraaisma:
It’s been a while, but pretty sure the depth is required when unsubscribing on a non default level.

wlk:
I have response from Kraken support (this mentions 'ticker', not 'book' but the point remains):
```
We recommend unsubscribing from WebSocket feeds using an unsubscribe request like the following:

{"event":"unsubscribe", "subscription":{"name":"ticker"}, "pair":["XBT/USD"]}

Note that the channel ID is not used, but the feed name (ticker) and currency pair (XBT/USD) are used instead.

For reference, the unsubscribe request is identical to the subscribe request, except that unsubscribe is used for the event value (instead of subscribe).
```

So I think @nielsdraaisma you are right that `args` are (or very soon will) be required, how do you want to progress regarding this PR? From my perspective (as an user) I'm looking forward to having this merged in as is.

nielsdraaisma:
Created https://github.com/knowm/XChange/pull/4145 to add the changes to the core streaming interface, hopefully we can get this PR sorted once that's done.

xchange-stream-bankera/src/main/java/info/bitrich/xchangestream/bankera/BankeraStreamingService.java
nielsdraaisma: Done.

nielsdraaisma:
I've addressed most comments, this should now be ready to go in @earce . Having done some more testing it looks like it's ok without the synchrized blocks in the end.

PR #4071, opened 2021-05-20
Add Vega.xyz protocol support

ConnorChristie:
Vega.xyz is a decentralized trading platform that is currently still in development and is only running a testnet.


pom.xml
```diff
@@ -446,6 +447,7 @@
                   <goal>jar</goal>
                 </goals>
                 <configuration>
+                  <javadocExecutable>${java.home}/bin/javadoc</javadocExecutable>
```
walec51, line 450: this should not be required

xchange-examples/src/main/java/org/knowm/xchange/examples/vega/VegaExampleUtils.java
```diff
@@ -0,0 +1,4 @@
+package org.knowm.xchange.examples.vega;
+
+public class VegaExampleUtils {
```
walec51, line 3: can be removed

xchange-vega/src/main/java/org/knowm/xchange/vega/service/VegaMarketDataService.java
```diff
@@ -0,0 +1,117 @@
...
+
+        return currencyPairs;
+    }
+
+    public List<CurrencyPair> getExchangeSymbols() {
+        return new ArrayList<>(getCurrencyPairs().keySet());
+    }
+
+    @Override
+    public Ticker getTicker(CurrencyPair currencyPair, Object... args) throws IOException {
```
walec51, line 47: please try to implement error mapping as described in: https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-7

earce:
Are you still pursuing this?

PR #4099, opened 2021-06-03
[CEX.IO streaming] reset the order book when an out of order message is received

scionaltera:
The documentation for CEX.IO at https://cex.io/websocket-api#orderbook-subscribe suggests that the order book representation will be "inconsistent" if we receive any messages out of order, or we miss even a single message. That suggests to me that we should be clearing the order book when that happens. This code originally threw an exception but then continued accepting messages and maintaining the old state without doing anything to correct the problem.

I have added code to clear the order book before throwing the exception so that the next message will be adopted into an empty `OrderBook` and we can build it back up over time.

I also cleared up a couple of warnings IDEA was showing for this file, but they should have no effect on the code.


xchange-stream-cexio/src/main/java/info/bitrich/xchangestream/cexio/CexioStreamingMarketDataService.java
```diff
@@ -29,17 +30,21 @@ public CexioStreamingMarketDataService(CexioStreamingRawService streamingOrderDa
...
     }
 
+    @SuppressWarnings("RedundantThrows")
     @Override
-    public OrderBook apply(CexioWebSocketOrderBookSubscribeResponse t) throws Exception {
+    public OrderBook apply(@NonNull CexioWebSocketOrderBookSubscribeResponse t) throws Exception {
       OrderBook retVal;
       if (prevID != null && prevID.add(BigInteger.ONE).compareTo(t.id) != 0) {
+        prevID = null;
+        orderBookSoFar = new OrderBook(new Date(), new ArrayList<>(), new ArrayList<>());
```
earce, line 46: is there some base layer of the orderbook that should be applied like a snapshot?

scionaltera: Good question. The documentation doesn't mention it but it seems like it might be a good idea. I can take a closer look and find out.

PR #4100, opened 2021-06-04
[CoinbasePro] add support for paged responses

matejsp:


earce:
this is breaking the build

matejsp:
Fixed, I also updated endpoints for prime:
> Our API endpoints will be moving to exchange.coinbase.com from prime.coinbase.com. prime endpoints will be deprecated on June 3rd, 2021:

PR #4107, opened 2021-06-07
[Bittrex] fix market order

matejsp:
fix bittrex market order

### DIFF
--- a/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeServiceRaw.java
+++ b/xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeServiceRaw.java
@@ -64,7 +64,7 @@ public class BittrexTradeServiceRaw extends BittrexBaseService {
   }
 
   public String placeBittrexMarketOrder(MarketOrder marketOrder) throws IOException {
-    return placeBittrexMarketOrder(marketOrder, TimeInForce.GOOD_TIL_CANCELLED);
+    return placeBittrexMarketOrder(marketOrder, TimeInForce.IMMEDIATE_OR_CANCEL);
   }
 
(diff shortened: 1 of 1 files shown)


xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeServiceRaw.java
```diff
@@ -64,7 +64,7 @@ public String placeBittrexLimitOrder(LimitOrder limitOrder, TimeInForce type) th
   }
 
   public String placeBittrexMarketOrder(MarketOrder marketOrder) throws IOException {
-    return placeBittrexMarketOrder(marketOrder, TimeInForce.GOOD_TIL_CANCELLED);
+    return placeBittrexMarketOrder(marketOrder, TimeInForce.IMMEDIATE_OR_CANCEL);
```
earce, line 67: is this correct/expected behavior on a market order?

matejsp:
@earce 

GOOD_TIL_CANCELLED is used for LIMIT order, but it does not make any sense for market order.

According to https://bittrex.github.io/api/v3#definition-NewOrder
- timeInForce is required (NewOrder data model)

According to https://bittrex.github.io/api/v3#topic-Placing-Orders
Order types and time in force. Market order can be used only with IMMEDIATE_OR_CANCEL or FILL_OR_KILL.

xchange-bittrex/src/main/java/org/knowm/xchange/bittrex/service/BittrexTradeServiceRaw.java
matejsp: The previous code returned 409 http error. GOOD_TIL_CANCELLED is used only for LIMIT order and not MARKET order. More in the comment.

PR #4146, opened 2021-06-30
[Coinbase] Update Coinbase v2 URL to correct one specified in docs

ataraymovich-bc:
https://developers.coinbase.com/api/v2


walec51 (changes requested):
did you test this change?

because looking at the code

```
@Path("/v2")
@Produces(MediaType.APPLICATION_JSON)
public interface CoinbaseAuthenticated extends Coinbase {
```

this will result in request URLs like /v2/v2/...

Issue #4153, opened 2021-07-05
[FTX] still  some checksum errors


mdvx:
the number is much reduced, but still getting them on my system, every few second (was 50x per seconds)

PR #4155, opened 2021-07-05
[FTX] getOrderbookChecksum FIX bad merge

mdvx:
that re-introduced the bug


mdvx:
https://github.com/knowm/XChange/issues/4153

PR #4162, opened 2021-07-09
[KRAKEN] Fix Errors when subscribing to trades and ticker, and orderbook depth…

mdvx:
… is supplied


wlk (approved):
I have ran into the same problem and this PR fixes the problem, thx @mdvx

PR #4174, opened 2021-07-13
[kucoin] add trading fee, fix remote init for anonymous setup (fee issue), cre…

matejsp:
Changes:
- fix remote init for anonymous setup (fee issue)
- create deposit address
- deposit address list (v2 & v1)
- add trading fee
- get orders api fixes

### DIFF
--- a/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAccountServiceRaw.java
+++ b/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAccountServiceRaw.java
@@ -10,6 +10,7 @@ import org.knowm.xchange.kucoin.dto.request.InnerTransferRequest;
 import org.knowm.xchange.kucoin.dto.response.AccountBalancesResponse;
 import org.knowm.xchange.kucoin.dto.response.AccountLedgersResponse;
 import org.knowm.xchange.kucoin.dto.response.ApplyWithdrawResponse;
+import org.knowm.xchange.kucoin.dto.response.DepositAddressResponse;
 import org.knowm.xchange.kucoin.dto.response.DepositResponse;
 import org.knowm.xchange.kucoin.dto.response.InternalTransferResponse;
 import org.knowm.xchange.kucoin.dto.response.Pagination;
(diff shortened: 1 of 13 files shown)


xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAccountServiceRaw.java
```diff
@@ -116,4 +117,49 @@ public InternalTransferResponse innerTransfer(InnerTransferRequest req) throws I
                 pageSize,
                 currentPage));
   }
+
+  public DepositAddressResponse createDepositAddress(
```
earce, line 121: can you keep parameters on the same line as method name for these three?

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinTradeService.java
```diff
@@ -141,7 +141,7 @@ public UserTrades getTradeHistory(TradeHistoryParams tradeHistoryParams) throws
     String nextPageCursor = null;
     if (startTime != null && startTime >= cutoffHistOrdersMillis) {
       Pagination<TradeResponse> fills =
-          getKucoinFills(symbol, page, TRADE_HISTORIES_TO_FETCH, startTime, endTime);
+          getKucoinFills(symbol, null, page, TRADE_HISTORIES_TO_FETCH, startTime, endTime);
```
earce, line 144: how does changing this to null affect behavior?

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinTradeServiceRaw.java
```diff
@@ -43,7 +59,7 @@ protected KucoinTradeServiceRaw(KucoinExchange exchange) {
                 nonceFactory,
                 passphrase,
                 symbol,
-                null,
+                orderId,
```
earce, line 62: what was this behavior before adding order id?

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAccountServiceRaw.java
matejsp: I too prefer them in the single line. But when contributing I try to replicate existing style in the same class. Other methods (like getDepositList, getWithdrawalsList, getAccountLedgers, etc).

earce: ah that's a shame then leave it as is

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinTradeServiceRaw.java
matejsp: OrderId is optional and supported by kucoin api. Since there is no common api to get fills for single order we are calling getKucoinFills directly. Well similar to DTO where all properties are listed but only some are mapped to xchange common api.

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinTradeService.java
matejsp: It is the same behaviour. It is not possible to get trades for a single order. If there is standard way I can implement it.
when you add something like: TradeHistoryParamOrderId I will implement it.

earce:
@matejsp can you run the build please, looks like this is affecting KucoinTradeRawDemo, and submit a PR to address?

matejsp:
@earce yes of course ... https://github.com/knowm/XChange/pull/4194
I was missing CI builds when doing PR. Any chance to get them back?

earce:
@matejsp not sure where they went @timmolter but you should always make sure the build passes locally.

timmolter:
I'm not sure either. I didn't delete them.

PR #4187, opened 2021-07-22
Kucoin Rate Limiting and other updates - WIP

nibarulabs:
Apologies up front to @matejsp - I probably should have forked off his PR for the changes he made instead of pulling them into my PR. I needed those changes for my project yesterday, so that's why I pulled them in.

This is a decent update of the Kucoin library. It adds the rate limiting and retry to the various rest methods. In addition, I could not find the method to actually fetch candle history (?) so that has been added in along with the relevant Kline parts. I am doing this in 3 phases: 

- [x]  market service and tests 
- [ ]  account service
- [ ]  trade service 

I cleaned up other small things as well and have been trying to make sure that we have as close to complete api support as possible.

Still WIP at this point, but creating the PR so @earce can have a look.


xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
```diff
@@ -17,12 +24,14 @@
    */
   public static final String PARAM_SANDBOX = "Use_Sandbox";
 
-  static final String SANDBOX_HOST = "openapi-sandbox.kucoin.com";
+  static final String SANDBOX_HOST = APIConstants.API_SANDBOX_HOST;
```
walec51, line 27: don't think having two constants with the same value is necessary - would be best to get rid of one

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinKline.java
```diff
@@ -0,0 +1,77 @@
+package org.knowm.xchange.kucoin.dto.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Data;
+import org.knowm.xchange.currency.CurrencyPair;
+import org.knowm.xchange.kucoin.dto.KlineIntervalType;
+
+import java.math.BigDecimal;
+
+@Data
```
walec51, line 10: this adds getters and setters which you then also added manually below

DTOs should be value objects so either change this annotation to `@Value` and remove those implicit getters bellow

or dont use lombok and make all fields final with just getters

walec51:
so far looks good, jus two small remarks

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
nibarulabs: So would you would prefer to just harcode the strings in the KucoinExchange class and get rid of the entries in the APIConstants? This seems to be the approach in the other exchange modules, so I think you're right. I'll make the change.

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinKline.java
nibarulabs: Ok, makes sense.

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
```diff
@@ -22,19 +26,22 @@
    */
   public static final String PARAM_SANDBOX = "Use_Sandbox";
 
-  static final String SANDBOX_HOST = "openapi-sandbox.kucoin.com";
-  static final String SANDBOX_URI = "https://" + SANDBOX_HOST;
+  static final String SANDBOX_URI = "https://openapi-sandbox.kucoin.com";
+  static final String LIVE_URI = "https://api.kucoin.com";
```
earce, line 30: should probably be PROD_URI no?

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
```diff
@@ -50,22 +57,34 @@ public void applySpecification(ExchangeSpecification exchangeSpecification) {
...
 
   @Override
   public ExchangeSpecification getDefaultExchangeSpecification() {
     ExchangeSpecification exchangeSpecification = new ExchangeSpecification(this.getClass());
     exchangeSpecification.setSslUri(LIVE_URI);
-    exchangeSpecification.setHost(LIVE_HOST);
+    try {
+      URL url = new URL(KucoinExchange.LIVE_URI);
+      exchangeSpecification.setHost(url.getHost());
+    } catch (MalformedURLException ignored) {
```
earce, line 72: log something here

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
```diff
@@ -22,19 +26,22 @@
...
     if (exchangeSpecification.getExchangeSpecificParameters() != null) {
       if (Boolean.TRUE.equals(
           exchangeSpecification.getExchangeSpecificParametersItem(PARAM_SANDBOX))) {
         logger.debug("Connecting to sandbox");
         exchangeSpecification.setSslUri(KucoinExchange.SANDBOX_URI);
-        exchangeSpecification.setHost(KucoinExchange.SANDBOX_HOST);
+        try {
+          URL url = new URL(KucoinExchange.SANDBOX_URI);
+          exchangeSpecification.setHost(url.getHost());
+        } catch (MalformedURLException ignored) {
```
earce, line 43: log something here

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
nibarulabs: Sure, I can make that change.

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinExchange.java
nibarulabs: Will do on both of these.

Issue #4195, opened 2021-08-02
[FTX] No implementation of MarketDataService.getTicker


mathvdh:
Hi All,

There is no implementation of `getTicker` for the Market Data Service.

According to the FTX docs, they don't really give tickers as defined in Xchange.
But by combining the information of `getMarket` and `getCandles` (which I added) we can fill a full ticker.
I will submit this implementation shortly.

Let me know if it's OK for you or if you have any comments or maybe a simple way of getting this info.

Thanks

cymp:
Good practices in this project is to use only one API call per operations. If not possible, please take care of rate limits by using [resilience](https://github.com/knowm/XChange/wiki/Implementing-resiliency-features).

mathvdh:
@cymp Ok thanks, I'll have a look ... and come back to discuss it

earce:
I saw this was merged I put a comment on there closing this as complete

PR #4196, opened 2021-08-02
Implementation of getTicker, fix for 4195

mathvdh:
Basically I implemented :

- Getting a single market info in getMarket (instead of getMarkets)
- Getting historical candle data (getCandles)
- Using both of the above I implemented the getTicker method.

### DIFF
--- a/xchange-ftx/src/main/java/org/knowm/xchange/ftx/FtxAdapters.java
+++ b/xchange-ftx/src/main/java/org/knowm/xchange/ftx/FtxAdapters.java
@@ -22,6 +22,7 @@ import org.knowm.xchange.dto.account.OpenPosition;
 import org.knowm.xchange.dto.account.OpenPositions;
 import org.knowm.xchange.dto.account.Wallet;
 import org.knowm.xchange.dto.marketdata.OrderBook;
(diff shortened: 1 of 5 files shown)


mathvdh:
As stated in the title this is a fix for #4195

xchange-ftx/src/main/java/org/knowm/xchange/ftx/FtxAdapters.java
```diff
@@ -350,4 +349,37 @@ public static OpenPositions adaptOpenPositions(List<FtxPositionDto> ftxPositionD
   public static BigDecimal lendingRounding(BigDecimal value) {
     return value.setScale(4, RoundingMode.DOWN);
   }
+
+  public static Ticker adaptTicker(FtxResponse<FtxMarketDto> ftxMarketResp,
+                                   FtxResponse<List<FtxCandleDto>> ftxCandlesResp,
+                                   CurrencyPair currencyPair) {
+
+    System.out.println("marketResp"  + ftxMarketResp.toString());
```
earce, line 357: just saw this, these should be done using the logger framework not system.println

PR #4202, opened 2021-08-05
Kraken open orders by pair

kevinstl:
The Kraken API does not allow for filtering by currency pair. This change filters by desired currency pair after retrieving all open orders from Kraken.


xchange-kraken/pom.xml
```diff
@@ -29,6 +29,10 @@
             <artifactId>xchange-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock-jre8</artifactId>
```
walec51, line 34: make this dependancy last and add `<scope>test</scope>`

xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenAdapters.java
```diff
@@ -259,13 +259,21 @@ public static CurrencyPair adaptCurrencyPair(String krakenCurrencyPair) {
...
     for (Entry<String, KrakenOrder> krakenOrderEntry : krakenOrders.entrySet()) {
       KrakenOrder krakenOrder = krakenOrderEntry.getValue();
       KrakenOrderDescription orderDescription = krakenOrder.getOrderDescription();
 
-      if (!"limit".equals(orderDescription.getOrderType().toString())) {
+      if (!"limit".equals(orderDescription.getOrderType().toString()) ||
+              ( currencyPair != null &&
+                !currencyPair.equals(adaptCurrencyPair(orderDescription.getAssetPair()))
+              )
+      ) {
```
walec51, line 276: please extract this check to some method with a meaningful name like `if (isNotPlainLimitOrderOr...()) {`

xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenTradeService.java
```diff
@@ -27,11 +29,15 @@ public KrakenTradeService(Exchange exchange) {
...
   public OpenOrders getOpenOrders() throws IOException {
-    return getOpenOrders(createOpenOrdersParams());
+    return getOpenOrders(null);
   }
 
   @Override
   public OpenOrders getOpenOrders(OpenOrdersParams params) throws IOException {
+    if (params != null && params instanceof OpenOrdersParamCurrencyPair) {
+      OpenOrdersParamCurrencyPair openOrdersParamCurrencyPair = (OpenOrdersParamCurrencyPair) params;
+      return KrakenAdapters.adaptOpenOrders(super.getKrakenOpenOrders(), openOrdersParamCurrencyPair.getCurrencyPair());
```
walec51, line 39: adapters should adapt DTO A to DTO B

they should not do client side filtering

the condition you added in the adapter because of this is hard to read and understand

if the exchange does not allow for server side filtering then please to the filtering here, before passing the orders to the adapter

xchange-kraken/src/test/java/org/knowm/xchange/kraken/service/KrakenTradeServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+                                        .withHeader("Content-Type", "application/json")
+                                        .withBodyFile(ORDERS_FILE_NAME)
+                        )
+        );
+
+        OpenOrders openOrders = classUnderTest.getOpenOrders();
+
+        assertThat(openOrders).isNotNull();
+        assertThat(openOrders.getOpenOrders()).hasSize(jsonRoot.get("result").
+                get("open").size());
```
walec51, line 48: both openOrders and jsonRoot are from jacksons parsing mechanism

i think it would be better to just hard code the expected values here

walec51:
look good, some small remarks

xchange-kraken/pom.xml
kevinstl: Okay, thanks.

xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenTradeService.java
kevinstl: Got it. Will do.

xchange-kraken/src/test/java/org/knowm/xchange/kraken/service/KrakenTradeServiceTest.java
kevinstl: Done.

PR #4203, opened 2021-08-06
[Kraken] Implement get trade history by currency pair

kevinstl:
The Kraken API does not allow for filtering by currency pair. This change filters by desired currency pair after retrieving all trade history from Kraken.


xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenAdapters.java
```diff
@@ -187,7 +187,7 @@ public static Ticker adaptTicker(KrakenTicker krakenTicker, CurrencyPair currenc
 
   public static List<Ticker> adaptTickers(Map<String, KrakenTicker> krackenTickers) {
     List<Ticker> tickers = new ArrayList<>();
-    for (Map.Entry<String, KrakenTicker> ticker : krackenTickers.entrySet()) {
+    for (Entry<String, KrakenTicker> ticker : krackenTickers.entrySet()) {
```
earce, line 190: fix spelling

PR #4213, opened 2021-08-10
[Coinbasepro-Stream] Add support for overriding websocket uri

theluckysuccess:
# Problem
Currently we are unable to override our websocket uris in our streaming projects. They are essentially hardcoded. There are situations that exist when one would want to change them like demos or due to internal servers structures.

# Solution
This update would allow for any `xchange-stream` project to expose an override for their websocket uri via their respective `StreamingExchange`. In this pr, I have done this in `xchange-stream-coinbasepro` in the `CoinbaseProStreamingExchange`.

# Considerations
I could have implemented this via `ExchangeSpecification.getExchangeSpecificParametersItem` however I thought this functionality was something all streaming exchanges could benefit from. If there are strong objects, I can rework this pr to override the websocket uri only for `xchange-stream-coinbasepro` via `ExchangeSpecification.getExchangeSpecificParametersItem`

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/ExchangeSpecification.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/ExchangeSpecification.java
@@ -24,6 +24,7 @@ public class ExchangeSpecification {
   private String apiKey;
   private String sslUri;
   private String plainTextUri;
+  private String overrideWebsocketApiUri;
   private String host;
   private int port = 80;
   private String proxyHost;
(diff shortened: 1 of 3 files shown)


earce:
Not sure how I feel about this BUT in any case you should not be renaming an existing property to do this, if anything add a new one.

@timmolter for second pair of eyes.

theluckysuccess:
Hey just checking in. Any chance that this will be approved soon? I'm open to any changes or suggestions.

theluckysuccess:
🥳

PR #4173, opened 2021-07-13
Options contract instrument

vasylp3:
#4154

PR #4214, opened 2021-08-13
[FTX] Ftx modify order fix a typo and change a call in order to work

makarid:


xchange-ftx/src/main/java/org/knowm/xchange/ftx/FtxAuthenticated.java
```diff
@@ -194,13 +194,13 @@ FtxResponse changeSubAccountName(
...
   @DELETE
   @Path("/orders/by_client_id/{client_order_id}")
-  FtxResponse<FtxOrderDto> cancelOrderByClientId(
-      @HeaderParam("FTX-KEY") String apiKey,
-      @HeaderParam("FTX-TS") Long nonce,
-      @HeaderParam("FTX-SIGN") ParamsDigest signature,
-      @HeaderParam("FTX-SUBACCOUNT") String subaccount,
-      @PathParam("client_order_id") String clientOrderId)
-      throws IOException, FtxException;
+  FtxResponse<String> cancelOrderByClientId(
```
earce, line 197: why are you moving to a string?

makarid: Because the response is a string and i got error when i had something else.

earce: What was the error message

makarid:
Could't parse the return using FtxOrderDto. It was an error about json parse

Στις Δευ, 16 Αυγ 2021 στις 5:55 μ.μ., ο/η Erick Arce <
***@***.***> έγραψε:

earce:
What was the error, can you just adjust the DTO to handle it? We never set to string.

makarid:
The cancelOrder return on the FTX is just a string. Why we cannot put a
String to a return value that is string? Here is the response from FTX docs

{
  "success": true,
  "result": "Order queued for cancelation"}


Στις Τρί, 17 Αυγ 2021 στις 9:37 μ.μ., ο/η Erick Arce <
***@***.***> έγραψε:

earce:
Can you not turn this into a DTO?

makarid:
I am sorry but i cannot understand the problem. Why is it so crucial to
have FtxOrderDto for cancelOrders even if the response is just a string?
Thanks

Στις Τετ, 18 Αυγ 2021 στις 5:16 μ.μ., ο/η Erick Arce <
***@***.***> έγραψε:

earce:
Not crucial but throughout the codebase we serialize to DTOs and this is about keeping the codebase consistent where possible. Anyone using this now this will break them.

makarid:
@earce you can check if you like if the cancelOrderByClientId works as it is right now, because when i check it, it didn't work, that is why i change it. I think it is ok to change something that it doesn't work, i didn't change it because i didn't like it or something.

earce:
> @earce you can check if you like if the cancelOrderByClientId works as it is right now, because when i check it, it didn't work, that is why i change it. I think it is ok to change something that it doesn't work, i didn't change it because i didn't like it or something. If someone has a problem with this fix, please i am more than happy to discuss about it and find a solution

We should change the return value to a DTO and get it working if something about it is failing as I've described before.

makarid:
> > @earce you can check if you like if the cancelOrderByClientId works as it is right now, because when i check it, it didn't work, that is why i change it. I think it is ok to change something that it doesn't work, i didn't change it because i didn't like it or something. If someone has a problem with this fix, please i am more than happy to discuss about it and find a solution
> 
> We should change the return value to a DTO and get it working if something about it is failing as I've described before.

Latest merge with Deribit has made some breaks on DeribitTests and because of that it fails to pass the maven tests. I think that his PR is the problem but maybe i am wrong. #4173

walec51 (approved):
I'll merge this because it makes the project build again

however the code responsible for parsing future and options codes is a total mess and I doubt it actually works well

PR #4215, opened 2021-08-14
[FTX] Adding ordersChanges stream in Ftx and minor code cleanups

makarid:


xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingService.java
```diff
@@ -79,6 +79,8 @@ protected String getChannelNameFromMessage(JsonNode message) {
 
     if ("fills".equals(message.get("channel").asText())) {
       channelName = message.get("channel").asText();
+    } else if ("orders".equals(message.get("channel").asText())) {
```
earce, line 82: if this is the same have the if contain an || since the action is the same

cymp: btw, why not adding sonar or something like that to the project?

makarid: lol i just saw it @earce

makarid: what is sonar @cymp ? thanks

PR #4220, opened 2021-08-17
[FTX] Orderbook checksum XRP/BTC

mdvx:
we are currently unable to calculate the checksum for XRP/BTC orderbook, it has been raised with FTX - need a workaround


mdvx:
better solution

Issue #4216, opened 2021-08-16
Android app?


hovi:
I am just wondering, why is there no android app using xchange on google play? Are there any legal or technical reasons why I cannot do that?
I'd like to use an app with this simple interface even if I have to make one, but in that case I might as well share it with the world :-)

cymp:
There is no reason you couldn't use XChange for an Android app. In fact, there is even a discussion somewhere here about migrating to java version>8, but people raised concerns about Android compatibility. If you want to develop an Android app and bring your experience here, it would be more than welcome.

hovi:
I see, I just read #2945 
Thank you, I will bring any useful feedback.

bschnider:
I do not know if it is the right place to post my question regarding an error I have been facing when trying to simply extract the ticker info from Kraken in my Android application. I am quite new in the Android world and might also have incorrectly imported the library but still thought it could be worth posting my issue here, as I have read that some incompatibilities have been faced with Java 11 (my application is based on a JDK 8).

**My import:**
_build.gradle (:app)_
`android {
    compileSdkVersion 30
    buildToolsVersion "30.0.3"

    defaultConfig {
        applicationId "com.example.trackmycrypto"
        minSdkVersion 19
        multiDexEnabled true
        targetSdkVersion 30
        versionCode 1
        versionName "1.0"

        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
    }

    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
}

dependencies {
    implementation 'androidx.multidex:multidex:2.0.1'
    implementation 'org.knowm.xchange:xchange-core:5.0.13'
    implementation 'org.knowm.xchange:xchange-kraken:5.0.13'
}`

build.gradle (APP_NAME)
`allprojects {
    repositories {
        google()
        jcenter()
        mavenCentral()
        maven { url 'https://jitpack.io' }
    }
}`

I have read your explanations regarding the dependencies to add in the `pom.xml` file, but I do not have such file in my application's project and usually, the third party libraries are directly imported when synchronizing the dependencies of the gradle files (which I did). Could it be that the import of the library was incomplete or should be performed manually ?

**My simple test code:**
`ExchangeSpecification exSpec = new KrakenExchange().getDefaultExchangeSpecification();
[...]

PR #4223, opened 2021-08-18
Android fixes

hovi:
Tried to run simple services on coinmate using android API 23. Run into a few problems, that I had to solve.

Coinmate fix is imho good to merge even without it being necessary to run (no reason for REST interface to contain HashMaps).

Core change (ThreadLocal.withInitial) is up for debate. Maybe this simpler version could run into a race condition, haven't really checked.

#4216


hovi:
There are actually a few more classes extending HashMap or ArrayList in coinmate codebase that need removing.

hovi:
Learning more about codebase, I will submit better PRs later..

PR #4229, opened 2021-08-19
Binance.us support

nibarulabs:
Adds support for Binance.US rest and streaming api. This does *not* guarantee 1:1 rest api support as .us and .com have diverged quite a bit. However, fetching candles, getting account info, and a few other things should work as expected.

For the streaming api, Trades, OrderBook, Tickers should all work (those I have in production already).

This addition is meant to get the ball rolling and will need to be further updated to support more rest features and probably streaming endpoints. The Binance.com api is quite different now so it's possible in the future that the .US code will need to be broken off into its own package(s).


nibarulabs:
Test updates are incoming..

nibarulabs:
I have this in production and it's working without any issues.

garciapd:
Nice, could that been merged?

nibarulabs:
Yah, just needs review and I can make any changes needed. Binance.us and .com api's have diverged quite a bit, so this pr is just meant to provide a baseline for .us support.

PR #4232, opened 2021-08-20
[Coinmate] Interfaces dtos

hovi:
- Extending collection interfaces for data objects rather than collection class implementations.
- Tests are only failing, because json parsing library used for tests cannot handle List/Map interfaces to create instance. This can be fixed by specifing, what implementation to use. However since tests do pretty much nothing useful, basically testing used json parser, which is even different from parser used when running the services, I don't think they are worth "fixing".


hovi:
This one actually breaks things, gonna have to fix this.

Issue #4226, opened 2021-08-19
BaseParamsDigest, ThreadLocal and multithreading


hovi:
[BaseParamsDigest](https://github.com/knowm/XChange/blob/develop/xchange-core/src/main/java/org/knowm/xchange/service/BaseParamsDigest.java) class creates this Mac instance in it's constructor and saves it to ThreadLocal variable.
This class is being used when communicating with APIs.

When you create Exchange and communicate, this class gets created and cached, but when you access it from different thread, you get NPE, because there is nothing in the ThreadLocal.

I came across this while running exchange services in android using kotlin coroutines, where I just say "run this on IO thread" and sometimes it's not the same IO thread, causing the NPE.

I have no idea, why is the ThreadLocal there in the first place at all. It's been there since first commit of the file in 2014.

From usage analysis, variable is either for internal calculations (although getMac() is public, not protected) or sometimes accessed in services, where the class is created only to get this Mac and then thrown away. So I don't think threadlocal has any use.

I suggest to save this Mac variable to regular local property of the class, remove ThreadLocal completely.

PR #4233, opened 2021-08-20
Made getMac return one-time calculated final property rather than Thr…

hovi:
…eadLocal variable.


hovi:
Problem described in #4226

PR #4234, opened 2021-08-20
[Coinmate] Added OrderStatus for streaming open orders.

hovi:


xchange-stream-coinmate/src/main/java/info/bitrich/xchangestream/coinmate/v2/CoinmateStreamingAdapter.java
```diff
@@ -71,12 +71,27 @@ public static OpenOrders adaptWebsocketOpenOrders(
...
                       Date.from(Instant.ofEpochMilli(coinmateWebsocketOpenOrder.getTimestamp())))
                   .limitPrice(BigDecimal.valueOf(coinmateWebsocketOpenOrder.getPrice()))
+                      .orderStatus(statusFromString(coinmateWebsocketOpenOrder.getOrderChangePushEvent()))
                   .build());
         });
     return new OpenOrders(openOrders);
   }
 
-  public static Trade adaptTrade(CoinmateWebSocketTrade webSocketTrade, CurrencyPair currencyPair) {
+    private static Order.OrderStatus statusFromString(String orderChangePushEvent) {
```
earce, line 80: can you add a null check here and if so return unknown status

hovi: Yes I can rename, but what do you mean by null check? There's fallback return in the method at the end, that returns UNKNOWN for everything else including null, doesn't it? Or what null check?

Do you prefer amended commit or should I split into more commits so that you squash it when it's done?

hovi: Ah, java switch throws NPE when argument is null.

PR #4239, opened 2021-08-22
[KRAKEN] Crossed books that pass checksum

mdvx:
Getting crossed books even after the checksum validation is passed, since we have re-subscription logic, triggered by IllegalStateException, we can use it.


mdvx:
NB: we are still getting checksum failure intermittently on this exchange, if i drop and restart the whole socket, i can usually get it to run for a few minutes.  The problem is not replicable by the example test case.

16:01:04.313[WARN ][nioEv..-4-1][i.b.x.k.KrakenStreamingMarketDataService]Resubscribing LTC/USD channel after adapter error Checksum did not match
16:01:05.092[WARN ][nioEv..-4-1][i.b.x.kraken.KrakenStreamingAdapters]ETH/USD checksum does not match, expected 794903018 but local checksum is 3288686771

mdvx:
Bitstamp, Kraken, FTX (very occasionally) are all producing these crossed books

PR #4243, opened 2021-08-24
[FTX] Resubscribe streaming orderbook if error occurs.

makarid:


xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingMarketDataService.java
```diff
@@ -23,19 +23,37 @@ public FtxStreamingMarketDataService(FtxStreamingService service) {
 
   @Override
   public Observable<OrderBook> getOrderBook(CurrencyPair currencyPair, Object... args) {
-    OrderBook orderBook = new OrderBook(null, new ArrayList<>(), new ArrayList<>());
+    OrderBook orderBook = new OrderBook(null, Lists.newArrayList(), Lists.newArrayList());
+    String channelName = "orderbook:" + FtxAdapters.adaptCurrencyPairToFtxMarket(currencyPair);
 
     return service
```
earce, line 29: is it possible here rather then to catch and clear the orderbook, unsubscirbe and subscribe, to throw and trigger the subscribe to create a new orderbook etc, we do this else where in the code

bitcoinmat-dev: Hello, i had issues with orderbook resubscription and never subscribing back if an error occurred on the stream(a checksum error or something else). I saw this implementation on Kraken streaming orderbook (in which i had the same issue if the checksum was not correct). So, i liked the implementation of Kraken stream and i implement the same one. Now Ftx orderbook streaming works like a charm without any issues. Is it possible to just do the same thing here on FTX?

makarid: Hello, i had issues with orderbook resubscription and never subscribing back if an error occurred on the stream(a checksum error or something else). I saw this implementation on Kraken streaming orderbook (in which i had the same issue if the checksum was not correct). So, i liked the implementation of Kraken stream and i implement the same one. Now Ftx orderbook streaming works like a charm without any issues. Is it possible to just do the same thing here on FTX?

earce: Sure, it's not the best way but we can leave it.

PR #4244, opened 2021-08-25
Cumulative Fee Calculation for Gemini Orders

gmax0:
Removed some boilerplate POJO + added UserReference to the limit order object returned from the getOrderDetails call.


xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/GeminiAdapters.java
```diff
@@ -122,6 +122,19 @@ public static Order adaptOrder(GeminiOrderStatusResponse geminiOrderStatusRespon
     OrderStatus orderStatus = adaptOrderstatus(geminiOrderStatusResponse);
     Date timestamp = new Date(geminiOrderStatusResponse.getTimestampms());
 
+    // Calculate Fees in counter currency
+    BigDecimal fee = null;
```
earce, line 126: set this to BigDecimal.ZERO and then in your loop you can just add for all and it becomes a 1-liner

xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/service/GeminiTradeServiceRaw.java
```diff
@@ -134,6 +132,47 @@ public GeminiOrderStatusResponse getGeminiOrderStatus(String orderId) throws IOE
...
+      if (orderId == null) {
+        clientOrderId = ((GeminiTradeService.GeminiOrderQueryParams) params).getClientOrderId();
+      }
+
+      includeTrades = ((GeminiTradeService.GeminiOrderQueryParams) params).isIncludeTrades();
+      account = ((GeminiTradeService.GeminiOrderQueryParams) params).getAccount();
+    }
+
+    try {
+      GeminiOrderStatusResponse orderStatus =
```
earce, line 154: just return inline, no need to set this var

xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/GeminiAdapters.java
gmax0: I ran into issues with the scale of the BigDecimal when setting to BigDecimal.ZERO originally.
Example, adding 0 to 0.00023 would still result in a BigDecimal displaying as 0. Tried to address this by setting the scale before and after the calculations, but it didn't seem to work.

gmax0: I was debugging the BigDecimal returned using IntelliJ's debugger.

xchange-gemini/src/main/java/org/knowm/xchange/gemini/v1/service/GeminiTradeServiceRaw.java
gmax0: Done

PR #4245, opened 2021-08-25
Update gateio payload shape for getOrderStatus

tonykwok1992:


xchange-gateio/src/main/java/org/knowm/xchange/gateio/GateioAuthenticated.java
```diff
@@ -113,7 +113,8 @@ GateioDepositsWithdrawals getDepositsWithdrawals(
   @POST
   @Path("private/getorder")
   GateioOrderStatus getOrderStatus(
-      @FormParam("order_id") String orderId,
+      @FormParam("orderNumber") String orderNumber,
+      @FormParam("currencyPair") String currencyPair,
```
earce, line 117: this seems suspicious, gateio just changed params for this? this would break everyone

Issue #4172, opened 2021-07-13
[Kraken] KrakenStreamingAdapters.adaptOrderbookMessage always sets lastTime to 1/1/1970


wlk:
There is a bug in `KrakenStreamingAdapters.adaptOrderbookMessage` the `lastTime` value is never updated and this results in always returning same value which is epoch time = 0.

line 60:
```
        final AtomicReference<Date> lastTime = new AtomicReference<>(Date.from(Instant.EPOCH));
```

line 91:
```
        return new OrderBook(lastTime.get(), Lists.newArrayList(asks), Lists.newArrayList(bids), true);
```

There are 2 options I see as a potential fix:
- Set time equal to latest timestamp observed in the individual orders
- Or, Set time to "current time"

ping @nielsdraaisma 

This affects current `develop` branch

Gepro83:
@wlk are you able to place orders with the current version of the kraken libarry? I get "EGeneral:Invalid arguments" exception when using like this:

    ```
kraken.tradeService.placeMarketOrder(MarketOrder.Builder(org.knowm.xchange.dto.Order.OrderType.BID, CurrencyPair.ETH_EUR)
        .originalAmount(BigDecimal("0.001"))
        .build())
```
API key etc works, i can querry the wallet balance...

earce:
@Gepro83 you probably are missing something in your request.

@wlk I think the last timestamp across all orders is more accurate otherwise you are just timestamping when XChange received the order

wlk:
The "per order" timestamp is set correctly, but "per orderbook" timestamp is set to null, it's not provided directly by Kraken, possible options:
- set to  `null` 
- set timestamp = current time
- set timestamp = MAX of all individual orders in the book

the current behavior of setting timestamp to `0` is a bug to me, wdyt?

earce:
Yea @wlk I understand, and what I was saying is that I think #3 makes the most sense otherwise as mentioned you are just timestamping the message when your systems receives it.

timmolter:
Traditionally, for this project, we would set the timestamp (Date) to null indicating that the server doesn't provide any timestamp. In this case however I think option #3 makes the most sense since the timestamp can be inferred by the other data.

mdvx:
Up vote! i see the Timestamp passed, and then get lost

ldrygala:
I would like to fix this issue. PR should be ready this week.

PR #4248, opened 2021-08-25
fix last time in order book

ldrygala:
fixes: https://github.com/knowm/XChange/issues/4172

### DIFF
--- a/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
+++ b/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
@@ -91,7 +91,10 @@ public class KrakenStreamingAdapters {
         } else if (bids.size() > 0 && asks.size() > 0 && bids.first().getLimitPrice().compareTo(asks.first().getLimitPrice()) >= 0) {
             throw new IllegalStateException("CROSSED book " + instrument + " " + bids.first().getLimitPrice() + " >= " + asks.first().getLimitPrice());
         }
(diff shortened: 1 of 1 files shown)


xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingAdapters.java
```diff
@@ -90,7 +91,8 @@ public static OrderBook adaptOrderbookMessage(int depth, TreeSet<LimitOrder> bid
         } else if (bids.size() > 0 && asks.size() > 0 && bids.first().getLimitPrice().compareTo(asks.first().getLimitPrice()) >= 0) {
             throw new IllegalStateException("CROSSED book " + instrument + " " + bids.first().getLimitPrice() + " >= " + asks.first().getLimitPrice());
         }
-        return new OrderBook(lastTime.get(), Lists.newArrayList(asks), Lists.newArrayList(bids), true);
+        final Date lastTime = Stream.concat(asks.stream(), bids.stream()).map(LimitOrder::getTimestamp).max(Date::compareTo).orElse(Date.from(Instant.EPOCH));
```
earce, line 94: the date should be null if we cannot figure out the data from the snapshot or bids/asks because otherwise you are just timestamping the time we received the message

ldrygala: fixed

ldrygala: fixed

PR #4253, opened 2021-08-27
[Binance] trade stream support maker and taker order id

moritzhabegger:


xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -170,6 +170,8 @@ public BinanceStreamingMarketDataService(
                     .originalAmount(rawTrade.getQuantity())
                     .instrument(currencyPair)
                     .price(rawTrade.getPrice())
+                    .makerOrderId(String.valueOf(rawTrade.getSellerOrderId()))
```
earce, line 173: maker or taker depends on if the limit order was a buy or sell, seller is not always a maker and buyer is not always a taker

moritzhabegger: Thanks for the review, I corrected the code in respect of trade.isBuyerMarketMaker()

Issue #3647, opened 2020-08-11
Support for tinyproxy at the connect method of NettyStreamingExchange


scrisini:
### **Overview**
Due the firm I work form's use of reverse proxy extensively based on tinyproxy.
deploying and connecting to a number of providers such as kraken,binance, coinbasepro we had to basically modify several classes to support wss connection stage 
changes were made 
NettyStreamingService
 - for handling proxy settings and connect initiattion 
BitstampStreamingExchange 
to ensure the default implementation of applyStreamingSpecification in StreamingExchange is invoked

below is a snippet of StreamingExchange, as modified
    default void applyStreamingSpecification(ExchangeSpecification exchangeSpec, NettyStreamingService<?> streamingService){
        streamingService.setSocksProxyHost((String) exchangeSpec.getExchangeSpecificParametersItem(SOCKS_PROXY_HOST));
        streamingService.setSocksProxyPort((Integer) exchangeSpec.getExchangeSpecificParametersItem(SOCKS_PROXY_PORT));
        _**streamingService.setHttpProxyHost(exchangeSpec.getProxyHost());
        streamingService.setHttpProxyPort(exchangeSpec.getProxyPort());**_
        streamingService.setBeforeConnectionHandler((Runnable) exchangeSpec.getExchangeSpecificParametersItem(ConnectableService.BEFORE_CONNECTION_HANDLER));

        Boolean accept_all_ceriticates = (Boolean) exchangeSpec.getExchangeSpecificParametersItem(ACCEPT_ALL_CERITICATES);

badgerwithagun:
Hi @scrisini - this sounds good. Would you be willing to raise a PR with the changes? I'm sure others would find them useful.

scrisini:
Hi @badgerwithagun great! will do. 
im a newbie on this so I just clone the latest and follow the guidelines and then ready I submit for peer review and PR?

scrisini:
Hi has taken me a while to catchup.
but im ready to create a PR but not sure how this is done here
1. I cloned the repo today
2. branched off from develop - named it tinyproxy_3647
3. however im now ready to push to GitHub and do a PR
4. but looks like my username isn't permitted to do that?
Cheers

scrisini:
> Yes, @scrisini, don't worry about making mistakes. You can't break anything ;).
> 
> Someone will let you know if your PR needs work.

Hello
just following this up as I now have incorporated the fix to the latest cloned copy and would like to know whats next step to create a PR?
thank you.

[...]

PR #4263, opened 2021-09-08
closes knowm/XChange#3647  Added support for tinyproxy acting as reverse proxy.

scrisini:
This will support a reverse proxy setup such as tinyproxy 
when connecting to any of the supported exchanges
used 
by a spring java app like this
image
use case 
is going thr a lightweight firewall like tinyproxy

### DIFF
--- a/xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
+++ b/xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
@@ -75,7 +75,7 @@ public class BinanceStreamingExchange extends BinanceExchange implements Streami
 
     ProductSubscription subscriptions = args[0];
     streamingService = createStreamingService(subscriptions);
-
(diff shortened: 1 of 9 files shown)


timmolter:
dumb question: couldn't you just run nginx or something to do that?

timmolter:
Or is it that the code in XChange for streaming somehow needs to be modified? Why exactly?

scrisini:
HI Tim
yes specifically for streaming and the additional functionality is to bridge the gap in setting up web socket session when a proxy is used in between 
Changes are
1. invoke applyStreamingSpefication on connect in class XXXSreamingExchange
2. set the proxy host and port in method applyStreamingSpec - StreamingExchange
3. Add a proxy friendly handler in nettStreamingService if proxys are given.
4. and importantly WebSocketClientHandler to manage the handshake when a proxy is used
I hope this is considered for inclusion as there shouldn't be an impact on the main flow but will be useful for proxy based implementation 
and tinyproxy is used as a lightweight firewall to block or white listed web socket url.
particularly important in our industry.

Issue #4190, opened 2021-07-23
Binance withdrawal exception on succesful service call


aligokalppeker:
Hi,

I am testing Binance withdrawal API. It successfully withdraws XRP from the wallet and triggers the transfer. But service response code can not be parsed correctly(maybe it is changed/updated). So it throws an exception on parsing the response of the service :

Exception in thread "main" org.knowm.xchange.exceptions.ExchangeException: null (HTTP status code: 0)
	at org.knowm.xchange.binance.BinanceErrorAdapter.adapt(BinanceErrorAdapter.java:49)
	at org.knowm.xchange.binance.service.BinanceAccountService.withdrawFunds(BinanceAccountService.java:168)
	at org.knowm.xchange.binance.service.BinanceAccountService.withdrawFunds(BinanceAccountService.java:138)
	at com.xxx.yyy.MarketOperationsTest.testWithdraw(MarketOperationsTest.java:51)
	at com.xxx.yyy.MarketOperationsTest.main(MarketOperationsTest.java:34)
Caused by: org.knowm.xchange.binance.dto.BinanceException: null (HTTP status code: 0)
	at org.knowm.xchange.binance.service.BinanceAccountServiceRaw.checkApiResponse(BinanceAccountServiceRaw.java:213)
	at org.knowm.xchange.binance.service.BinanceAccountServiceRaw.withdraw(BinanceAccountServiceRaw.java:70)
	at org.knowm.xchange.binance.service.BinanceAccountServiceRaw.withdraw(BinanceAccountServiceRaw.java:48)
	at org.knowm.xchange.binance.service.BinanceAccountService.withdrawFunds(BinanceAccountService.java:160)
	... 3 more

I am testing this code on the dev branch(5.0.10-SNAPSHOT).

PR #4286, opened 2021-10-11
[binance] Fix withdraw response parsing

pvyhnal-generalbytes:
The bug seems to be introduced in wapi removal in f9e8ae29802e80bb8a2f188b37ab790bf1d55a06
The withdraw response parsing failed with: "BinanceException: null (HTTP status code: 0)"
The ApiResponse was assuming different json response format. The actual responses were added in unit test.

AssetDribbletLogResponse was using the same ApiResponse parent class (which was probably not
functioning properly) and it wasn't used anywhere so I removed that too.

### DIFF
--- a/xchange-binance/src/test/java/org/knowm/xchange/binance/AbstractResilienceTest.java
+++ b/xchange-binance/src/test/java/org/knowm/xchange/binance/AbstractResilienceTest.java
@@ -29,10 +29,6 @@ public class AbstractResilienceTest {
     return createExchange(false, true);
   }
 
-  protected BinanceExchange createExchange() {
-    return createExchange(false, false);
-  }
-
   protected BinanceExchange createExchange(boolean retryEnabled, boolean rateLimiterEnabled) {
     BinanceExchange exchange =
         (BinanceExchange)
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAuthenticated.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAuthenticated.java
@@ -300,7 +300,7 @@ public interface BinanceAuthenticated extends Binance {
    * @throws IOException
    * @throws BinanceException
    */
-  WithdrawRequest withdraw(
+  WithdrawResponse withdraw(
       @FormParam("coin") String coin,
       @FormParam("address") String address,
       @FormParam("addressTag") String addressTag,
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/WithdrawResponse.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/WithdrawResponse.java
@@ -2,10 +2,10 @@ package org.knowm.xchange.binance.dto.account;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-public final class WithdrawRequest {
+public final class WithdrawResponse {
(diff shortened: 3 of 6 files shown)


pvyhnal-generalbytes:
would you mind releasing a version after this is merged? thanks

pvyhnal-generalbytes:
Fixes #4190

xchange-binance/src/test/java/org/knowm/xchange/binance/AbstractResilienceTest.java
```diff
@@ -29,6 +29,10 @@ protected BinanceExchange createExchangeWithRateLimiterEnabled() {
     return createExchange(false, true);
   }
 
+  protected BinanceExchange createExchange() {
```
walec51, line 32: this is the same as createExchangeWithRetryDisabled and therefore redundant

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/WithdrawRequest.java
```diff
@@ -2,19 +2,14 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
-public final class WithdrawRequest extends ApiResponse<String> {
-  public final String id;
+public final class WithdrawRequest {
```
walec51, line 5: I think it would be better to rename this to `WithdrawResponse`

xchange-binance/src/test/java/org/knowm/xchange/binance/service/account/AccountServiceResilienceTest.java
```diff
@@ -0,0 +1,52 @@
...
+import java.math.BigDecimal;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+public class AccountServiceResilienceTest extends AbstractResilienceTest {
```
walec51, line 20: you are not testing any resiliency features here, please do not extend AbstractResilienceTest 

change the test name to `AccountServiceTest `

and setup you own WireMock here

walec51 (changes requested):
look ok

I have just 3 small remarks

xchange-binance/src/test/java/org/knowm/xchange/binance/AbstractResilienceTest.java
pvyhnal-generalbytes: removed

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/account/WithdrawRequest.java
pvyhnal-generalbytes: renamed

xchange-binance/src/test/java/org/knowm/xchange/binance/service/account/AccountServiceResilienceTest.java
pvyhnal-generalbytes: done

PR #4292, opened 2021-10-18
No longer true - so remove?

mdvx:
If it is available for your exchange, you may also want to use the streaming API:

```xml
<dependency>
  <artifactId>xchange-stream-XYZ</artifactId>
  <version>5.0.11</version>
</dependency>
```


mdvx:
Your right, confused.  Maybe i pulled the wrong pull request.

mdvx:
Bad PR

PR #4315, opened 2021-11-05
Removed usage of deprecated v1 Bitstamp API

tommq:
I had some issues with BTC withdrawals and ended up doing quite some changes including:

- Removing the deprecated v1 API entirely (https://www.bitstamp.net/api-deprecated/)
- Adding functionality to withdraw multiple new coins (XLM, USDC, LINK and more)

Fixes for:
- Cancel all orders: expected boolean to be returned when in reality it is JSON
- XRP deposit address: was returning Ripple IOU address instead of XRP. There are now two separate methods.


tommq:
According to emails set out by Bitstamp the v1 API will stop working on the 1.12.2021. Can someone take a look at this PR?

jtrimmer-nydig (approved):
Looks good to me. Hope this can get in before the old Bitstamp API goes dead.

xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampAccountServiceRaw.java
```diff
@@ -14,556 +9,591 @@
...
-  }
+    /**
+     * This method can withdraw any currency if withdrawal endpoint is configured in BitstampAuthenticatedV2
+     **/
+    public BitstampWithdrawal withdrawBitstampFunds(
+            Currency currency, BigDecimal amount, final String address, final String tag)
+            throws IOException {
+        BitstampWithdrawal response;
+
+        amount = amount.setScale(5, RoundingMode.HALF_UP);
```
jtrimmer-nydig, line 96: Why was this rounding to 5 decimal places added? Is there some significant figure requirement with the new API? I am not seeing this in the new Bitstamp API documentation?

jtrimmer-nydig (approved):

xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/service/BitstampAccountServiceRaw.java
jtrimmer-nydig: Super, thanks! Tested with BTC too.

PR #4316, opened 2021-11-05
"Cancel order params" provider method for generic exchange implementations.

aligokalppeker:
This PR aims to provide a generic method to provide API consumers to get which parameters are required for "cancelOrder" method.

Most of the exchange implementations realizes "cancelOrder(CancelOrderParams orderParams)". In generic implementations, it is not possible to know which parameter classes should be extended/implemented to make "cancelOrder" method work, and developers have to depend on the specific implementation.

This implementation solves this problem and enables a developer to implement its "cancelOrder" capability without depending on the specific implementation.

Default interface implementation provides the most widely used "id" parameter and We override the implementation in the required exchange implementations. We have also put an example on "Binance" to show the usage of the method.


aligokalppeker:
Any one to review ? 😀😀@earce

earce (commented):
One bit that isn't clear, this method isn't actually getting used anywhere? It's just an optional method to look up valid param classes? Because other then in demos I don't see usages of it?

aligokalppeker:
Thanks for reviewing the PR.

This method is proposed to make developers perform order cancellations without any dependency on the specific exchange API. This method is expected to be called from the consumers of the generic TradeService API. So it is not called inside the "xchange" repo implementations. 

So I think it should be expected that only usage is in the examples.

PR #4361, opened 2021-12-14
[binance-stream] Optimised subscription for orderbooks / deltas

jwisniewski:
Previous solution:
1. Fetched book snapshots on Web Socket thread, causing delays of other events (serious delays in case of multiple books + correctly implemented and used API rate limiter),
2. Fetched book snapshot firstly, then counted that it will be possible to patch the snapshot with deltas received later via Web Socket, which could lead to a necessity of "re-syncing" (downloading a new snapshot), which definitely didn't help to match exchange quota of API usage.

Solution proposed in this PR:
1. Fetches book snapshots on a dedicated thread, mitigating any delays on other, not related Web Socket events.
2. Firstly subscribes for book deltas, starts buffering them, then downloads the book snapshot and patches from applicable buffered deltas. This approach highly increases a probability, we will have all the deltas required to patch the downloaded snapshot. In practice I didn't see any re-syncing of snapshots unless the Web Socket reconnected, which invalidates the snapshots.

Disclaimer:
It is rather complicated PR. I provide the solution as it is;p don't quarantee it's bugfree for all the imagineable corner cases. It just works for me and in my use cases it makes a difference as day and night, books synchronise much faster, without this change I couldn't successfully synchronise 200+ books, it just taked ages. I would be thankful if sombody experienced in the library would review the changes and maybe, used his trusty benchmarks, testing if the change doesn't break the thing ;)

Assuming from the activity on bitrich-info/xchange-stream, I think @badgerwithagun is a good person to review these changes ;)


xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -300,41 +325,204 @@ private void initRawOrderBookUpdatesSubscription(CurrencyPair currencyPair) {
...
+    private boolean isBookInitialised() {
+      return getBook() != null;
+    }
+
+    private OrderBook getBook() {
+      return bookRef.get();
+    }
+
+    private void bufferDelta(DepthBinanceWebSocketTransaction delta) {
+      synchronized (monitor) {
```
earce, line 417: why are you synchronizing here if you already use a `ConcurrentLinkedQueue`

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -300,41 +325,204 @@ private void initRawOrderBookUpdatesSubscription(CurrencyPair currencyPair) {
...
+              delta.getLastUpdateId());
+
+          return false;
+        } else {
+          bookLastUpdateId.set(delta.getLastUpdateId());
+
+          // FIXME
+          // The underlying impl is far from being optimal, LimitOrders could be created directly.
+          extractOrderBookUpdates(currencyPair, delta)
+              .forEach(update -> bookRef.get().update(update));
```
earce, line 483: why are you synchronizing here as well, you are using atomic data structures.

It seems like the issue is that you want all this stuff to get updated atomically together, I would make more sense to create a single object here and wrap it atomically updating it with `updateAndGet `or just use synchronize and don't use atomics.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -300,41 +325,204 @@ private void initRawOrderBookUpdatesSubscription(CurrencyPair currencyPair) {
...
+      return bookRef.get();
+    }
+
+    private void bufferDelta(DepthBinanceWebSocketTransaction delta) {
+      synchronized (monitor) {
+        deltasBuffer.add(delta);
+      }
+    }
+
+    private Disposable asyncInitialiseOrderBookSnapshot() {
```
earce, line 422: initialize*

earce (changes requested):
The mix of atomics and synchronization here is unecessary, we can probably clean this up quite a bit choosing one or the other

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: To explain my thinking:
- atomic references and concurrent queue ensure thread safety on level of these particular objects (in isolation, every object is thread safe for itself)
- in our case the entire state of `OrderBookSubscription` needs to remain in sync
- there is some logic which involves multiple of these individualy thread safe objects, thus entire conditions need to be atomic, they are wrapped with `synchronized (monitor) { ...}`
- I did it more by convention:
-- Logic on multiple objects? Synchronise that with the monitor
-- Accessing one of these objects like current snapshot of an order book? Then skip `synchronisation` of the entire state of the `OrderBookSubscription`, and rely just on thread safety provided by the atomic reference:
```
private OrderBook getBook() {
   return bookRef.get();
}
```

@earce At the end. It's true, I could remove some of synchronisation, I think it would be a bit less schematic, and more prone to bugs caused by further maintainers, but its so small risk, I can modify the code as you prefer.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: See the explanation added to previous comment. Just to add on top of that, the idea is to keep book snapshot and its lastUpdateId etc, synchronised, they need to be in sync to each other.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: Sure, pushed a new commit, all "synchronise" replaced with "synchronize".

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
earce: I would prefer to have a minimum extra synchronization as possible because 1) it is not necessary 2) may be misleading or hard to understand why something unnecessary is there for future maintainers

earce: Also just to recap, I do understand the difference in synchronizations you have here (multiple operations needed to be treated atomically vs individual thread safe objects). But from a pure code perspective, we should only be doing what we need, everything else doesn't make sense.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: Update: I have simplified thread synchronisation, more in other comments.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: @earce applied Your suggestions, finally resigned from all atomic references and concurrent queue, left only the synchronised blocks, used to guard access to: book, bookLastUpdateId, deltasBuffer.

The last commit contains also a few related changes and fixed typos.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -300,41 +325,204 @@ private void initRawOrderBookUpdatesSubscription(CurrencyPair currencyPair) {
...
+    @Override
+    public boolean isDisposed() {
+      return booksSubject.hasComplete() || booksSubject.hasThrowable();
+    }
+
+    private boolean isBookInitialized() {
+      return getBook() != null;
+    }
+
+    private OrderBook getBook() {
```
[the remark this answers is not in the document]
jwisniewski: Was close to remove it, but after I have resigned from atomic references, this method was handy to keep the place when it's used simpler, after recent changes, it wraps the book reference around synchronised block.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
```diff
@@ -300,41 +320,207 @@ private void initRawOrderBookUpdatesSubscription(CurrencyPair currencyPair) {
...
+      }
+    }
+
+    @Override
+    public boolean isDisposed() {
+      return booksSubject.hasComplete() || booksSubject.hasThrowable();
+    }
+
+    private boolean isBookInitialized() {
+      synchronized (bookIntegrityMonitor) {
```
earce, line 403: just make the variable volatile, it will use memory barriers, no need to sync for this

jwisniewski:
@earce please take a look, I think I have addressed Your comments.

What is important: I suggest to keep this PR open for a few next days, if something bad would happen after the changes, and I would detect that, I will push a fix here.

earce (commented):
One more comment, also, I honestly am very allergic to synchronization blocks, it's significantly more expensive, atomics use CAS which is more efficient, also synchronization if you ever introduce another lock or even a non-re-entrant lock incorrectly can create deadlocks, just my 2 cents....

jwisniewski:
> One more comment, also, I honestly am very allergic to synchronization blocks, it's significantly more expensive, atomics use CAS which is more efficient, also synchronization if you ever introduce another lock or even a non-re-entrant lock incorrectly can create deadlocks, just my 2 cents....

@earce please keep in mind the point where we have started, without my change, books are waiting for initial snapshots blocking the thread used to deliver web socket events, delaying everything, also each particular pair can download snapshots multiple times, trying to blindly catch the situation where it will be possible to patch it from upcomming deltas. And now we are discussing the drawbacks of synchronisation blocks, which are marginal in comparison;) I'm have seen on my eyes, dramatic change in performance after the changes proposed in this PR have been applied to my project.

earce:
> > One more comment, also, I honestly am very allergic to synchronization blocks, it's significantly more expensive, atomics use CAS which is more efficient, also synchronization if you ever introduce another lock or even a non-re-entrant lock incorrectly can create deadlocks, just my 2 cents....
> 
> @earce please keep in mind the point where we have started, without my change, books are waiting for initial snapshots blocking the thread used to deliver web socket events, delaying everything, also each particular pair can download snapshots multiple times, trying to blindly catch the situation where it will be possible to patch it from upcomming deltas. And now we are discussing the drawbacks of synchronisation blocks, which are marginal in comparison;) I'm have seen on my eyes, dramatic change in performance after the changes proposed in this PR have been applied to my project.

That is fine and the improvement is much appreciated but if we are going to make a change this drastic and invasive, makes sense to do it the best way we can and set good precedents.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: The case is, `isBookInitialized()` and `getBook()` want the instantiated and fully patched, initial book snapshot. If I would use volatile book reference, the thread observing books, could access instantiated book reference, which is still being patched (in corner cases of cource, but thread safety is abaout eliminating all that corner cases).

jwisniewski: Maybe there is something you are seing but I can't grasp. I have allowed collaborators to modify my PR, if you like you should be able to propose better synchronization method.

earce: I am not following how synchronized would protect against what you are describing. You are simply asking if book is set or not, having a volatile would enforce memory barriers and the thread to read the latest state from memory. It's like an atomic operation this isn't getting updating and setting. It's just an atomic read.

earce:
I hope you are finding this review helpful I just want to make sure we do this the most efficient/correct way possible.

jwisniewski:
> > > One more comment, also, I honestly am very allergic to synchronization blocks, it's significantly more expensive, atomics use CAS which is more efficient, also synchronization if you ever introduce another lock or even a non-re-entrant lock incorrectly can create deadlocks, just my 2 cents....
> > 
> > 
> > @earce please keep in mind the point where we have started, without my change, books are waiting for initial snapshots blocking the thread used to deliver web socket events, delaying everything, also each particular pair can download snapshots multiple times, trying to blindly catch the situation where it will be possible to patch it from upcomming deltas. And now we are discussing the drawbacks of synchronisation blocks, which are marginal in comparison;) I'm have seen on my eyes, dramatic change in performance after the changes proposed in this PR have been applied to my project.
> 
> That is fine and the improvement is much appreciated but if we are going to make a change this drastic and invasive, makes sense to do it the best way we can and set good precedents.
> 
> In general this library uses observables and we aim to avoid the usage of synchronization blocks as much as possible, these act closer to actors and this starts straying away to be antipattern from the rest of the lib so I am trying to reduce how much we do this.

Understood ;) I spent a considerable amount of time, trying to solve everything reactiveX way. However failed to use rx in order to:
- buffer book update events
- at the same time try to download matching book snapshot
- then patch the initial snapshot with deltas
- then allow further events to be passed to observers of order books

And finally ended with these sync blocks... due to the main reason, that we need to sync book initialisation and all these patches, and then allow the book observable to access the book. Maybe if we would move this initial patching to the thread, where the book is observed... we could resign from most of the synchronisation, that would be possible, some unnecessary computations would be moved to the observable thread, but still, its a matter of milliseconds if not nanoseconds to apply all the buffered patches.

earce:
I do think if you left the code mostly as it was and just try and use the atomics mostly you would be fine. You will need synchronization when you want to modify 3 different objects atomically (or alternatively you could create a wrapper object and update it atomically and totally avoid synchronization).

Edge a wrapper object with
```java 
    private OrderBook book;
    private long bookLastUpdateId;
    private final Queue<DepthBinanceWebSocketTransaction> deltasBuffer = new LinkedList<>();
```
inside of an atomic called by updateAndGet(), or just how you previously had the code and a few synchronization blocks but just less of them, there's a few ways to do this, I just know this current iteration can be made better.

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingMarketDataService.java
jwisniewski: That what you are saying, can be true, but I needed to change my thinking, with sync blocks I could assign outdated snapshot to the book reference, then update it. It can be also achieved without sync blocks, just firstly update the snapshot with all the deltas, then assign to the book reference.

Whatever, I'm forking on that :)

jwisniewski:
@earce see the updated impl, I have removed all the synchronisation blocks, now everything is guarded by `volatile boolean bookInitialized` attribute.

Remarks:
- I had to restore the concurrent queue of buffered deltas, this queue is accessed from two threads
- Other, past atomics, are volatiles now, I can keep them like that or change back to atomics, what do you think?

Personally I'm a bit allergic to volatiles :D

jwisniewski:
> I don't think the current iteration is threadsafe, theres a few points here that break the threadsafety. This doesn't exactly do what I've mentioned.
> 
> Do you not prefer to wrap all these in a single atomic class. Otherwise your synchronized implementation was better. There was only a few places I had suggested removing the synchronized.
> 
> I think you original implementation was better there were just a few places we could improve it, removing sync blocks everywhere like it currently is done isn't threadsafe.

We are in a dedicated class from the start: `OrderBookSubscription`. With main purpose to sync the asynchronous initialisation of the order book, hmm?

After I have removed all the sync blocks, it should be still thread safe. All volatiles are private in a dedicated `OrderBookSubscription` class, which protects external access to its state. Thread safety is now based on:
- concurent queue of buffered book deltas
- `volatile boolean bookInitialized` which protects access to other volatile attributes, see that when we open the gate, the `bookInitialised` is set to true as the latest one, so partially filled state won't be used by other thread which correctly verifies state: `if (bookInitialised)`. Also, when we close the gate, `bookInitialised` is set to false as the first of attibures which will be cleaned up.

The key difference is, with sync blocks it was harder to break everything with minor changes in the `OrderBookSubscription` ;)

So, finally, I understand we can restore one of my intermediate implelemtations, where:
- I have removed atomics
- But still keept sync blocks

Right? I could update PR this way.

earce:
Lets go back to an earlier implementation and I can revisit thread-safety, I think we can just cut down on the sync blocks, but unless you use an atomic object we can't remove them all.

gewure:
Has there been any progress on this? Id skip elegancy for any bugfix who increases resiliency :)

jwisniewski:
@gewure According the latest suggestion from @earce, I have reverted back to one of previous solutions (months ago), I use the code in state from this PR since then, without any issues. There were also no known issues with the intermediate proposals by the way ;) I have tested quite heavily, books quantities which were just impossible to synchronise without this change, work reliably, just due to the nature of the Binance API, it takes a while to collect (respecting the API limits).

A hint: the PR is conflicted currently on an order of imports, somebody with the configured project, could fix that easily.

Issue #4372, opened 2021-12-26
[Binance] Configurable fetch binance order book limit parameter


0xfaulty:
Now when we fetch initial order book for stream receiving we can't configure 'limit' parameter, that is 1000 constantly now (BinanceMarketDataService ->fetchBinanceOrderBook), that can be not enough (can be incresed to 5000). Good idea to pass that parameter with ExchangeSpecification.

0xfaulty:
fixed by 4372

PR #4373, opened 2021-12-26
4372 Binance configurable fetch limit

0xfaulty:
Solution for issue https://github.com/knowm/XChange/issues/4372


xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
```diff
@@ -27,6 +27,7 @@
   protected static final String USE_HIGHER_UPDATE_FREQUENCY =
       "Binance_Orderbook_Use_Higher_Frequency";
   protected static final String USE_REALTIME_BOOK_TICKER = "Binance_Ticker_Use_Realtime";
+  protected static final String USE_MAX_FETCH_ORDER_BOOK_LIMIT = "Binance_Max_Fetch_Order_Book_Limit";
```
walec51, line 30: the words "use" and "max" make little sense for this parameter

just name it

FETCH_ORDER_BOOK_LIMIT = "Binance_Fetch_Order_Book_Limit"

walec51 (commented):
looks ok

just a small naming remark

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/BinanceStreamingExchange.java
0xfaulty: Does this mean that this parameter will no longer be Boolean, but will become an Integer with a specific limit value? Do I need to validate it for valid values (5, 10, 20, 50, 100, 500, 1000, 5000) in this case, or is the correctness of the value the responsibility of the api user?

0xfaulty: @walec51 to speed up the acceptance of the PR, decided to redo it before your answer. Renamed parameter name and used only if integer inside parameter.

PR #4376, opened 2022-01-02
[Kucoin] NullPointerException when retrieving currencies metadata

straumat:
With the new update of the Kucoin Sandbox API, in the `adaptCurrencyMetaData` method from `KucoinAdapters`, we get a `NullPointerException` because, for USDN currency, there is a null value for `currenciesResponse.getWithdrawalMinFee()` and `currenciesResponse.getWithdrawalMinSize()`.

With this update, I set a Zero value instead of triggering a `NullPointerException`.

### DIFF
--- a/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAdapters.java
+++ b/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAdapters.java
@@ -173,8 +173,8 @@ public class KucoinAdapters {
       CurrencyMetaData currencyMetaData =
           new CurrencyMetaData(
               precision.intValue(),
-                  withdrawalMinFee,
-                  withdrawalMinSize,
+              withdrawalMinFee,
+              withdrawalMinSize,
               walletHealth);
       stringCurrencyMetaDataMap.put(currenciesResponse.getCurrency(), currencyMetaData);
     }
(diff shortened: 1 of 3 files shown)


xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/KucoinAdapters.java
```diff
@@ -161,14 +161,20 @@ public static ExchangeMetaData adaptMetadata(
     HashMap<String, CurrencyMetaData> stringCurrencyMetaDataMap = new HashMap<>();
     for (CurrenciesResponse currenciesResponse : list) {
       BigDecimal precision = currenciesResponse.getPrecision();
-      String withdrawalMinFee = currenciesResponse.getWithdrawalMinFee();
-      String withdrawalMinSize = currenciesResponse.getWithdrawalMinSize();
+      BigDecimal withdrawalMinFee = BigDecimal.ZERO;
+      BigDecimal withdrawalMinSize = BigDecimal.ZERO;
```
walec51, line 165: this fix is ok but I think it would be better to treat unknown values as null and pass nulls to  CurrencyMetaData

straumat: Thx but i'm quite scared to what it could mean "later" with null value...

straumat: @walec51  you think I should try with:
```
BigDecimal withdrawalMinFee = null;
BigDecimal withdrawalMinSize = null;
```

straumat: @walec51 ok i changed it to null

PR #4379, opened 2022-01-04
Corrected fee currency in adaptKrakenUserTrade()

Stefan-Fritzsche:
Corected fee currency from **base** to **counter** currency in adaptKrakenUserTrade(), because the documentation of KrakenOwnTrade states that the fee relates to the "**quote** currency".

### DIFF
--- a/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
+++ b/xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
@@ -198,7 +198,7 @@ public class KrakenStreamingTradeService implements StreamingTradeService {
                 .type(KrakenAdapters.adaptOrderType(KrakenType.fromString(dto.type)))
                 .price(dto.price)
                 .feeAmount(dto.fee)
-                .feeCurrency(currencyPair.base)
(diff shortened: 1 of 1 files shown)


xchange-stream-kraken/src/main/java/info/bitrich/xchangestream/kraken/KrakenStreamingTradeService.java
```diff
@@ -198,7 +198,7 @@ else if ("stop".equals(orderType))
                 .type(KrakenAdapters.adaptOrderType(KrakenType.fromString(dto.type)))
                 .price(dto.price)
                 .feeAmount(dto.fee)
-                .feeCurrency(currencyPair.base)
+                .feeCurrency(currencyPair.counter)
```
walec51, line 201: isn't it actually dependant on the orders type?

Stefan-Fritzsche: The documentation of class KrakenOwnTrade states "_fee - total fee (quote currency)_".
And the documentation in [Kraken's Websocket API](https://docs.kraken.com/websockets/#message-ownTrades) also states "_fee - decimal - total fee (quote currency)_".

Sounds like it always relates to the qoute/counter currency of the currency pair and not to the order type.

Issue #2966, opened 2019-03-05
toString conventions


walec51:

makarid:
How can we do the 3?

nielsdraaisma:
As much as I dislike boilerplate, the other options have downsides as well.

2) As funky as Lombok looks at times it does has its issues when JVM versions changes, it more a hack than anything. If anything embracing a different compiler like kotlin and it's data classes would not have that problem.

PR #4380, opened 2022-01-06
[coinmate] stop order status, extra raw order fields

stachon:
* Determine status and price of triggered stoploss order.
* Add extra fields to raw order DTO.
* Add extra param `orderId` in getCoinmateTradeHistory


xchange-coinmate/src/main/java/org/knowm/xchange/coinmate/dto/trade/CoinmateOrderHistoryEntry.java
```diff
@@ -58,4 +92,72 @@ public BigDecimal getOriginalAmount() {
...
+        ", trailingUpdatedTimestamp=" + trailingUpdatedTimestamp +
+        ", originalStopPrice=" + originalStopPrice +
+        ", marketPriceAtLastUpdate=" + marketPriceAtLastUpdate +
+        ", marketPriceAtOrderCreation=" + marketPriceAtOrderCreation +
+        ", hidden=" + hidden +
+        ", avgPrice=" + avgPrice +
+        ", stopLossOrderId=" + stopLossOrderId +
+        ", originalOrderId=" + originalOrderId +
+        '}';
+  }
```
walec51, line 162: those toString's seam not needed

https://github.com/knowm/XChange/issues/2966

xchange-coinmate/src/main/java/org/knowm/xchange/coinmate/service/CoinmateTradeService.java
```diff
@@ -67,8 +68,22 @@ public OpenOrders getOpenOrders(OpenOrdersParams params) throws IOException {
...
     for (OrderQueryParams orderQueryParam : orderQueryParams) {
       CoinmateOrders response = this.getCoinmateOrderById(orderQueryParam.getOrderId());
-      List<Order> orders = CoinmateAdapters.adaptOrders(response);
-      result.addAll(orders);
+      Order order = CoinmateAdapters.adaptOrder(response.getData());
+      if (order.getStatus() == Order.OrderStatus.STOPPED) {
+        // fetch generated market order and get its average price and amount
+        CoinmateOrders marketOrderRaw = this.getCoinmateOrderById( response.getData().getStopLossOrderId());
+        if (marketOrderRaw != null) {
+          order.setAveragePrice(marketOrderRaw.getData().getAvgPrice());
```
walec51, line 76: it's unfortunate that some added setter to `Order` which should be immutable

this results in adaptation being done outside of the adapter

I think it would be better to make the adapter method look like this:

```
public static Order adaptOrder(CoinmateOrder coinmateOrder, Function<CoinmateOrder, String> orderByIdFetcher) {
```

move the logic below to it

and call the adapter like this

```
Order order = CoinmateAdapters.adaptOrder(response.getData(), orderId -> this.getCoinmateOrderById(orderId));
```

xchange-coinmate/src/main/java/org/knowm/xchange/coinmate/dto/trade/CoinmateOrderHistoryEntry.java
stachon: @walec51 removed

xchange-coinmate/src/main/java/org/knowm/xchange/coinmate/service/CoinmateTradeService.java
stachon: @walec51 I have modified it with it with Function. One downside is that `Fuction` cannot throw so I added a catch block and return `null` instead. Is there a better solution?

timmolter: Agreed. it should be immutable.

PR #4381, opened 2022-01-06
Make CurrencyPair equals use instanceof instead of class. Allows use …

jaj2276:
CurrencyPair::equals() uses a conditional check of "getClass() != obj.getClass()". If this conditional is true, equals() returns false.

If a user wants to supply a CurrencyPair-derived class, Binance (and maybe others) will fail to deliver data. They fail because they are comparing their own copies of CurrencyPair objects with CurrencyPair-derived objects and CurrencyPair.getClass() != UserDerivedCurrencyPair.getClass() will fail.

By changing the conditional check _**getClass() != obj.getClass()**_ to **_!(obj instanceof CurrencyPair)_**, the CurrencyPair::equals method can handle equality against user derived classes of CurrencyPair.

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/currency/CurrencyPair.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/currency/CurrencyPair.java
@@ -483,7 +483,7 @@ public class CurrencyPair extends Instrument implements Comparable<CurrencyPair>
     if (obj == null) {
       return false;
     }
-    if (getClass() != obj.getClass()) {
+    if (!(obj instanceof CurrencyPair)) {
       return false;
     }
     CurrencyPair other = (CurrencyPair) obj;


walec51:
what you are proposing is a bad practice which would mage equals not symmetric

read more here:

https://docs.oracle.com/javase/6/docs/api/java/lang/Object.html#equals(java.lang.Object)

https://stackoverflow.com/questions/12239344/why-should-i-not-use-equals-with-inheritance

btw. a currency pair is an attribute of an derivative, it should not be its super class

take a look at this:

https://github.com/knowm/XChange/blob/develop/xchange-core/src/main/java/org/knowm/xchange/derivative/FuturesContract.java

jaj2276:
Hey Adam,

Thanks for the feedback.

-- Clearly my suggestion breaks the equals() method symmetric property. In
fact your second link to StackOverflow describes as much. Although I will
offer that the "solution" provided in that StackOverflow thread states
CurrencyPair's current implementation of CurrencyPair::equals() is "bad" as
it prevents one from doing exactly what I'm attempting to do (i.e. adding a
trivial amount of data to the original class that has no bearing on the
equality part of the class).

-- Regarding your FuturesContract/CurrencyPair link, that doesn't really
help me (although maybe that wasn't your intent!) with the problem I'm
trying to solve. I understand I can compose a class that contains a
CurrencyPair (i.e. Has-A relationship as opposed to Is-A relationship) but
the methods for creating a product subscription (used when calling
connect()) and/or calling OrderBook::subscribe() take a CurrencyPair. I
could do the same with my class (i.e. create a class that contains a
currency pair), but I will still only be given a CurrencyPair in the onNext
method which will require an additional lookup.

Let me state my issue and maybe there's something else I could/should be
doing.

When I get a callback in my onNext consumer, I would like to NOT have to do
an additional lookup for user-data. This is often achieved in many other
market data APIs via some sort of userData object that gets passed in via
the subscription process. There doesn't seem to be a way for me to achieve
that with the XChange API. My solution was to simply extend CurrencyPair -
ExtendedCurrencyPair - that adds a single reference that points to my
user-data.

In my code I only subscribe using ExtendedCurrencyPairs so a downcast from
CurrencyPair to ExtendedCurrencyPair on the call to
orderBook.getBids().get(0).getInstrument() is all that's needed for me to
access my user-defined data (i.e. no additional lookup).

This mechanism works fine for both FTX and COINBASEPRO exchanges. The
CurrencyPair objects it returns to me are the same ones I supplied it with
when connecting/subscribing. It does not work for Binance and it appears to
not do so because Binance creates a new/different CurrencyPair object as a
key in filtering out updates instead of simply using my
ExtendedCurrencyPair object (which itself is a CurrencyPair).

Ok, with that out of the way, do you have any suggestions on how to solve
my actual problem? Is there a mechanism when subscribing to provide a
user-supplied resource that will be given to the user when the onNext
method is called?

Thanks,
Josh

PR #4383, opened 2022-01-08
[Binance] Fixed mapping of binance trade "isBuyerMaker" to ASK/BID order types (reversed it...)

jwisniewski:
This PR reverses the trade type mapping for Binance market...

I have cross checked a few other markets, how they map trade side reported by exchange to XChange OrderType.BID/ASK.

The mapping seems to be: 
- BUY from exchange → OrderType.BID
- SELL from exchange → OrderType.ASK

However Binance doesn’t report BUY/SELL information, just the boolean flag if the buyer was the market maker.

I had good feeling that the trade type should be from the taker perspective. It is the taker, who initiates the trade. My understanding has been confirmed at last here:
https://money.stackexchange.com/a/102005

But I really wonder, why nobody detected this issue earlier. It is quite major, to reverse the trade side… I hope I misunderstood something, and this PR is not necessary;)

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAdapters.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceAdapters.java
@@ -121,8 +121,18 @@ public class BinanceAdapters {
     }
   }
 
-  public static OrderType convertType(boolean isBuyer) {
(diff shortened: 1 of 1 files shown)


jwisniewski:
Closed the PR, maybe the source I have found was wrong... will test it more deeply.

PR #4386, opened 2022-01-13
coinmate getFundingHistory add support for querying by dates

kubacech:
adding support for timestampFrom and timestampTo for querying coinmate funding history


xchange-coinmate/src/main/java/org/knowm/xchange/coinmate/service/CoinmateAccountService.java
```diff
@@ -143,9 +146,19 @@ public TradeHistoryParams createFundingHistoryParams() {
       order = ((TradeHistoryParamsSorted) params).getOrder();
     }
 
+    if (params instanceof TradeHistoryParamsTimeSpan) {
```
walec51, line 149: looks ok

does CoinmateTradeHistoryHistoryParams also implement this interface?

if not it would also be good to add it there

kubacech: It does and it is already handled in CoinmateTradeService.getTradeHistory() - https://github.com/knowm/XChange/blob/f36daa0825ae02b369a47ba0b7e70046d95feb9f/xchange-coinmate/src/main/java/org/knowm/xchange/coinmate/service/CoinmateTradeService.java#L219-L227  Its just missing here.

PR #4392, opened 2022-01-19
Issue gemini get tickers

straumat:
getTicker(Instrument instrument) is not overried in Gemini so it returns a notYetImplemented exception but getTicker(CurrencyPair) is available ! so just making a bridge:

```
    @Override
    public Ticker getTicker(Instrument instrument, Object... args) throws IOException {
        return getTicker((CurrencyPair) instrument, args);
    }
```


timmolter:
Could it be you're not using the same formatter as we are? It's hard to manage the diffs with different formatting.

straumat:
@timmolter sorry! i will create a new PR but on the global project. Thx

PR #4393, opened 2022-01-22
GEMINI: Cancel all active orders

gmax0:


timmolter:
`cancelAllOrders` would  be a better name IMO.

gmax0:
@timmolter I'll open another PR to modify the method names later today, looks like I missed the boat :(

gmax0:
Didn't realize you already took care of it. Thanks @timmolter!

PR #4403, opened 2022-02-04
Bugfix/etd 235 connectors coinbase

martinkyov:
- Bug fix - Null point exception in case of cancelled transactions


walec51:
@martinkyov could you simplify this according to @earce advice?

martinkyov:
Hi, yes, I could do that next week!

PR #4408, opened 2022-02-15
[Binance] Added support for the future websocket stream

Scrier:
Added support for the future stream.

Quick and dirty to change the endpoint. And as I didn't find another way to instantiate with URL I just created a futures websocket url.


Scrier:
Bit trigger happy.

PR #4413, opened 2022-02-23
Implementation for Upward heartbeat on private

mdvx:
if you send a limit order, and it stays alive for more than 45 seconds, this needs to be called (or it will be cancelled).


mdvx:
@earce

mdvx:
Final piece of the puzzle, You can monitor downward heartbeats in the private channel, and send this upward heartbeat every 15 seconds.  
* downward comes via websocket
* upward is REST call

PR #4429, opened 2022-03-12
Code format on PR

kennethjor:
I don't know if this is something you'd consider doing, but it would keep things consistent.


kennethjor:
Not sure why this is failing. It could be because the repo isn't allowing writes from Actions, perhaps?

kennethjor:
Alternatively, this could be configured to just fail if there are any code formatting issues.

kennethjor:
Closing as there has been no movement on it.

PR #4439, opened 2022-03-15
[kraken] kraken deposit addresses

donald-jackson:
This pull request adds support for protocols and tags to Kraken

### DIFF
--- a/xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenUtils.java
+++ b/xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenUtils.java
@@ -54,11 +54,18 @@ public class KrakenUtils {
     }
   }
 
+  public static String getAltName(String key, KrakenAsset value) {
+    if(key.equals("XBT")) {
+      return "BTC";
+    }
+    return value.getAltName();
+  }
+
   public static void setKrakenAssets(Map<String, KrakenAsset> assetSource) {
     if (assetsMap.isEmpty()) {
       for (Map.Entry<String, KrakenAsset> entry : assetSource.entrySet()) {
-        assetsMap.put(entry.getKey(), Currency.getInstance(entry.getValue().getAltName()));
-        assetsMapReverse.put(Currency.getInstance(entry.getValue().getAltName()), entry.getKey());
+        assetsMap.put(entry.getKey(), Currency.getInstance(getAltName(entry.getKey(), entry.getValue())));
+        assetsMapReverse.put(Currency.getInstance(getAltName(entry.getKey(), entry.getValue())), entry.getKey());
       }
     }
   }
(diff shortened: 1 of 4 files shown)


xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenAccountService.java
```diff
@@ -92,50 +90,108 @@ public String withdrawFunds(WithdrawFundsParams params) throws IOException {
...
+
+    Currency krakenCurrency = Currency.getInstance(KrakenUtils.getKrakenCurrencyCode(currency));
+
+    KrakenDepositAddress[] depositAddresses = requestDepositAddress(false, krakenCurrency, protocol);
+    if(depositAddresses == null || depositAddresses.length <= 0) {
+      depositAddresses = requestDepositAddress(true, krakenCurrency, protocol);
+    }
+
+    if(depositAddresses == null || depositAddresses.length <= 0) {
+      throw new IOException("Unable to create wallet currency: "+currency+" protocol: "+protocol);
```
walec51, line 180: this and many other places have wrong formatting 

build your code using maven - it will automaticaly fix it to our guidelines

xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenUtils.java
```diff
@@ -54,11 +54,18 @@ public static void setKrakenAssetPairs(Map<String, KrakenAssetPair> pairs) {
     }
   }
 
+  public static String getAltName(String key, KrakenAsset value) {
+    if(key.equals("XBT")) {
+      return "BTC";
+    }
```
walec51, line 60: won't this break existing code that already assumes that XBT is the code for bitcoin on this exchange

please read this: https://github.com/knowm/XChange/issues/2394

until an universal solution is found we should assume that we operate on currency code as they are used on a given exchange

xchange-kraken/src/main/java/org/knowm/xchange/kraken/service/KrakenAccountService.java
```diff
@@ -92,50 +90,108 @@ public String withdrawFunds(WithdrawFundsParams params) throws IOException {
...
+
+    if(depositMethods == null || depositMethods.length <= 0) {
+      return null;
+    }
+
+    if(depositMethods.length > 1 && protocol == null) {
+      LOGGER.warn("Multiple methods for currency "+currency+" protocol "+protocol);
+      for(KrakenDepositMethods method : depositMethods) {
+        LOGGER.warn(" -- "+method.getMethod());
+      }
```
walec51, line 104: change the above to `LOGGER.warn("Multiple methods for currency {} protocol {}, returning first method from: {}", currency, protocol, depositMethods);`

walec51:
any progress on this PR?

PR #4443, opened 2022-03-17
Ftx hotfix: Server time format did not match local time format

Svenmet:
The FTX Exchange sends e.g. "40000**.**0" as the bitcoin price in the orderbook.
But my local time format is "40000**,**0", so the orderbook checksum was wrong and I could not receive the orderbook.
This hotfix solve the problem :-)

### DIFF
--- a/xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
+++ b/xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
@@ -29,11 +29,17 @@ import org.knowm.xchange.ftx.FtxAdapters;
 import org.knowm.xchange.ftx.dto.marketdata.FtxTradeDto;
 import org.knowm.xchange.ftx.dto.trade.FtxOrderFlags;
 import org.knowm.xchange.instrument.Instrument;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class FtxStreamingAdapters {
 
+  private static final Logger LOG = LoggerFactory.getLogger(FtxStreamingAdapters.class);
+
   private static final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
-  /** Incoming values always has 1 trailing 0 after the decimal, and start with 1 zero */
+  /**
+   * Incoming values always has 1 trailing 0 after the decimal, and start with 1 zero
+   */
   private static final ThreadLocal<DecimalFormat> dfp =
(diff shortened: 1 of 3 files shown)


xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -110,6 +118,8 @@ public static OrderBook adaptOrderbookMessage(
                     getOrderbookChecksum(orderBook.getAsks(), orderBook.getBids());
 
                 if (!calculatedChecksum.equals(message.getChecksum())) {
+                  LOG.error("calculated checksum: {}", calculatedChecksum);
+                  LOG.error("message checksum:    {}", message.getChecksum());
```
walec51, line 122: please add the above information to the exceptions message

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -61,6 +67,8 @@ public static OrderBook adaptOrderbookMessage(
         .forEach(
             message -> {
               if ("partial".equals(message.getAction())) {
+                orderBook.getAsks().clear();
+                orderBook.getBids().clear();
```
walec51, line 71: this is an input to an adapter - its not something you should modify

xchange-stream-ftx/src/main/java/info/bitrich/xchangestream/ftx/FtxStreamingAdapters.java
```diff
@@ -29,11 +29,17 @@
...
 
+  private static final Logger LOG = LoggerFactory.getLogger(FtxStreamingAdapters.class);
+
   private static final ObjectMapper mapper = StreamingObjectMapperHelper.getObjectMapper();
-  /** Incoming values always has 1 trailing 0 after the decimal, and start with 1 zero */
+  /**
+   * Incoming values always has 1 trailing 0 after the decimal, and start with 1 zero
+   */
   private static final ThreadLocal<DecimalFormat> dfp =
       ThreadLocal.withInitial(() -> new DecimalFormat("0.0#######"));
```
walec51, line 44: this is the place where this issue should be fixed 

https://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormat.html

we should force the proper separator here instead of fixing in with string replace

walec51:

PR #4448, opened 2022-03-30
Kraken - Rate API LIMITS

martinkyov:
I created new PR to solve RATE API LIMITS "https://docs.kraken.com/rest/#section/Rate-Limits". 
Solution uses thread sleep for 3s and max 9 requests.


earce:
Should we pull this into first class rate limiting lib?

walec51:
please read this to see how to implement rate limiting: https://github.com/knowm/XChange/wiki/Implementing-resiliency-features

timmolter:
We don't do the retries in this library like you implemented.

PR #4461, opened 2022-04-13
CoinbasePro streaming exchange: NPE if disconnect() method is called twice

eschava:


eschava:
please, restart the checks. I guess that everything should be OK now

eschava:
please rerun the checks

PR #4463, opened 2022-04-17
[Feature] Candle data support

aligokalppeker:
This PR includes basic data structures and interfaces for fetching candlestick data from an exchange. In addition, it includes candlestick Implementations for two exchanges, poloniex and okex(more impl. will come)


aligokalppeker:
Anyone to progress the review? This and many other PRs are waiting for review? @earce @walec51

timmolter:
somehow you broke unit test(s) it looks like. Please fix first.

aligokalppeker:
@timmolter We did not broke it. It was already broken in the master. Other PRs are seems broken due to this problem. It is fixed in the master, if you can re-trigger the build process, it should be corrected with no commit

timmolter:
This one worked: https://github.com/knowm/XChange/pull/4478

Also, I just pulled develop and ran the unit tests with `mvn clean package`. It succeeded.

aligokalppeker:
I have merged with upstream, it should be fixed now. 👍

xchange-core/src/main/java/org/knowm/xchange/service/marketdata/MarketDataService.java
```diff
@@ -177,4 +179,25 @@ default Trades getTrades(Instrument instrument, Object... args) throws IOExcepti
...
+   * request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the
+   * requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the
+   * requested function or data, but it has not yet been implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  default CandleStickData getCandleStickData(CurrencyPair currencyPair, Date startDate, Date endDate, Object... args) throws IOException {
+    throw new NotYetImplementedForExchangeException("getCandleStickData");
+  }
```
walec51, line 202: why must we add `CandleStick` when we have `Ticker` which has essentially the same fields?

why can't you use this existing interface method

```
List<Ticker> getTickers(Params params) throws IOException {
```

to implement this feature

timmolter: I was just going to ask the same thing. ;)

aligokalppeker: At different generic exchange library implementations(like ccxt), ticker and candle data and methods are designed separately.

We should add a new method because exchange sites have separate services for tickers and candlesticks.

ticker and candle data are separate concepts. They may seem similar due to that they show information about a currency trading pair. But in fact, the candlestick is temporal information with many additional fields, such as high, low price, and open, close price. And also it has a definite time frame.

image

Actually, we assume it has been modeled better in exchange, but we see that lots of concepts are missed or modeled incorrectly. So we did our best to model it appropriately.

aligokalppeker: Any comments? @walec51 @timmolter

walec51: https://www.investopedia.com/terms/s/stocksymbol.asp actually describes a concept more align with a Currency Symbol and not out ticker class

the concept of a candleStick is very similar as the "ticker" term in our code - the "ticket" could be described as the last 24h candle stick for a given currency pair

however I'm starting to see and acknowledge we need a separate model as the method I've suggested is usually used to get those tickers for multiple currency pairs and not for many intervals of one currency pair

so taking this to account adding a `CandleStick` model makes sense and in the future the `Ticker` model could look like this

```
Ticker {
  Instrument instrument;
  CandleStick lastCandleStick; 
  // ...
}
```

that being said I think you should try to keep names of fields your `CandleStick` class in the same convention as they are seen in `Ticker`

```
  private final BigDecimal open;
  private final BigDecimal last; // would be also useful to add
  private final BigDecimal high;
  private final BigDecimal low;
  private final BigDecimal volume;
  private final BigDecimal quoteVolume; // instead of currencyVolume
  private final BigDecimal vwap; // these 5 fields can be null if not provided by the exchange
  private final BigDecimal bid; 
  private final BigDecimal bidSize;
  private final BigDecimal ask;
  private final BigDecimal askSize;
  private final Date timestamp; // instead of date
```

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/CandleStick.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @JsonPOJOBuilder(withPrefix = "")
+    public static class Builder {
+        protected Date date;
+        protected BigDecimal open;
+        protected BigDecimal high;
+        protected BigDecimal low;
+        protected BigDecimal close;
+        protected BigDecimal volume;
+        protected BigDecimal currencyVolume;
```
walec51, line 70: why not private?

xchange-core/src/main/java/org/knowm/xchange/service/marketdata/MarketDataService.java
```diff
@@ -177,4 +179,25 @@ default Trades getTrades(Instrument instrument, Object... args) throws IOExcepti
...
+   * @return The CandleStickData, null if some sort of error occurred. Implementers should log the error.
+   * @throws ExchangeException - Indication that the exchange reported some kind of error with the
+   * request or response
+   * @throws NotAvailableFromExchangeException - Indication that the exchange does not support the
+   * requested function or data
+   * @throws NotYetImplementedForExchangeException - Indication that the exchange supports the
+   * requested function or data, but it has not yet been implemented
+   * @throws IOException - Indication that a networking error occurred while fetching JSON data
+   */
+  default CandleStickData getCandleStickData(CurrencyPair currencyPair, Date startDate, Date endDate, Object... args) throws IOException {
```
walec51, line 200: looking at your implementations I think that doing this method in this convention would be more generically usable

```
CandleStickData getCandleStickData(CurrencyPair currencyPair, CandleStickDataParams params) throws IOException;

CandleStickDataParams createCandleStickDataParams();
```

please take a look how `TradeHistoryParams` are implemented for an example

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/CandleStick.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private final Date date;
+    private final BigDecimal open;
+    private final BigDecimal high;
+    private final BigDecimal low;
+    private final BigDecimal close;
+    private final BigDecimal volume;
+    private final BigDecimal currencyVolume;
+
+    public CandleStick(Date date,BigDecimal open, BigDecimal high, BigDecimal low, BigDecimal close, BigDecimal volume, BigDecimal currencyVolume) {
```
walec51, line 24: code formatting is not according to our conventions

please build using maven - the autoformatter will fix this

walec51 (changes requested):

xchange-core/src/main/java/org/knowm/xchange/service/marketdata/MarketDataService.java
aligokalppeker: Converted signature to use CandleStickParams.
But I did not implement the createCandleStickDataParams() method which seems unnecessary this time.

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/CandleStick.java
aligokalppeker: Formatted.

xchange-core/src/main/java/org/knowm/xchange/dto/marketdata/CandleStick.java
aligokalppeker: Converted to private

xchange-core/src/main/java/org/knowm/xchange/service/marketdata/MarketDataService.java
aligokalppeker: Added new fields and renamed the ones with the suggested alternatives.

aligokalppeker:
All requests are implemented. @walec51

aligokalppeker:
Anyone? @walec51 ?

walec51 (approved):

kennethjor:
I know this is an old PR, but was there a reason this was added using `Date` and `int` instead of `Instant` and `Duration`, respectively?

PR #4468, opened 2022-04-21
Binance streaming rate limit

eschava:
Binance closes the WebSocket connection if more than one per second subscription message is sent
(even though Binance docs state that 5 messages per second are allowed)


gewure:
has this been fixed elsewhere? 
the mergeconflicts should be resolveable. i noticed similiar behaviour with binance streams in the past.

rizer1980:
Hi.
1. You can use subscription on connect, there is no limit on it.
2. If I remember correctly, in reality there are 4 op/sec without error...

PR #4475, opened 2022-05-02
Binance: margin trading

eschava:
Adds new order flags to send margin order instead of spot one.
To cancel/retrieve margin order either custom order param structures could be used or account type could be included in order ID (if Include_Margin_Account_Type_In_OrderId exchange parameter is set to true)


eschava:
If this approach is accepted then I would like to add support for margin trading to BinanceStreamingExchange. But I do not see examples where connect method accepts custom product subscription classes. And this is required because there should be two additional subscription types for two additional channels: margin trades and isolated margin trades

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/margin/MarginSideEffectType.java
```diff
@@ -0,0 +1,19 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import org.knowm.xchange.dto.Order;
+
+public enum MarginSideEffectType implements Order.IOrderFlags {
+  NO_SIDE_EFFECT,
+  MARGIN_BUY,
+  AUTO_REPAY;
+
+  @JsonCreator
+  public static MarginSideEffectType getSideEffectType(String s) {
```
walec51, line 12: this method seams pointless as jacksons standard error handling will do a much at reporting an unknown value

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
```diff
@@ -206,23 +205,42 @@ public boolean cancelOrder(String orderId) {
...
+                null,
+                null);
+      }
       return true;
     } catch (BinanceException e) {
       throw BinanceErrorAdapter.adapt(e);
     }
   }
 
+  private MarginAccountType getMarginAccountTypeFromOrderId(String orderId) {
```
walec51, line 239: using order id to smuggle the margin accounts id look hacky

is this some kind of convention that comes from binance?

could you point to the documentation where its described?

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/margin/MarginSideEffectType.java
eschava: Agree, but this is how it's implemented for other Binance enums. So I just copied the same approach

xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceTradeService.java
eschava: This is not a part of official documentation. It's a hack added by me to allow using regular cancelOrder/getOrder API for margin orders without any special parameter types. And it's optional though

PR #4508, opened 2022-06-09
Fixing build problems with Maven

scuevas-bc:


scuevas-bc:
Hi @timmolter already checked https://github.com/knowm/XChange/runs/6810546690?check_suite_focus=true
a .json file was lost and that's why a test was failing, it's been corrected.

PR #4509, opened 2022-06-14
[Ftx] Adding volumeScale support + liquidationPrice on openPositions + feeUsd field on BorrowingHistoryDto

makarid:


imTigger:
The time parameter seems does not work

FTX API accept seconds while Date.getTime() return millseconds

PR #4510, opened 2022-06-15
Adding new exchange: Bitrue; fix OKX v5 ticker

dust:
1. add tickers in OkexMarketDataService


xchange-stream-coinbasepro/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.knowm.xchange</groupId>
         <artifactId>xchange-parent</artifactId>
-        <version>5.0.14-SNAPSHOT</version>
+        <version>5.0.12</version>
```
timmolter, line 9: This shouldn't be set like this

timmolter:
Can you please change the title of this to match the contents of the PR?

xchange-stream-coinbasepro/pom.xml
dust: Sorry, I made a temporary change and forgot to back out.

Now, Fixed.

PR #4527, opened 2022-07-12
[Core] Fixing an issue with UserTrades serialization/deserialization

makarid:
Hello, everyone there is a issue with UserTrades class when someone wants to parse it to json and then deserialize it again. If you run the below code you will see that even though the List<UserTrade> has 1 element, the deserialization creates the UserTrades class with 2 elements (dublicate the element). The issue is fixing when i rename the getUserTrades() method to userTrades(). If you have a better solution please tell me in order to implement it.

`String userTradesString = new ObjectMapper().writeValueAsString(new UserTrades(
                    List.of(new UserTrade.Builder().build()),
                    Trades.TradeSortType.SortByTimestamp));
    System.out.println(userTradesString);
    UserTrades userTrades = new ObjectMapper().readValue(userTradesString, UserTrades.class);
    System.out.println(userTrades.userTrades());`

I have also corrected the timestamp on adaptUserTrades on FtxStreamingExchangeAdapter in order to have the time of the exchange.


walec51:
changing our API to make jackson serialization work?

you can use @JsonIgnore on getUserTrades

makarid:
Ok, I will do that.

makarid:
I have put the JsonIgnore and it works.Can you please check it.Thanks

PR #4530, opened 2022-07-19
fix for Huobi withdraw fee (SOLO currency)

atrcek:


timmolter:
Is it really zero fee if it's returned null?

atrcek:
Yeah, it's more like "not available". This is the currency in question:
`{
	"currency": "solo",
	"assetType": 1,
	"chains": [
		{
			"chain": "solo",
			"displayName": "SOLO",
			"fullName": "XRP",
			"baseChain": "XRP",
			"baseChainProtocol": "XRPTOKEN",
			"numOfConfirmations": 1,
			"numOfFastConfirmations": 1,
			"depositStatus": "allowed",
			"minDepositAmt": "5",
			**"withdrawStatus": "prohibited",**
			"minWithdrawAmt": "5",
			"withdrawPrecision": 7,
			"maxWithdrawAmt": "9911613.000000000000000000",
			"withdrawQuotaPerDay": "9911613.000000000000000000",
			"withdrawQuotaPerYear": null,
			"withdrawQuotaTotal": null,
			"withdrawFeeType": "ratio",
			**"transactFeeRateWithdraw": "0.0002",**
			"addrWithTag": true,
			"addrDepositTag": true
		}
	],
	"instStatus": "normal"
}`

Without this fix, service using the lib crashes at init, so this is a quick patch for the problem. I didn't want to set null, to avoid problems in other parts of code.
If anybody wants to use SOLO and requires info about withdraw fee, there are bigger changes needed in xchange-core - withdraw fee ratio (percent) would need to be added to meta data and possibly other places (note that currently SOLO is disabled for withdraw).

walec51:
I think ` BigDecimal transactFeeWithdraw = huobiCurrency.getTransactFeeWithdraw() != null ? new BigDecimal(huobiCurrency.getTransactFeeWithdraw()) : null;` would be a better solution as we use null to indicate unknown values

if you could change it to that that resolve the above conflict then I'll merge it this week

PR #4533, opened 2022-07-23
moving stream-gateio to a package

pragmaxim:
`stream-gateio` is for some reason without `info.bitrich.xchangestream` package as the others. The reason for fixing this is because in Scala/SBT, root namespace classes from an imported library are missing in classpath as it is some kind of violation of best practices.


bigscoop:
is done in https://github.com/knowm/XChange/pull/4766 please close the PR

PR #4502, opened 2022-06-03
Moved getTradeHistoryForSubaccount() to use /fills

imTigger:
Change for #4496 

----------

Moved `getTradeHistoryForSubaccount()` to use that `/fills?market={market}`
DTO was added to process that API response.

Before:
`Pulled Trade: UserTrade[type=BID, originalAmount=0.001, instrument=ETH/USD, price=1752.0, timestamp=Sat Jun 04 02:06:55 HKT 2022, id=151761986698, orderId='151761986698', feeAmount=null, feeCurrency='null', orderUserReference='98d3e174-8d69-4431-89ba-2e9c7755831e']
`

After:
`Pulled Trade: UserTrade[type=BID, originalAmount=0.001, instrument=ETH/USD, price=1752.0, timestamp=Sat Jun 04 02:07:30 HKT 2022, id=8434794907, orderId='151761986698', feeAmount=-5E-9, feeCurrency='ETH', orderUserReference='null']`

 `fee`, `feeCurrency` and `id` now consitent with `getStreamingTradeService().getUserTrades()` callback.

Reference: Callback result by `getStreamingTradeService().getUserTrades()`
`Trade callback: UserTrade[type=BID, originalAmount=0.001, instrument=ETH/USD, price=1751, timestamp=Sat Jun 04 02:37:22 HKT 2022, id=8434996155, orderId='151767264661', feeAmount=-5E-9, feeCurrency='ETH', orderUserReference='0a7bd422-86bb-46cc-b1a2-b5645228f57d']`

----------

Added new method `getOrderHistoryForSubaccount()` to `FtxTradeServiceRaw`, 
using the original `/orders/history` API to retrive order history, 
which including both open and closed orders.

The return should be consitent with `getStreamingTradeService().getOrders()` callback.

----------

Personally I use those two APIs to recover orders/trades missed during websocket disconnection.

PR #4535, opened 2022-07-26
[FTX] Fix getOrder/TradeHistoryForSubaccount parameters

imTigger:
FTX API accept seconds while Date.getTime() return millseconds, #4509

Also fixed a merge error incorrectly casting TradeHistoryParamCurrencyPair to TradeHistoryParamsAll introduced in #4502

### DIFF
--- a/xchange-ftx/src/main/java/org/knowm/xchange/ftx/service/FtxTradeServiceRaw.java
+++ b/xchange-ftx/src/main/java/org/knowm/xchange/ftx/service/FtxTradeServiceRaw.java
@@ -172,43 +172,49 @@ public class FtxTradeServiceRaw extends FtxBaseService {
     }
   }
 
-  public OpenOrders getOrderHistoryForSubaccount(String subaccount, TradeHistoryParams params)
-          throws IOException {
-    if (params instanceof TradeHistoryParamCurrencyPair) {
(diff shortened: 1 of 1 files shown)


xchange-ftx/src/main/java/org/knowm/xchange/ftx/service/FtxTradeServiceRaw.java
```diff
@@ -172,43 +172,49 @@ public FtxResponse<List<FtxOrderDto>> getFtxOpenOrders(String subaccount, String
     }
   }
 
-  public OpenOrders getOrderHistoryForSubaccount(String subaccount, TradeHistoryParams params)
-          throws IOException {
-    if (params instanceof TradeHistoryParamCurrencyPair) {
+  public OpenOrders getOrderHistoryForSubaccount(String subaccount, TradeHistoryParams params) throws IOException {
+    if (params instanceof TradeHistoryParamsAll) {
```
walec51, line 176: `TradeHistoryParamsAll` implements way more parameters then this implementation supports

I think we should stick to checking individual interfaces like `TradeHistoryParamCurrencyPair`

walec51:

imTigger:
No, unlikely it will be useful anymore due to FTX's situation ☹️

PR #4540, opened 2022-08-07
Fix Poloniex after API changes for returnChartData

alinbreaban:
Poloniex API for returnChartData now returns only 100 candles + it returns millis instead of seconds for the time of the candle.
We are fixing millis - but anyone using the data needs to improve the fetching inside their project


alinbreaban:
This is fixing the following issue: [https://github.com/knowm/XChange/issues/4541](url)

PR #4546, opened 2022-08-14
[Huobi] Fixing instantiation issue when transactionWithdrawFee is null

makarid:


eschava:
new FtxMarketDataTest always fails the build

PR #4551, opened 2022-08-16
Fix Huobi NullPointerException bug

vangogiel:
- Jackson library is instructed to deserialise `transactFeeWithdraw` into `BigDecimal` directly to avoid `NullPointerException` issue.


vangogiel:
issue with commit signing

PR #4570, opened 2022-09-06
Threads are leaking if disconnecting service that was closed by the server

eschava:
If reconnection is handled by the client code every call of disconnect() and connecting again creates a new thread while the old one still lives


eschava:
please rerun the checks

eschava:
please rerun the checks

PR #4571, opened 2022-09-07
Kraken: default order book size is 10 (not 25)

eschava:
Otherwise, checksum calculation fails too often (if the server book doesn't match local one)


eschava:
> why would checksum calc fail if subscription logic is correct?

Let me describe the issue by example. E.g. we have asks with the next prices:
1, 2, 3, 4, 5, 6, 7, 8, 9, 10
And receive an update that says that a new price of 2.5 appeared. So order book is the next:
1, 2, 2.5, 3, 4, 5, 6, 7, 8, 9, 10.
10 is not in the top 10 so it is still in the book but not included in the checksum calculation.
And now we receive an update that removes 2.5 price and adds 11. So order book changes to
1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11
(11 is excluded from the checksum calculation) 

But this is incorrect because there is no quote with the price of 10 anymore. As soon as it goes out of the top 10 quotes we need to remove it from the internal structures to have the same top 10 order book as at the exchange. So correct order book has to be:

1, 2, 3, 4, 5, 6, 7, 8, 9, 11

timmolter:
Please fix conflicts. I will merge and push a release.

eschava:
done

PR #4597, opened 2022-10-13
[Okex] Move the OKEX streaming exchange to new module

makarid:


timmolter:
why?

makarid:
Because we cannot call the exchange using class string and in order to be
more organized. There is a confict when you call the exchange using class
string as parameters.

makarid:
I change the name of the module from okx to okex in order to be consistent with xchange-okex module

PR #4602, opened 2022-10-23
[okx] transfer funds between accounts, and add frozen to balances

mdvx:


xchange-core/src/main/java/org/knowm/xchange/dto/account/Balance.java
```diff
@@ -561,7 +561,7 @@ public Balance build() {
       if (frozen == null) {
         if (total == null || available == null) {
           frozen = BigDecimal.ZERO;
-        } else if (total != null) {
```
timmolter, line 564: we don't need a null check here?

timmolter:
it still fails. Something in bitthumb module. Maybe your changes to core broke something.

eschava:
please adjust package (remove v5) and add the next to the OkexAssetTransferRequest:
@JsonInclude(JsonInclude.Include.NON_NULL)

I get an exchange error if null values are serialized in the request body

PR #4622, opened 2022-11-22
[Kucoin] Add liquidity field to UserTrade

zsombok:
Kucoin API provides liquidity field in response, which is consumed by TradeResponse, but this information is truncated when converted to UserTrade. This field is added to UserTrade and KucoinAdapters builder in order to carry on maker-taker info.


timmolter:
Would an enum be more appropriate? Do other exchanges provide this info?

zsombok:
@timmolter could you please review this PR?

timmolter:
How many exchanges support this distinction? It looks like very very few.

PR #4635, opened 2022-12-13
[okex] Adding perpetual swap support

makarid:


.gitignore
```diff
@@ -43,3 +43,5 @@ GTAGS
 GRTAGS
 GPATH
 GSYMS
+/xchange-okex/src/test/resources/keys.properties
```
timmolter, line 46: don't we have a global ignore for these types?

timmolter:
look in the main `.gitignore` file:
```
# externalized configuration file for keys
exchangeConfiguration.json
*secret.keys
```

Can you do it that way?

makarid:
Sure no problem

PR #4646, opened 2022-12-31
[Binance] Added trailing delta support via value TrailingFlag

mrmx:
The trailing delta was already added as a param, but not directly accesible. Thus the dto implemented in this PR as an IOrderFlags.


mrmx:
Any advance in merging this PR?

PR #4647, opened 2023-01-01
[krakenfutures] Adding Kraken futures support

makarid:


timmolter:
is kraken-futures a replacement to crypto-facilities? Same API, just a rename?

makarid:
> is kraken-futures a replacement to crypto-facilities? Same API, just a rename?

Yes, but i have added more endpoints. Cryptofacilities has been bought by Kraken so cryptofacilities doesn't exists anymore.

makarid:
Right now this implementation has an issue on kraken-futures-stream websocket authentication endpoints. If the stream gets disconnection or something happens and the connection restarts, it doesn't re-authenticate itself. I need to fix it

makarid:
I finally fixed the issue

PR #4659, opened 2023-01-13
Replace dependency to resilience4j-all by actually used modules (ratelimiter and retry)

gtoison:
Hello, first of all thank you for providing XChange!

We are looking to migrate our project from the javax to the jarkarta namespace and XChange currently brings a transient dependency to javax's cache API because XChange depends on resilience4j cache.
This PR replaces the resisilience4j dependency to the module that actually seem to be used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -246,7 +246,13 @@
 
       <dependency>
         <groupId>io.github.resilience4j</groupId>
-        <artifactId>resilience4j-all</artifactId>
+        <artifactId>resilience4j-ratelimiter</artifactId>
+        <version>${version.resilience4j}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>io.github.resilience4j</groupId>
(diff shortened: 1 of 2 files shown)


walec51:
core is the base for every module so I do not see what you gained here

plus we want resilience4j  to be used everywhere

we should check is a newer version of resilience4j  fixed those dependecies

gtoison:
Thank you for looking into it
With the proposed change the dependencies look like this:

image

And without the change they are:

image

So in particular the `cache-api` transient dependency is removed by that change. That one is a bit problematic because projects moving from the `javax` namespace to the `jakarta` namespace will find a conflict here.
Since `resilience4j-cache` (and other resilience4j modules) are not used, we can trim the dependency tree.

As you said there's also the pending upgrade to resilience4j 2.x but it seems like more work is required: https://github.com/knowm/XChange/pull/4631
So I felt like it should be an other Pull Request

Issue #4661, opened 2023-01-14
[Binance] Futures pull request changes broke xchange for some users


kovalu:
Hi @makarid, I've found out after upgrade to xchange 5.1.0 that your Futures changes in following PR broke xchange for some users
https://github.com/knowm/XChange/pull/4614

I'm in contact with couple of people using our service and for them the call to BinanceAccountService.getAccountInfo fails:
![image](https://user-images.githubusercontent.com/12104541/212476520-f75a9025-8492-4d58-b7a3-8e5474a00d1b.png)

The reason is the call to futuresAcount(). I don't have it confirmed yet, but I think that for people who never used futures this fails.
I don't think this should be called, or it should be called somehow conditionally based on getting some info that the user has futures account or the api key has permissions to call it.

Will you be able to fix that, @makarid?

kovalu:
I can confirm now, that with one user having this issue it helped that he activated the futures account on Binance.
Weirdly enough, the original API key still didn't work. He needed to create new API key.

This needs to be fixed so it works also for new accounts registered on Binance without activating the futures account.

TheCookieLab:
This is a blocking issue for those using Binance.us as the workaround of activating the future's account is not possible. Accessing any of the BinanceAccountService APIs results in:

>  Service unavailable from a restricted location according to 'b. Eligibility' in https://www.binance.com/en/terms. Please contact customer service if you believe you received this message in error. (HTTP status code: 451)

makarid:
Sure guys, I will make the futures disabled by default. This will fix the
issue, I didn't know that you have to activate the futures on Binance
first, I thought it is enabled by default.

TheCookieLab:
Thanks, appreciate the quick acknowledgement.

nibarulabs:
Any update on this as to when it'll get into a PR and a release?

nibarulabs:
I have built the Binance module from the develop branch and can confirm it is again working as expected with no futures enabled on Binance.us.

PR #4665, opened 2023-01-19
[Binance] Fix issue with Binance US init after adding Binance Futures

makarid:


makarid:
Possible fix for #4661 issue.

xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
```diff
@@ -77,6 +78,11 @@ public boolean isFuturesSandbox(){
             exchangeSpecification.getExchangeSpecificParametersItem(SPECIFIC_PARAM_USE_FUTURES_SANDBOX));
   }
 
+  public boolean isFuturesEnabled(){
+    return Boolean.TRUE.equals(
+            exchangeSpecification.getExchangeSpecificParametersItem(SPECIFIC_PARAM_FUTURES_ENABLED));
+  }
```
walec51, line 84: wont this turn off futures for every one after the upgrade?

wont this backward incompatibility be painful for existing clients?

rizer1980: support for futures on binance is just being introduced, the first commit was about three months ago. I don't think there's anything to break in terms of backwards compatibility...

PR #4669, opened 2023-01-30
[Binance] future order book impl#1

rizer1980:
The problem with binance FUTURES is that when you ping to the servers, there are about 350, i.e. one way delay 180ms - orderBook is never initialized.
We get eternal
`Orderbook snapshot for out of date. This is normal. re-syncing.`

I do not know how the method will behave with less delay, but in the place where I am developing the situation is this.
I also note that the problem concerns only futures, everything is fine on binance SPOT.

I added OrderBookFutureSubscription class to BinanceStreamingMarketDataService which saves received orderBook updates
, which allows for such delays to normally initialize orderBook.
The code is almost entirely taken from this pull request
https://github.com/knowm/XChange/pull/4361/commits
There are a lot of discussions, maybe the code is not perfect, but it works - I don’t understand why the request was not accepted in the end.
But the fact is that features only with these changes begin to work.
I have been using this code for binance FUTURES for several months.

Removed the subscription added by @makarid DEPTH20 from ProductSubscription because
this leads to the fact that for any subscription to orderBook we get two subscriptions DEPTH and DEPTH20 at once.
@makarid
I don't know how to implement it correctly...

Corrected the tests.


rizer1980:
bump

PR #4696, opened 2023-03-19
Idea codestyle

bigscoop:
Guys, how about commiting code style config for favorite editors and let the files be formatted by them instead of using `google-java-format` plugin?

Here I took official google code style for idea (https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml) and stored it within the project config. 

Also tested a fresh clone - the code style is picked up without any problems.


rizer1980:
if it works, then I'm all for it.
This google-java-format formatting often causes code merging problems.

bigscoop:
I can confirm it's working, but feel free to test

bigscoop:
Yes, exactly

PR #4706, opened 2023-03-23
[general] Change the scope of lombok to provided

bigscoop:
Lombok dependency is needed only at runtime, not at compile-time, so it has to be used with scope `provided`

see https://projectlombok.org/setup/maven


bigscoop:
ping @timmolter @makarid

walec51:
@bigscoop it we use it only at compile time I think it would be safe to just add it to the root pom.xml in

  
  <dependencies>

instead of the pom.xml of each module

as modules that do not use it will stay unaffected anyway

bigscoop:
@walec51 good idea! moved the dependency to parent pom only

also fixed wrong version of parent in xchange-stream-kucoin-module

bigscoop:
ping @walec51 @timmolter @makarid

bigscoop:
ping @walec51 @timmolter @makarid

PR #4710, opened 2023-03-29
feat: use jakarta namespace for SB3 support

martinformi:
I create this mainly for sake of discussion. Spring Boot 3 (and possibly quarkus) is using jakarta APIs instead of javax APIs. This PR makes it compatible, however, most probably it is a breaking change for other environments using `javax` APIs. How to solve this? XChange is for us quite important and we would like to not have a custom repo for that.

I do think this will be needed in the future, since Spring Boot 2.x has limited support already (maybe until the end of the year?). From other projects, I saw that they create two artifacts, one with javax and one with jakarta. Any ideas?


ppalaga:
Any chance to review, @makarid or @timmolter?

ppalaga:
Kindly ping?

walec51:
done in another PR

PR #4720, opened 2023-05-11
[refactor] Replace constructors in DefaultWithdrawFundsParams with according builders

bigscoop:
- Replaced constructors with according builders from lombok


walec51:
merging this would break backward compatibility because you are removing a few constructors from the public API

bigscoop:
agree, I'll close it

PR #4722, opened 2023-05-21
Add TRAILING_STOP_MARKET order type (Binance futures)

romanbsd:
If there's interest, I can create a PR with the support for placing such orders.


romanbsd:
Should I make a PR for actually using this order type?

PR #4723, opened 2023-05-24
Feature/polo okex candle data catch up

maghteriton:
Fixed the **periodInSecs** calculation


bigscoop:
looks like the version of org.knowm.xchange:xchange-stream-kucoin is wrong

aligokalppeker:
> 

It may be but we did not change any version, it has been done before. One another issue may be the compilation issue.

bigscoop:
yes, it was merged in develop by mistake, so you can fix it in your branch as well. 
just replace version of parent pom in `xchange-stream-kucoin/pom.xml` so that it looks like that
```
    <parent>
        <version>5.1.1-SNAPSHOT</version>
    </parent>
```

PR #4724, opened 2023-05-29
[core] Add support for junit 5

bigscoop:
Added dependency to `junit-vintage-engine` to enable junit 5 runner, but leaving existing junit 4 tests available and untouched


bigscoop:
wrong branch will redo

PR #4725, opened 2023-05-29
[core] Introduce junit5

bigscoop:
- Added dependency to `junit-vintage-engine` to enable junit 5 runner, but leaving existing junit 4 tests available and untouched
- Changed project code level to java 11


timmolter:
for compatibility with Android, we need to leave it on Java 8

bigscoop:
but the project can't be built with java 8 anymore

iromu:
> for compatibility with Android, we need to leave it on Java 8

Android supports 11 and 17

bigscoop:
ping @timmolter @walec51 @makarid

walec51:
Android does not support full Java 11. Android is Java 8 plus features backported from newer versions.

To bring back Android support we should add a CI Job that build a small example app that is build using Android SDK an that imports this library. 

Any one up for this task?

bigscoop:
Rolled back the code level changes. Now it is only junit 5

PR #4732, opened 2023-06-23
[binance] add order book orders timestamp

rizer1980:
fix limitorder null timestamp


xchange-binance/src/main/java/org/knowm/xchange/binance/service/BinanceMarketDataService.java
```diff
@@ -102,15 +102,16 @@ private OrderBook getBinanceOrderBook(Instrument instrument, Object... args) thr
   }
 
   public static OrderBook convertOrderBook(BinanceOrderbook ob, Instrument pair) {
+    Date timeStamp = Date.from(Instant.now());
     List<LimitOrder> bids =
             ob.bids.entrySet().stream()
-                    .map(e -> new LimitOrder(OrderType.BID, e.getValue(), pair, null, null, e.getKey()))
+                    .map(e -> new LimitOrder(OrderType.BID, e.getValue(), pair, null, timeStamp, e.getKey()))
```
walec51, line 108: this conflicts with this fields documentation:

a Date object representing the order's timestamp according to the exchange's server, null if not provided

timmolter:
If no timestamp from server, we set it to null.

PR #4733, opened 2023-06-23
[core] Order book synchronized

rizer1980:
Added OrderBook synchronization through the StampedLock mechanism, which is much faster and more flexible than usual Synhcronzed.
That gives us the opportunity to receive a full copy of the orderbook without a chance to get an Concurrent exception.


rizer1980:
bump

rizer1980:
ping

timmolter:
It's hard for me to tell if this PR works as designed in addition to not breaking things for someone relying on this.

rizer1980:
added concurrency tests.
1. New way, in use StampedLock
2. Old way, Stampdelock wrapped in Synchonized

PR #4735, opened 2023-06-29
switch to jakarta namespace

rvullriede:
An updated version of https://github.com/knowm/XChange/pull/4710 base on the current develop branch. With the major frameworks (like Spring Boot 3.x) moving to jakarta I personally think there is no need for two builds (javax, jakarta) but simply a new major version XChange. Whoever needs the latest version can/must should move to jakarta.


walec51:
look good, two questions:

1) in the previous PR we had an explicit dependancy to  jakarta.ws.rs-api
in this PR is this library still being added trasitavly by mmazi ?

2) does this in any way bump up the minimal java version one has to use?

rvullriede:
Hi, 

thanks for looking into this!

1. explicit declaration is not required since mmazi pulls the correct version already.
2. No. It still compiles with source and target set to 1.8 and it should run fine on such a JVM.

PR #4745, opened 2023-07-24
Ascendex cancel order fix

eeropu:
Current implementation of cancelOrder method for Ascendex trade service does not work correctly. It removes all orders based on currency pair and ignores the orderId.

Current version of cancelOrder is renamed to cancelAllOrdersByCurrencyPair and the cancelOrder method has been changed to remove only the order with the given id.

cancelOrder method's functionality is now inline with other trade services' method with the same name.


xchange-ascendex/src/main/java/org/knowm/xchange/ascendex/service/AscendexTradeService.java
```diff
@@ -32,6 +31,20 @@ public String placeLimitOrder(LimitOrder limitOrder) throws IOException {
 
   @Override
   public boolean cancelOrder(CancelOrderParams orderParams) throws IOException {
+    if (orderParams instanceof DefaultCancelOrderByInstrumentAndIdParams) {
```
walec51, line 34: you should type check by interface not their default implementation

please do `orderParams instance of CancelOrderByIdParams && orderParams instanceof CancelOrderByInstrument`

if both are needed

if only one is needed then use `||`

walec51 (changes requested):

PR #4764, opened 2023-08-17
[minor-refactoring] Replace explicit constructor calls with builder() method

bigscoop:
Replaced explicit `UserTrade.Builder` constructor call with equivalent `builder()` method call


bigscoop:
ping @makarid @walec51 @timmolter

PR #4766, opened 2023-09-02
[gateio] [gateio-streaming] Add gate.io api v4, gate.io streaming support

bigscoop:
- Added gate.io api v4 
- Extended gate.io streaming


bigscoop:
ping @timmolter @walec51

xchange-stream-gateio/pom.xml
```diff
@@ -1,26 +1,110 @@
...
+        <artifactId>xchange-parent</artifactId>
         <version>5.1.1-SNAPSHOT</version>
     </parent>
-    <modelVersion>4.0.0</modelVersion>
+    <artifactId>xchange-stream-gateio</artifactId>
 
     <name>XChange Gate.io Stream</name>
-    <artifactId>xchange-stream-gateio</artifactId>
+
+    <properties>
```
timmolter, line 13: versions should be defined in parent pom

bigscoop: can do it, but nothing from the defined versions is used in the other modules

timmolter: why introduce these 4 dependencies anyway?

bigscoop: - archunit - extends the tests, in these particular case ensures that there is no `Instant.now()` or `LocalDateTime.now()` calls in code. Only calls of now(Clock clock) are allowed. Otherwise impossible to test. See `ArchitectureTest`
- maven-enforcer-plugin - here makes sure that in final binary there are no dependencies with different versions, that can cause problems in runtime
- mockito - allows mocking during testing
- sortpom-maven-plugin - maintains the predefined structure of the pom file. Makes life way easier when merging the incoming changes.

None of these dependencies land in the final jar, they're used only during the tests.

pom.xml
```diff
@@ -327,6 +328,13 @@
         <scope>test</scope>
       </dependency>
 
+      <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-engine</artifactId>
```
timmolter, line 333: Is `junit-jupiter-engine` absolutely necessary?

bigscoop: It is JUnit 5 that is used in tests. And it is not going in the final jar because of scope `test`

bigscoop:
@douggie @timmolter any chance to review and merge?

I'm using this code for sophisticated high frequency trading 24/7 on gate.io for several months now. Any bugs discovered were fixed right away and merged in the branch, so it is tested and working.

Would be good to share it with the community

Issue #4779, opened 2023-10-15
is okx thread safe?


douggie:
I noticed a lot of okx requests get rejected with invaild sign when submitting multiple requests at same time.

```2023-10-15 09:27:46.33 [pool-18-thread-7] DEBUG si.mazi.rescu.HttpTemplate - Executing POST request at https://www.okx.com/api/v5/trade/cancel-order  body
{"instId":"BTC-USDT","ordId":"633717666398052352","clOrderId":null}
2023-10-15 09:27:46.33 [pool-18-thread-7] TRACE si.mazi.rescu.HttpTemplate - Request headers = {Accept=application/json, OK-ACCESS-KEY=abc key, OK-ACCESS-TIMESTAMP=2023-10-15T09:27:46.332Z, OK-ACCESS-SIGN=qr7/wIi21sfltMSzysANBSwoaqBOCM/XFrWhR8kk+Wc=, OK-ACCESS-PASSPHRASE=abc phrase, Content-Type=application/json}```

```2023-10-15 09:27:46.33 [pool-18-thread-14] DEBUG si.mazi.rescu.HttpTemplate - Executing POST request at https://www.okx.com/api/v5/trade/cancel-order  body
{"instId":"BTC-USDT","ordId":"633717666570018817","clOrderId":null}
2023-10-15 09:27:46.33 [pool-18-thread-14] TRACE si.mazi.rescu.HttpTemplate - Request headers = {Accept=application/json, OK-ACCESS-KEY=abc key, OK-ACCESS-TIMESTAMP=2023-10-15T09:27:46.332Z, OK-ACCESS-SIGN=LTCY6+x9eKDUlajK6evemhNCl62YbxG/q6zy1ml2q6w=, OK-ACCESS-PASSPHRASE=abc phrase, Content-Type=application/json}
```

Such issues don't happen on binace xchange implementation, they both seem to use the same decorator pattern, so wondering if there is something else here is not thread safe.

Binance
``` public BinanceCancelledOrder cancelOrderAllProducts(
      Instrument pair, Long orderId, String origClientOrderId, String newClientOrderId,  Boolean isMarginOrder)
      throws IOException, BinanceException {
    return decorateApiCall(
            () ->
                    (pair instanceof FuturesContract)
            ? binanceFutures.cancelFutureOrder(
                            BinanceAdapters.toSymbol(pair),
                            orderId,
                            origClientOrderId,
                            getRecvWindow(),
                            getTimestampFactory(),
[...]

PR #4780, opened 2023-10-16
4779_make_getmac_threadsafe

douggie:
https://github.com/knowm/XChange/issues/4779

### DIFF
--- a/xchange-core/src/main/java/org/knowm/xchange/service/BaseParamsDigest.java
+++ b/xchange-core/src/main/java/org/knowm/xchange/service/BaseParamsDigest.java
@@ -63,6 +63,10 @@ public abstract class BaseParamsDigest implements ParamsDigest {
   }
 
   public Mac getMac() {
-    return mac;
+    try {
+      return (Mac) mac.clone();
+    } catch (CloneNotSupportedException e) {
+      throw new IllegalStateException(e);
+    }
   }
 }


douggie:
These test failures are unrelated, seems like an issue with develop.

rizer1980:
it's seems that some error in(kraken futures)
1. no BTC in https://futures.kraken.com/derivatives/api/v3/instruments response (very strange)
2. lower case parsing "pf" prefix in adaptInstrumentsMetaData method
trying to deep in... 
@makarid ?

rizer1980:
I think I fixed it
post PR a little bit later

rizer1980:
done 
https://github.com/knowm/XChange/pull/4787

rizer1980:
@douggie 
is this fix just in case?
or is there code that does not work without this change?

timmolter:
Good fix — merging this in.

**Analysis:** `javax.crypto.Mac` is stateful and not thread-safe. It accumulates data via `update()` and resets on `doFinal()`. Returning the raw `mac` field from `getMac()` means any two concurrent callers (e.g., two parallel authenticated requests) share the same `Mac` instance, corrupting each other's HMAC state and producing incorrect signatures. Returning `mac.clone()` gives each caller an independently usable instance pre-initialized with the same key — the standard pattern for thread-safe MAC usage in Java. The `CloneNotSupportedException` wrapping is correct since `Mac` supports `clone()` per the JCA spec.

Issue #4769, opened 2023-09-10
Support binance portfolio margin 1.5


douggie:
Hi, binance's new portfolio margin 1.5 (offsetting spot and perp positions) requires new end points to place, cancel orders etc etc (https://binance-docs.github.io/apidocs/pm/en/#change-log).

Unless there are any objections I am proposing I add the following methods to `org.knowm.xchange.binance.service.BinanceTradeServiceRaw` and then add `USE_PORTFOLIO_MARGIN` order flag  to `org.knowm.xchange.binance.dto.trade.BinanceOrderFlags` to control the switching to the portfolio margin endpoints (i.e https://papi.binance.com) by using the  `USE_PORTFOLIO_MARGIN` order flag in `org.knowm.xchange.binance.service.BinanceTradeService`

- openPortfolioMarginOrdersAllProducts
- newPortfolioMarginFutureOrder
- newPortfolioMarginMarginOrder
- orderStatusPortfolioMarginAllProducts
- orderStatusPortfolioMarginAllProducts
- cancelPortfolioMarginOrderAllProducts
- cancelPortfolioMarginOrderAllProducts
- cancelAllPortfolioMarginOpenOrdersAllProducts
- allPortfolioMarginOrders
- myPortfolioMarginTradesAllProducts
- openPortfolioMarginPositions

douggie:
any comments? thoughts?

douggie:
@makarid any thoughts?

think i will add 2 flags`PORTFOLIO_MARGIN` and `INVERSE` so we can route new portfolio margin futures or spot orders to the `POST /papi/v1/um/order ` endpoints, as well as route the the coinmargin endpoints (i.e. inverse) https://binance-docs.github.io/apidocs/delivery/en/#new-order-trade and the current default usd margin endpoints, (i.e. linear) https://binance-docs.github.io/apidocs/futures/en/#new-order-trade

douggie:
As Binance has separate base URL's (i.e. https://papi.binance.com) i controlled it by setting exchange specific parameters.

douggie:
@walec51 @timmolter any chance of a review and merge?

douggie:
arggh tests failing on kraken, will try to take a look and fix them up.

PR #4785, opened 2023-11-02
4769 binance pm

douggie:
Support binance portfolio margin 1.5 #4769


douggie:
@walec51 @timmolter any chance of a merge?

PR #4787, opened 2023-11-08
[kraken futures] fix, instruments to uppercase

rizer1980:


rizer1980:
fix kraken futures stream also.
all pulibc test are passed, hope it's last commit

PR #4801, opened 2023-12-06
Ab chronos kucoin 5.1.2

abalasky-chronos:


xchange-kucoin/pom.xml
```diff
@@ -5,10 +5,16 @@
 	<parent>
 		<groupId>org.knowm.xchange</groupId>
 		<artifactId>xchange-parent</artifactId>
-		<version>5.1.1-SNAPSHOT</version>
+		<version>5.1.0</version>
```
timmolter, line 8: do not change this from `5.1.1-SNAPSHOT`

PR #4810, opened 2023-12-23
[CORE] Give ability to get separate snapshots of orderbook updates.

rizer1980:
Currently used in Binance Stream and Okx stream.
I also have a ready Bybit stream, which also has such a method.


rizer1980:
ping

Issue #4828, opened 2024-03-14
Bybit problem with Instrument conversion for options. java.text.ParseException: Unparseable date: "31MAY24"


wesssup:
Exception thrown : java.text.ParseException: Unparseable date: "31MAY24" in BybitAdapters.java line 180

In  BybitAdapters.java line 172:
.expireDate(OPTION_DATE_FORMAT.parse(expireDateString)) it is trying to parse date "31MAY24" as option date.
==>Will work for all environments in a US setting
==> Will not work for non-us environments. In Dutch it's MEI for example but will work for JAN,FEB,....DEC...)

It can be solved by changing line 47
From
private static final SimpleDateFormat OPTION_DATE_FORMAT = new SimpleDateFormat("ddMMMyy");
TO
 SimpleDateFormat OPTION_DATE_FORMAT = new SimpleDateFormat("ddMMMyy",Locale.US);

rizer1980:
please close

ycyangcai:
this issue have also in version 5.2.1，why?

rizer1980:
Hi, @ycyangcai 

Please provide more info or/and stacktrace of error.

rizer1980:
Hello, @wesssup 

Gurrent develop branch have working code, with Locale.US
https://github.com/knowm/XChange/blob/2a77edb9ded8b8c76dd7f5b06914b3ff4f605d90/xchange-bybit/src/main/java/org/knowm/xchange/bybit/BybitAdapters.java#L54-L55

PR #4829, opened 2024-03-14
Fix locale for date conversion in non-us environments

wesssup:
https://github.com/knowm/XChange/issues/4828

### DIFF
--- a/xchange-bybit/src/main/java/org/knowm/xchange/bybit/BybitAdapters.java
+++ b/xchange-bybit/src/main/java/org/knowm/xchange/bybit/BybitAdapters.java
@@ -55,8 +55,8 @@ public class BybitAdapters {
       balances.add(
           new Balance(
               new Currency(bybitCoinBalance.getCoin()),
-              new BigDecimal(bybitCoinBalance.getEquity()),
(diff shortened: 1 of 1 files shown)


xchange-bybit/src/main/java/org/knowm/xchange/bybit/BybitAdapters.java
```diff
@@ -53,8 +55,8 @@ public static Wallet adaptBybitBalances(List<BybitCoinWalletBalance> coinWalletB
       balances.add(
           new Balance(
               new Currency(bybitCoinBalance.getCoin()),
-              new BigDecimal(bybitCoinBalance.getEquity()),
-              new BigDecimal(bybitCoinBalance.getAvailableToWithdraw())));
+              new BigDecimal((bybitCoinBalance.getWalletBalance().isBlank()?"0":bybitCoinBalance.getWalletBalance())),
```
timmolter, line 58: The pattern we've decided on for this project is that if the server returns a blank value, then we set the value to null and not zero. Setting to zero in code could be dangerous in the case where from whatever reason the server makes a mistake and sends a blank value. Whereas when the server sends ZERO, it means zero.

PR #4859, opened 2024-04-28
Enable Compilation on JDK21

rvullriede:
The currently used lombok version is not compatible with JDK 21.
In addition, Its advised to switch from source/target to release (see https://openjdk.org/jeps/247 for details), introduced in Java 9.
The maven compiler plugin with translate this automatically to the old source/target/bootclasspath options for Java 8 compilers.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
-    <version.java>1.8</version.java>
+    <version.java>8</version.java>
     <version.assertj>3.25.3</version.assertj>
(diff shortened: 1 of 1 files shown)


rizer1980:
Hello.
Maybe it’s worth separating these (OHLC for Kraken, JDK 21) for two separate PR?

rvullriede:
@rizer1980, sorry that was unintentional. I'll cleaned this up. In addition, I noticed that meanwhile the problematic lombok dependency was updated already. What remains are the more robust compilation settings.

bigscoop:
How about changing the code level to 11? Since we're anyway building witth SDK 11. What do you think, @timmolter ?

timmolter:
@bigscoop ok, let's try it after the next release (coming this week). That way Java 11 will be present in the SNAPSHOT versions for a while before releasing it. We've tried several times in the past, but people reported issues with Android. Maybe it will work out this time around.

PR #4872, opened 2024-05-07
[general] Update mockito

bigscoop:
- Moved mockito version management to parent pom
- Upgraded mockito to version `5.11.0`


timmolter:
wow, nice!

PR #4875, opened 2024-05-09
[general] Remove method implementation that just throw exception

bigscoop:
- Removed empty method implementations that just throw `NotYetImplementedForExchangeException` because it is already thrown in default method in corresponding interface

@timmolter I don't expect any breaking changes here, but you never know... That's why I propose merging it after the release `5.1.2`


timmolter:
just going to merge it now...

PR #4879, opened 2024-05-20
[general] Define slf4j-api version

bigscoop:
Defined the used version of `log4j-api` so that it is not overwritten via transitive dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -353,6 +353,12 @@
         <version>${version.qos.logback}</version>
       </dependency>
 
+      <dependency>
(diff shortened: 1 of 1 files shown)


timmolter:
you meant to write slf4j, I believe...
Anyway, that wasn't defined? Weird.

bigscoop:
ah, sure `slf4j-api`

bigscoop:
my bad, I missed the already defined dependency in `dependencyManagement` section

good catch @timmolter :+1:

PR #4896, opened 2024-06-05
[Bitstamp] include txid in toString()

tross-nydig:
Small followup to [the recent PR that added txid to the WithdrawalRequest object](https://github.com/knowm/XChange/pull/4889).


bigscoop:
Hi. Please use `@ToString` from lombok and delete the method `toString()`

tross-nydig:
> Hi. Please use `@ToString` from lombok and delete the method `toString()`

Great idea, the previous `toString()` also included the additional derived `status` enum (i.e. `status=finished`) but I was able to keep it in there by using the @ToString.Include annotation.

PR #4902, opened 2024-06-11
[Bitstamp] Implement client code for withdrawal-fees endpoint

tross-nydig:
The Bitstamp V2 API has [an endpoint for requesting all withdrawal fees](https://www.bitstamp.net/api/#tag/Fees/operation/GetAllWithdrawalFees):

`https://www.bitstamp.net/api/v2/fees/withdrawal/`

This code implements the client code necessary to call that endpoint with the XChange lib: a `WithdrawalFee` object, the web service skeleton in `BitstampAuthenticatedV2`, and the implementation in the `BitstampAccountServiceRaw` class.

### DIFF
--- a/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAuthenticatedV2.java
+++ b/xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAuthenticatedV2.java
@@ -17,6 +17,7 @@ import org.knowm.xchange.bitstamp.dto.account.BitstampDepositAddress;
 import org.knowm.xchange.bitstamp.dto.account.BitstampRippleDepositAddress;
 import org.knowm.xchange.bitstamp.dto.account.BitstampWithdrawal;
 import org.knowm.xchange.bitstamp.dto.account.DepositTransaction;
+import org.knowm.xchange.bitstamp.dto.account.WithdrawalFee;
 import org.knowm.xchange.bitstamp.dto.account.WithdrawalRequest;
 import org.knowm.xchange.bitstamp.dto.trade.BitstampCancelAllOrdersResponse;
 import org.knowm.xchange.bitstamp.dto.trade.BitstampOrder;
@@ -653,6 +654,16 @@ public interface BitstampAuthenticatedV2 {
       @HeaderParam("X-Auth-Timestamp") SynchronizedValueFactory<String> timeStamp,
       @HeaderParam("X-Auth-Version") String version,
       @FormParam("timedelta") Long timeDelta)
+      throws BitstampException, IOException;
+
+  @POST
+  @Path("fees/withdrawal/")
+  WithdrawalFee[] getWithdrawalFees(
+      @HeaderParam("X-Auth") String apiKey,
+      @HeaderParam("X-Auth-Signature") ParamsDigest signer,
+      @HeaderParam("X-Auth-Nonce") SynchronizedValueFactory<String> nonce,
+      @HeaderParam("X-Auth-Timestamp") SynchronizedValueFactory<String> timeStamp,
+      @HeaderParam("X-Auth-Version") String version)
       throws BitstampException, IOException;
 
   @POST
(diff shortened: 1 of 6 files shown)


xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/dto/account/WithdrawalFee.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.ToString;
+import org.knowm.xchange.currency.Currency;
+
+@Getter
+@ToString
+@Builder
+@AllArgsConstructor
+public class WithdrawalFee {
```
bigscoop, line 14: consider using `@Value` together with `@Builder`

xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/dto/account/WithdrawalFee.java
```diff
@@ -0,0 +1,25 @@
...
+
+  private String network;
+
+  private BigDecimal fee;
+
+  private Currency currency;
+
+  public WithdrawalFee() {
```
bigscoop, line 22: can be replaced with `@NoArgsConstructor`

tross-nydig: I'll just scrap this constructor in lieu of an immutable `WithdrawalFee` and `@Value` + `@Builder`

tross-nydig: Actually, that breaks the unmarshalling test: 
`Error:    WithdrawalFeesJSONTest.testUnmarshal:26 » InvalidDefinition Cannot construct instance of `org.knowm.xchange.bitstamp.dto.account.WithdrawalFee` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)`

bigscoop: imho, you can go with `@Jacksonized` so that jackson uses builder, or also add `@NoArgsConstructor`

xchange-bitstamp/src/main/java/org/knowm/xchange/bitstamp/BitstampAuthenticatedV2.java
```diff
@@ -655,6 +656,16 @@ WithdrawalRequest[] getWithdrawalRequests(
       @FormParam("timedelta") Long timeDelta)
       throws BitstampException, IOException;
 
+  @POST
+  @Path("fees/withdrawal/")
+  WithdrawalFee[] getWithdrawalFees(
```
bigscoop, line 661: can you use List here instead of array?

bigscoop: As far as I know there is no performance differences there. I personally see the replacement with `List` as a part of code update activities, but again it is my humble opinion :smile: 

Let's ask @timmolter for his opinion.

I've also created small test for it and replaced the return type with list. Successfully tested with integration test, so it is working. See https://github.com/knowm/XChange/pull/4907

tross-nydig: Happy to `List`-ify this new endpoint as well as `getWithdrawalRequests()` since both are already being translated into Lists inside `BitstampAuthenticatedV2`. The `getUserTransactions()` endpoint is a little trickier, since that returns `BitstampUserTransaction[]` all over the place, in an outward-facing way that even my own application is currently using.

tross-nydig: Made the `List` change for `getWithdrawalFees()` and `getWithdrawalRequests()`

PR #4949, opened 2024-09-18
[bitstamp] add ledger types, fix 'None' fee

donald-jackson:


xchange-bitstamp/src/test/java/org/knowm/xchange/bitstamp/BitstampCliTester.java
```diff
@@ -0,0 +1,38 @@
+package org.knowm.xchange.bitstamp;
```
timmolter, line 1: turn this into an Integration test. See: <https://github.com/knowm/XChange/wiki/New-Implementation-Best-Practices#step-9>

PR #4952, opened 2024-10-08
[bybit] place advanced order

rizer1980:
implements all BYBIT specified parameters according to 
https://bybit-exchange.github.io/docs/v5/order/create-order


rizer1980:
ping

PR #4967, opened 2024-11-22
[okex-stream] all login-required channels become invalid after a network disconnection and reconnection

lvxiao1:
**Steps to Reproduce:**  
1. Use the `getUserTrades` method to subscribe to order events.  
2. Disconnect and then reconnect the network.  
3. Place order on OKX; the order events will not be received.

image
image


lvxiao1:
@timmolter please check

PR #4969, opened 2024-11-26
[CORE] orderbook recheckidx fix

rizer1980:
In the previous implementation there was a bug if the size of the order array was zero.
I did not go through all the check options. And limited myself to only the two most common options. I do not want to make a mistake somewhere again.
add tests to recheckIdx method


rizer1980:
@timmolter 
What needs to be added or fix?

PR #4974, opened 2024-12-03
[fix]When using a proxy with WebSocket, the DNS resolution should be performed within the proxy server.

lvxiao1:
Some exchanges use DNS and CDN technologies. If the domain name is resolved locally, it will return the IP address of the nearest server to the exchange from the local network. In this case, the SOCKS client will pass this IP to the proxy server, and the proxy server will attempt to connect using this IP, leading to a timeout. The best approach is to have the proxy server handle the DNS resolution.

Here is an analysis of part of the Netty source code:
1. `Bootstrap.java`
image
image
2.`Socks5ProxyHandler.java`
image


lvxiao1:
@timmolter please check

PR #5028, opened 2025-04-23
Fix TradeOgre currency pair adapter

kevintanhongann:
The TradeOgre exchange uses a base-counter currency pair format, but the adapter was incorrectly formatting it as counter-base. This commit fixes the adapter to correctly format the currency pair as base-counter.


rizer1980:
looks ok now.
wait for @timmolter for merge

PR #5046, opened 2025-05-22
[core, okx, binance, bybit] Get fee rates

rizer1980:
[core]
AccountService getDynamicTradingFeesByInstrument - add Optional category param

[okx, binance, bybit]
add/fix getDynamicTradingFeesByInstrument

[binance]
add separate resilience to Future account
add/fix stream futures api

[okx stream]
merge public and private channels in one connection


rizer1980:
@timmolter 
somethink wrong?
Or merge after Xchange release?

xchange-core/src/main/java/org/knowm/xchange/service/account/AccountService.java
```diff
@@ -221,7 +223,8 @@ default List<FundingRecord> getFundingHistory(TradeHistoryParams params) throws
    *     requested function or data, but it has not yet been implemented
    * @throws IOException - Indication that a networking error occurred while fetching JSON data
    */
-  default Map<Instrument, Fee> getDynamicTradingFeesByInstrument() throws IOException {
+  default Map<Instrument, Fee> getDynamicTradingFeesByInstrument(Object... category)
+      throws IOException {
```
timmolter, line 228: is there a more specific object such as String... that could be used instead of Object...

rizer1980: Yeah, I didn't think about that.

rizer1980:
bump

PR #5049, opened 2025-05-26
[general] Integration tests badges

bigscoop:
Currently we skip the integration tests during the build process because they need to access real exchange API some of them are not stable, some are disabled.

However there are still some integration tests implemented that are stable and access the exchanges' public API (without authentication etc), most of them get the market data. We can periodically execute them to make sure that the implementation is still working and the real API contracts are not broken (e.g. due to the bug or end-of-life of certain API version).

This PR contains following:
- Added workflows that execute integration tests once per day at 05.00am or on commit in `develop` for selected exchanges 
- Extended README with badges for results of the corresponding actions

P.S. I could test locally, README should look like that

image

Text can be surely extended, and also I'm not sure if feature will work from git runners ip addresses, could not test the actions with them. I propose we merge and see, in the worst case I can happily revert.

P.P.S Idea for the future improvement - notifications to maintainers if integration tests are broken.


bigscoop:
@timmolter any thoughts?

timmolter:
I'm leaning towards not adding this feature. Main reason is that probably no one is going to react anyway to broken tests. We might also run into GH Actions limits since this repo already runs quite a bit of workflow time.

bigscoop:
I'm also not strongly convinced, to be honestly. 
Tried the badges for some time with self-hosted git instance and they helped me to reveal the new status in bitmex (fixed here https://github.com/knowm/XChange/pull/5072).
My only hope would be that the red status of any badge will be reported as an issue by someone who notices.

On one hand we can try and see if the badges will be useful, on the other hand we can skip the idea.

Feel free to either merge or close

rizer1980:
Personally, I like the idea.
If there are problems with the GH Actions load, can make the launch less frequent

PR #5060, opened 2025-06-11
[globitex] Remove non-existing exchange

bigscoop:
Looks like globitex.com doesn't exist anymore. Removed module


bigscoop:
ping @timmolter

PR #5061, opened 2025-06-11
[kucoin] Extend functionality

bigscoop:
- Added annotations to some dtos
- Introduced error adapter for some calls
- Added helper method `Balance.builder()`
- Extended getting of account info
- Added deposit/withdraw functionality


bigscoop:
ping @timmolter

PR #5075, opened 2025-07-04
[binance, binance-stream]

rizer1980:
draft for now, i think to test it 1-2 weeks before allow to merge

[binanace-stream]
add get stream position changes
add get stream order changes
add get stream user trades
more separation before FUTURES and Portfolio Margin mode
test rework - tried to arrange everything into logical groups. With an eye on this pull #5049 some bug fixes

[binance]
add modify order
add set leverage
update rate limiter limits


rizer1980:
seems stable now

PR #5079, opened 2025-07-14
xchange-002: add default cast for instrument to currency pair

apwiegman:
Goal: add support in Bitso for getOrderBook by Instrument as support for getOrderBook by CurrencyPair has been deprecated


timmolter:
someone else had a different solution that's already been merged. Does it work for you?

PR #5089, opened 2025-07-16
[okx, okx-stream]

rizer1980:
draft, need test some time
[okx]
add setLeverage
add user reference id param to cancelOrder method
adaptTicker fix
[okx-stream]
add websocket stream get position changes
add websocket stream get user orders changes

some OkexAdapters fixs


rizer1980:
seems stable now

PR #5091, opened 2025-07-20
[bleutrade] Remove exchange

bigscoop:
Removed non-existing exchange `bleutrade`

https://www.cryptowisser.com/exchange/bleutrade/


timmolter:
https://www.cryptowisser.com/exchange-graveyard/

PR #5092, opened 2025-07-20
[cryptowatch] Remove exchange

bigscoop:
Looks like exchange was merged with kraken. 

https://cryptowat.ch/  leads to

https://blog.kraken.com/product/cryptowatch-to-sunset-kraken-pro-to-integrate-cryptowatch-features


timmolter:
@bigscoop it looks like bittrex is also dead: https://www.cryptowisser.com/exchange/bittrex/

bigscoop:
> @bigscoop it looks like bittrex is also dead: https://www.cryptowisser.com/exchange/bittrex/

@timmolter good catch. created https://github.com/knowm/XChange/pull/5095

PR #5117, opened 2025-08-20
[general] Move mockito dep to parent

bigscoop:
Moved `mockito-core` dependency with scope test to parent, since most of the modules are using it for the tests anyway


timmolter:
cool. please fix merge conflicts. Sorry.

PR #5133, opened 2025-09-12
[binance] websockets trade

rizer1980:
binance futures websocket trade:
*place(market nad limit) order
*change order
*cancel order
RateLimiters for this.
new UserTradeService for websocket trade
add support ed25519 keys support. It's need to UserTradeService one time login.


xchange-stream-binance/pom.xml
```diff
@@ -18,6 +18,12 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.bouncycastle/bcpkix-jdk18on -->
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
```
bigscoop: I think it is not needed here, since it comes as transitive dependency from `xchange-binance`

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/dto/trade/BinanceWebsocketLoginResponse.java
```diff
@@ -0,0 +1,23 @@
+package info.bitrich.xchangestream.binance.dto.trade;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.ToString;
+
+@Getter
```
bigscoop: you can use `@Value` instead of `@Getter` + `@AllArgsConstructor` + `@ToString`

xchange-stream-binance/pom.xml
rizer1980: Thx. fixed

xchange-stream-binance/src/main/java/info/bitrich/xchangestream/binance/dto/trade/BinanceWebsocketLoginResponse.java
rizer1980: yeap. fixed

timmolter:
fix conflicts and I'll merge...

PR #5138, opened 2025-09-20
Kraken: Fixed parsing of instrument metadata

ivans083:


xchange-coinex/src/main/java/org/knowm/xchange/coinex/CoinexAdapters.java
```diff
@@ -111,6 +112,8 @@ public InstrumentMetaData toInstrumentMetaData(CoinexCurrencyPairInfo coinexCurr
         .minimumAmount(coinexCurrencyPairInfo.getMinAssetAmount())
         .volumeScale(coinexCurrencyPairInfo.getBaseCurrencyPrecision())
         .priceScale(coinexCurrencyPairInfo.getQuoteCurrencyPrecision())
+        .amountStepSize(BigDecimal.ONE.movePointLeft(coinexCurrencyPairInfo.getBaseCurrencyPrecision()))
+        .priceStepSize(BigDecimal.ONE.movePointLeft(coinexCurrencyPairInfo.getQuoteCurrencyPrecision()))
```
timmolter, line 116: what's the reason to not just change the base precision?

ivans083: yes I am inferring step size from precision, if that's the concern, as the best fallback possible the API allows. Ideally I'd like to have a BaseBroker request step size regardless the specific exchange implementation (being step size a more meaningful constraint than precision). I'd like to know your thoughts on how to approach this issue.

ivans083: another option would be to handle the fallback from the BaseBroker itself, although my reasoning was that if the exchange does not provide stepSize is likely because they allow any stepSize within the precision, therefore that calculation would fit fine.

timmolter:
all good. I didn't realize the value came from the exchange.

PR #5144, opened 2025-09-26
Sync fork issue

yanyasha228:


rizer1980:
Hello, may be add some description?

rizer1980 (commented):
We seem to have switched to java 11 not long ago

xchange-binance/src/main/java/org/knowm/xchange/binance/dto/trade/TimeInForce.java
```diff
@@ -7,7 +7,8 @@ public enum TimeInForce implements IOrderFlags {
   GTC,
   GTX,
   FOK,
-  IOC;
+  IOC,
+  GTE_GTC;
```
rizer1980: there is no GTE_GTC in Binance

PR #5165, opened 2025-10-08
[binance] Make concludeHostParams overridable to customize endpoints (private static → protected)

fiserro:
Summary

Change BinanceExchange#concludeHostParams from private static void to protected void so subclasses can override it. This enables adjusting sslUri (and related host params) before super.applySpecification(...) runs.

Change
```
- private static void concludeHostParams(ExchangeSpecification exchangeSpecification)
+ protected void concludeHostParams(ExchangeSpecification exchangeSpecification)

```
And it’s invoked from:

```
@Override
public void applySpecification(ExchangeSpecification exchangeSpecification) {
  concludeHostParams(exchangeSpecification);
  super.applySpecification(exchangeSpecification);
}
```

Motivation

In local and test environments, we run the app against a mock server / reverse proxy. With the current hard-coded endpoints, changing sslUri isn’t possible without altering core code. Allowing an override lets consumers redirect Binance endpoints (e.g., to a proxy) cleanly in a subclass.

Alternative considered

Expose all hard-coded endpoints via configuration and have BinanceExchange honor them. That would be more comprehensive but larger in scope. This PR is a low-hanging fruit that unblocks real use cases immediately with minimal surface change.

Backward compatibility

No public API removed; default behavior is unchanged unless overridden.

Moving from private static to protected does not break existing clients.

Example usage
```
public class BinanceExchangeProxy extends BinanceExchange {
  @Override
  protected void concludeHostParams(ExchangeSpecification spec) {
    spec.setSslUri("https://binance-proxy.test.mydomain.io:30443");
  }
}
```

Testing

Verified default flow keeps original endpoints.

Verified subclass can redirect sslUri to a mock/proxy and requests succeed.

### DIFF
--- a/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
+++ b/xchange-binance/src/main/java/org/knowm/xchange/binance/BinanceExchange.java
@@ -156,7 +156,7 @@ public class BinanceExchange extends BaseExchange implements Exchange {
   }
 
   /** Adjust host parameters depending on exchange specific parameters */
-  private static void concludeHostParams(ExchangeSpecification exchangeSpecification) {
+  protected void concludeHostParams(ExchangeSpecification exchangeSpecification) {
     if (exchangeSpecification.getExchangeSpecificParametersItem(EXCHANGE_TYPE) != null) {
       switch ((ExchangeType)
           exchangeSpecification.getExchangeSpecificParametersItem(EXCHANGE_TYPE)) {
--- a/xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/BitmexExchange.java
+++ b/xchange-bitmex/src/main/java/org/knowm/xchange/bitmex/BitmexExchange.java
@@ -42,7 +42,7 @@ public class BitmexExchange extends BaseExchange {
       };
 
   /** Adjust host parameters depending on exchange specific parameters */
-  private static void concludeHostParams(ExchangeSpecification exchangeSpecification) {
+  protected void concludeHostParams(ExchangeSpecification exchangeSpecification) {
 
     if (exchangeSpecification.getExchangeSpecificParameters() != null) {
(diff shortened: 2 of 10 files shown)


bigscoop:
did you try creating the exchange like that?

```
            ExchangeSpecification exSpec = new ExchangeSpecification(BinanceExchange.class);
            exSpec.setSslUri("youproxy");
            var exchange = (BinanceExchange) ExchangeFactory.INSTANCE.createExchange(exSpec);


```

fiserro:
@bigscoop Yes, the problem is that I'd like to make it work for a sandbox as well, and the only way is not to set the EXCHANGE_TYPE parameter. What if somebody adds some logic expecting this parameter to be required. Or, there might be a default value in the future. It just seems to be too fragile to have a possibility of changing the sslUrl only if you didn't set the exchange type.
I don't really see any advantage of having that method private static

fiserro:
@bigscoop Changing the private static to protected is the easiest way, but I'm open to any other solutions. We can add a new parameter e.g. SKIP_CONCLUDE_HOST, or override the hardcoded URIs with parameters.

bigscoop:
ok, I see your point, don't see any problems then. 

May I ask you then to do the same for other exchanges that have the same private static method? I found 10 occurrences in total

fiserro:
Sure

fiserro:
@bigscoop I've changed all 10 private methods to protected

bigscoop:
thanks. looks good to me

fiserro:
@bigscoop, will you approve and merge, or should I take further action?

bigscoop:
@timmolter can do it

fiserro:
@timmolter hey Tim! 
Thank you for accepting my pull request. 
When do you guys plan to release a new version of Xchange?

timmolter:
I think I will soon

PR #5174, opened 2025-10-15
Improve the efficiency of StreamingAccountService when obtaining balances

tuxmonteiro:
Some exchanges (e.g., Kraken) deliver all balances when subscribing "balances," but the actual implementation requires multiple subscribes to get all account balances.
This PR proposes a more efficient approach by adding the "getBalancesChanges" method, returning Set<Balance> with all balances per currency.

### DIFF
--- a/xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingAccountService.java
+++ b/xchange-stream-core/src/main/java/info/bitrich/xchangestream/core/StreamingAccountService.java
@@ -7,6 +7,8 @@ import org.knowm.xchange.exceptions.ExchangeSecurityException;
 import org.knowm.xchange.exceptions.NotYetImplementedForExchangeException;
 import org.knowm.xchange.service.account.AccountService;
 
+import java.util.Set;
+
 public interface StreamingAccountService {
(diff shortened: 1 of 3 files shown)


bigscoop:
`KrakenStreamingService` emits all the received balances as single messages, thus no payload is lost. You just getting the balances separately, but no payload is lost or skipped

tuxmonteiro:
Hi @bigscoop, thanks for your comment. I think my confusion is due to the fact that `KrakenDataMessage::getPayload` has the following code:
```java
  public T getPayload() {
    if (data != null && !data.isEmpty()) {
      return data.get(0);
    }
    return null;
  }
```
The 'data.get(0)' onlys get the first element (in this case, the first balance), appearing to discard the entire remaining list.

I've confirmed that this is not true and if the currency is null in `Observable<Balance> getBalanceChanges(Currency currency, Object... args)`, the actual code creates a balance stream (an observable event per balance), not losing the data provided by the subscription.

I'm proposing #5175  to document this behavior and help others with the same question.  Thanks again.

bigscoop:
great, thanks :+1:

PR #5197, opened 2025-10-28
[bybit]

rizer1980:
orderbookupdate timestamp add more precies timestamp 
according bybit docs, it must use in case of sync orderbooks and trades


timmolter:
check failing tests
> The failing job is caused by an assertion error in the adaptOrderBookTest method in BybitStreamAdaptersTest.java. The test expects orderBook.getTimeStamp().getTime() to be 1672304484978L, but the actual value is 1672304484976L—a mismatch of 2 milliseconds.

PR #5207, opened 2025-11-05
[bybit] websocket batch cancel order

rizer1980:
add websocket batch cancel order
add bybit user trade service ping pong


rizer1980:
ping

PR #5215, opened 2025-11-15
feat(kucoin): add Earn holdings API support

sohnoun:
Adds support for retrieving Kucoin Earn holdings via the /api/v1/earn/hold-assets endpoint, enabling users to fetch their staking, savings, and other earn product positions.

API Integration
  • Added getEarnHoldings() method to AccountAPI interface
      • Supports optional query parameters: currency, productId, productCategory, currentPage, pageSize
      • Endpoint: GET /api/v1/earn/hold-assets
  • Implemented service methods:
      • KucoinAccountServiceRaw.getEarnHoldings() Raw API call with error handling
      • KucoinAccountService.getEarnHoldings() High-level service method

### DIFF
--- a/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinEarnHolding.java
+++ b/xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinEarnHolding.java
@@ -1,0 +1,60 @@
+package org.knowm.xchange.kucoin.dto.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.math.BigDecimal;
+import lombok.Data;
+
+/** DTO for Kucoin Earn Holding */
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class KucoinEarnHolding {
+
+  @JsonProperty("orderId")
+  private String orderId;
+
+  @JsonProperty("productId")
+  private String productId;
+
+  @JsonProperty("productCategory")
+  private String productCategory;
+
+  @JsonProperty("productType")
+  private String productType;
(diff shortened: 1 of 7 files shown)


xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinEarnHolding.java
```diff
@@ -0,0 +1,109 @@
+/** Copyright 2019 Mek Global Limited. */
```
bigscoop: typo?

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinEarnHolding.java
```diff
@@ -0,0 +1,109 @@
...
+  private String currency;
+
+  @JsonProperty("incomeCurrency")
+  private String incomeCurrency;
+
+  @JsonProperty("returnRate")
+  private String returnRate;
+
+  @JsonProperty("holdAmount")
+  private String holdAmount;
```
bigscoop: why not `BigDecimal`?

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinEarnHolding.java
sohnoun: Removed

xchange-kucoin/src/main/java/org/knowm/xchange/kucoin/dto/response/KucoinEarnHolding.java
sohnoun: Right, BigDecimal is the appropriate type

PR #5217, opened 2025-11-16
fix(kraken): handle currency codes with suffixes in toBalance()

### LINKED ISSUE #5216: [Bug] Kraken toBalance() throws ExchangeException for currency codes with suffixes (e.g., .F for earn subscriptions)
## Description
The `KrakenAdapters.toBalance()` method throws `ExchangeException` when processing currency codes with suffixes, such as `.F` used for Kraken Earn subscription redemptions (e.g., `AAVE.F`, `BTC.F`).

## Problem
When `getAccountInfo()` is called on a Kraken account that has active Earn subscriptions, the `getKrakenExtendedBalance()` API returns currency codes with `.F` suffixes (e.g., `AAVE.F`, `BTC.F`). 
The `toBalance()` method attempts to translate these codes using `adaptCurrency()`, which calls `KrakenUtils.translateKrakenCurrencyCode()`. Since these suffixed codes are not present in the `assetsMap`, the method throws:
`Kraken does not support the currency code AAVE.F`

## Impact
- Users with Kraken Earn subscriptions cannot retrieve account balances
- `getAccountInfo()` fails completely when `.F` currencies are present
- Future Kraken currency suffixes will cause the same issue

## Expected Behavior
The method should handle unknown currency codes gracefully by falling back to `Currency.getInstance()` when the standard mapping fails, similar to the old implementation. This ensures:
- Known currencies (BTC, ETH, etc.) are properly mapped via `assetsMap`
- Unknown currencies with suffixes (AAVE.F, etc.) are created directly with their full code
- Future suffixes are handled automatically without code changes

## Proposed Solution
Add a try-catch fallback in `toBalance()` to preserves the standard mapping for known currencies while gracefully handling unknown codes with suffixes.

sohnoun:
Add fallback for unknown currency codes (e.g., .F for Earn subscriptions) to prevent ExchangeException when processing account balances.

Fixes #5216

### DIFF
--- a/xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenAdapters.java
+++ b/xchange-kraken/src/main/java/org/knowm/xchange/kraken/KrakenAdapters.java
@@ -40,6 +40,7 @@ import org.knowm.xchange.dto.trade.MarketOrder;
 import org.knowm.xchange.dto.trade.OpenOrders;
 import org.knowm.xchange.dto.trade.UserTrade;
 import org.knowm.xchange.dto.trade.UserTrades;
+import org.knowm.xchange.exceptions.ExchangeException;
 import org.knowm.xchange.exceptions.NotYetImplementedForExchangeException;
 import org.knowm.xchange.instrument.Instrument;
 import org.knowm.xchange.kraken.dto.account.KrakenDepositAddress;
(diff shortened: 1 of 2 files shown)


bigscoop:
I would implement the fallback in `translateKrakenCurrencyCode` instead of throwing an exception, catching it etc

bigscoop:
see https://github.com/knowm/XChange/pull/5220

sohnoun:
> see #5220

Yes, t's simpler

PR #5231, opened 2025-12-04
[deribit-stream] Add streaming exchange

bigscoop:
### Deribit stream
- added tickers
- added trades
- added auth
- added user trades
- added position updates
- added rest requests


bigscoop:
@timmolter ping

PR #5241, opened 2026-01-10
[koineks] Remove exchange

bigscoop:
Removed obsolelete exchange.

As of https://www.cryptowisser.com/exchange/koineks/ it is not-existent


timmolter:
Thanks for removing all these old defunct exchanges!!!

PR #5256, opened 2026-02-04
[binance]  stream fundingrate fix

rizer1980:
[binance]
add orderbook ticker timestamp
stream fundingrate subscription on connect fix

[core]
Orderbook.class - make updatetime method public. It's need for external update orderbook states


rizer1980:
@timmolter ping

PR #5280, opened 2026-05-19
Bump to 6.0.0-SNAPSHOT: Java 17, resilience4j 2.4.0, crypto 5.0.2, json-path-assert 3.0.0

timmolter:
## Changes

- Bump project version: `5.2.6-SNAPSHOT` → `6.0.0-SNAPSHOT`
- Update Java source/target: `11` → `17`
- Update `resilience4j`: `1.7.1` → `2.4.0` (migrate from `io.vavr.control.Either` to `io.github.resilience4j.core.functions.Either`)
- Update `crypto` (web3j): `4.9.8` → `5.0.2`
- Update `json-path-assert`: `2.9.0` → `3.0.0`
- Update GitHub Actions workflows to JDK 17
- Update README version references to 6.0.0

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -14,10 +14,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 11
+    - name: Set up JDK 17
       uses: actions/setup-java@v4
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'temurin'
     - name: Build with Maven
       run: mvn clean deploy --no-transfer-progress --batch-mode --settings etc/settings.xml -Dfmt.skip=true;
--- a/.github/workflows/maven_on_pull_request.yml
+++ b/.github/workflows/maven_on_pull_request.yml
@@ -14,10 +14,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 11
+    - name: Set up JDK 17
       uses: actions/setup-java@v4
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'temurin'
     - name: Build with Maven
       run: mvn clean verify --no-transfer-progress --batch-mode --settings etc/settings.xml -Dmaven.javadoc.skip=true -Dfmt.skip=true;
--- a/README.md
+++ b/README.md
@@ -166,12 +166,12 @@ Add the following dependencies in your pom.xml file. You will need at least xcha
 <dependency>
   <groupId>org.knowm.xchange</groupId>
   <artifactId>xchange-core</artifactId>
-  <version>5.2.5</version>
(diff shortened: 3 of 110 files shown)


timmolter:
We need to bump to Java 17 in order to not get left behind with outdated dependencies. Since this is a major change, the 6.0.0 branch. After I merge this, I will continue with dependency upgrades. The reason I hesitated so long in upgrading to Java 17 is Android compatibility for older versions and IDE tools. But I think it's fine to do now.

timmolter:
## FAQ: Android Support for Java 17 Libraries

### Question
Do modern Android runtimes and IDEs support libraries built with Java 17?

### Answer
**Yes, modern Android runtimes and IDEs fully support libraries built with Java 17.** However, there are a few technical nuances regarding how that code is executed on older devices.

---

### 1. IDE and Toolchain Support
Modern Android development tools natively expect and support Java 17. 
* **Android Studio:** Android Studio (from Flamingo onwards) ships with a bundled JDK 17 (or newer JetBrains Runtime) to run the IDE and handle builds.
* **Android Gradle Plugin (AGP):** AGP 8.0 and above **require** JDK 17 or higher to run. 
* **Compilation:** You can safely set your `sourceCompatibility` and `targetCompatibility` (or Kotlin’s `jvmTarget`) to `JavaVersion.VERSION_17`. 

```groovy
// build.gradle (.kts)
android {
    compileOptions {
        sourceCompatibility = JavaVersion.VERSION_17
        targetCompatibility = JavaVersion.VERSION_17
    }
    kotlinOptions {
        jvmTarget = "17"
    }
}
```
2. Runtime Support (ART)
The physical Android Runtime (ART) on consumer devices handles Java 17 bytecode differently depending on the device's operating system version:

Android 14 (API 34) and Above: Devices running Android 14+ have core runtime support for Java 17 bytecode and APIs (such as sealed classes, pattern matching, records, and text blocks).

Android 13 and Older: ART on older devices cannot natively read Java 17 bytecode or recognize its standard library APIs.

How do Java 17 libraries run on older devices?
To prevent apps from crashing on older phones, the Android build toolchain uses a process called Desugaring (handled by the R8/D8 compiler).

During compilation, desugaring translates modern Java 17 bytecode, language features, and specific APIs into an older bytecode format (essentially Java 8-compatible DEX code) that legacy versions of ART can seamlessly understand.

Note on Library APIs: While language features (like switch expressions or text blocks) desugar completely, some core Java 17 standard library APIs (like java.util.random) cannot be fully desugared. Using them will require you to either bump your app's minSdk to 34 or perform a manual runtime API level check before execution.

Summary for Dependency Management
If you are importing a third-party library compiled with Java 17 into your project:

It will compile perfectly using modern versions of Android Studio and AGP.

It will run successfully on older devices because the Android build tools automatically "desugar" the library's Java 17 bytecode down to a format compatible with your project's defined minSdk.

PR #5305, opened 2026-05-24
Add Java 21 and 25 advisory builds to CI (#5304)

### LINKED ISSUE #5304: [build] Add Java 21 and Java 25 to CI as advisory builds
## Motivation

XChange currently builds and tests only against Java 17 (`<version.java>17</version.java>` in `pom.xml`). As Java 21 and 25 see growing adoption, we want early visibility into compile/test compatibility on those JDKs without changing our supported target. The new matrix entries should be informational only — failures must not block PR merges or develop pushes.

## Proposed change

Convert the single-JDK build in both workflows to a matrix over Java 17, 21, and 25. Java 17 remains the sole gating entry; 21 and 25 are marked experimental and use `continue-on-error: ${{ matrix.experimental || false }}` so they cannot fail the workflow.

Files:
- `.github/workflows/maven.yml` — push to `develop`. Java 17 entry keeps `mvn clean deploy`; 21/25 use `mvn clean verify` so no snapshots are published from non-17 JDKs.
- `.github/workflows/maven_on_pull_request.yml` — PRs to `develop`. All entries use `mvn clean verify`.

Both workflows set `fail-fast: false` so one experimental failure does not cancel the other JDK's run.

The Maven `<release>` target stays at 17 — we are testing that the project still builds/tests on newer JDKs, not retargeting bytecode.

## Acceptance criteria

- [ ] Each workflow shows three matrix entries: `build (17)`, `build (21)`, `build (25)`.
- [ ] A failure in `build (21)` or `build (25)` leaves the workflow's overall conclusion green.
- [ ] A failure in `build (17)` still turns the workflow red.
- [ ] On `maven.yml`, only the Java 17 entry runs `deploy`; 21 and 25 stop at `verify`.

## Follow-up / caveats

- [ ] **Branch protection.** The required status check name changes from `build` to `build (17)`. A maintainer with admin access to `knowm/XChange` needs to update the branch protection rule on `develop` accordingly, otherwise the required check will appear missing on new PRs.

kybercryst4l:
Closes #5304

### DIFF
--- a/.github/workflows/it-template.yaml
+++ b/.github/workflows/it-template.yaml
@@ -8,13 +8,24 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ '17' ]
+        include:
+          - java: '21'
+            experimental: true
+          - java: '25'
+            experimental: true
+    continue-on-error: ${{ matrix.experimental || false }}
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-java@v4
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
-          java-version: '17'
+          java-version: ${{ matrix.java }}
           cache: 'maven'
 
       - name: Build with maven
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -12,15 +12,30 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - java: '17'
+            goal: deploy
+          - java: '21'
+            goal: verify
+            experimental: true
+          - java: '25'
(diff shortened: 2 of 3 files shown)


kybercryst4l:
I am currently running the build with Java 25 (Temurin) and language level 17.
Any plans yet when to bump java version?

.github/workflows/maven.yml
```diff
@@ -12,15 +12,30 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
+      matrix:
```
kybercryst4l, line 17: should i add the java version matrix to `.github/workflows/it-template.yaml` as well? So not only the code can be build, but all the tests are executed and testet.

.github/workflows/it-template.yaml
kybercryst4l, line 1: I added it to each pull request as well, but would be enough to just execute on the pr's (maven.yml) or even only on the nightly test-run builds.

Because it's still lightweight, i added it everywhere to see a clear picture, when something breaks.

